PR #1013, opened 2014-12-29
Automation of end-date based on start-date

gauravmittal1995:
Hey,
@nursix Please Review
@flavour Removed the seperate js file and put in s3.datepicker.js

### DIFF
--- a/static/scripts/S3/s3.datepicker.js
+++ b/static/scripts/S3/s3.datepicker.js
@@ -55,3 +55,289 @@ S3.start_end_date = function(start_field, end_field) {
         }
     });
 };
+
+// Automation of end date if start date is present and automation is allowed
+// Widget Used: S3DateWidget
+// Will be activated only if end_date_selector and start_date_selector are both used
+(function($) {
+
+    "use strict";
+
+    /**
+     * Start Date - End Date Interval widget
+     */
+    $.widget('s3.end_date_interval', {
+
+        /**
+         * Default options
+         *
+         * @prop {bool} show_clear_button - Display the clear button next to the selector
+         */
+        options: {
+            show_clear_button: true
+        },
+        /**
+         * Create the widget
+         */
+        _create: function() {
+
+            var options;
+            options = this.options;
+
+            this.namespace = ".enddateinterval"
+
+            this.start_element = $(options.start_date_selector);
+
+            this.default_interval = options.interval;
+            this.start_date_prev = this.start_element.datepicker('getDate');
+
+        },
+
+        /**
+         * Update the widget options
+         *
+         */
+        _init: function() {
+
+            var default_interval,
+                select_end_date,
+                start_element,
+                start_month,
+                start_year,
+                start_date,
+                interval,
+                options,
+                element,
+                self;
+
+            // Set local variables
+            options = this.options;
+            self = this;
+            start_element = this.start_element;
+            element = $(this.element);
+            interval = this.interval;
+            default_interval = this.default_interval;
+
+            element.datepicker('option',{
+                yearRange:'c-100:c+100',
+                showButtonPanel:true,
+            }).one('click', function() {
+                $(this).focus()
+            });
+
+            // Insert a clear button next to the end_date which clears the field
+            // Only insert if needed (options)
+            if (options.show_clear_button) {
+                var clear_button = $('<button id="end_date_clear" class="btn date-clear-btn" type="button">'+ i18n.btn_clear +'</button>').click( function() {
+                    element.val('');
+                });
+                element.next().after(clear_button);
+            }
+
+            element.unbind(self.namespace).on('click' + self.namespace, function() {
+                self._inputclick();
+            });
+
+            select_end_date = element.datepicker('getDate');
+
+            if (select_end_date == null) {
+                start_element.change(function() {
+                    interval = self.interval;
+
+                    // Set end date only if its not set explicitly by the user.
+                    select_end_date = element.datepicker('getDate');
+                    if (select_end_date != null) {
+                        var start_date_prev = self.start_date_prev;
+
+                        // Confirm that the end date was set by the script before updating
+                        start_month = start_date_prev.getMonth() + 1;
+                        start_year = start_date_prev.getFullYear();
+                        start_date = start_date_prev.getDate();
+                        start_date_prev = new Date(start_year, start_month, start_date);
+                        self._add_months(start_date_prev, interval - 1);
+                        select_end_date = element.datepicker('getDate');
+                        if (self._equal_date(select_end_date, start_date_prev)) {
+                            self._end_date_automate(interval);
+                        }
+                    }
+                });
+            }
+
+        },
+
+        /**
+         * Sets the end date based on the start date and months given
+         *
+         * @param {integer} month - months to be added to the start date
+         */
+        _end_date_automate: function(months) {
+
+            var select_start_date,
+                start_element,
+                start_month,
+                start_year,
+                start_date,
+                element,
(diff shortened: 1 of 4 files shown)


gauravmittal1995:
@nursix , I still have to inject the script in a seperate file, This is just to verify whether the functionality is correct or not.

So, initally the end date is set to one year ahead of start date. The user has an option to clear this if he wants a permanent credential. this can be seen in [1]. Next if he changes the start date, end date changes with 1 year as an interval. If he puts his own end date (or clears it), then end date wont change if start date is changed

Next if he opens the end date widget, he sees [2]. Here he has an option to select the interval. Selecting any option adds the interval to start date and sets the end date. Then if he changes the start date, end date automatically changes according to the newly selected interval.

[1] -> http://picpaste.com/_1_-rGkD7ukx.png 
[2] -> http://picpaste.com/_2_-v2QBy0Dy.png

gauravmittal1995:
@nursix Any suggestions?

nursix:
Yup, that does what I had in mind - except for the explicit "clear" instead if explicit default (it should be blank at first, and default on click, maybe make this an option). Styling can of course be improved, but shouldn't be the initial focus.

Now you need to move and clean up the JavaScript - ideally, as suggested before, into a datepicker plugin, and only inject the options. Let's make sure that we do not hardcode any element IDs or relationships, and avoid globals (full encapsulation).

Good work so far :)

gauravmittal1995:
@nursix , Hey, so should i put the js in s3.datepicker.js ? or create a seperate file and inject the file from the python script?

nursix:
s3.datepicker.js is fine

gauravmittal1995:
@nursix , Please Review
I was thinking that for the default, it should be the one as above (i.e initally it should be one year and not blank) because, if he wants blank and default is one year, he can simply click the clear button and be done with it. But if he wants  1 year and default is blank, then he has to click on the end date and then somewhere else to hide the widget. Therefore , in the first case, only one click is required and it is easy as the method to clear is clearly visible. But in the second case, two clicks are required. Therefore I think that by default, it should be one year (i.e end date based on start date) and not blank.

flavour:
+1 to maintaining the 1 year default

nursix:
Well, to me that depends on whether or not the majority of instances have an end-date. If the normal case is "no end date", then an implicit default is nasty (as most of the time the user would have to remove it).

If we only look at credentials, and it is indeed true that the majority of credentials has a validity period of 1 year, then that may not be a problem.

But for a generic plugin I would make both options of the plugin: whether to set a default upon initialization, and what time period. Both options can well default to what Fran says - yet this should be options.

nursix:
And if you look at all the things which have start/end date combinations, then I doubt an implicit end-date default is always useful.
However, I appreciate that you wanted to solve this particular case, so you can make this a ToDo

nursix:
NB I do have a use-case for this plugin where the implicit default isn't nice: RDRT deployments.

While the users want to be able to set the end date from a time period (and hence this plugin would be a useful UX enhancement), they can not set the end date when initially creating the record, because by that time the period is usually unknown.

But once the member is deployed, the record would be updated with a deployment role and with a (prospective) end date - and then a period selection would be helpful. However, not the implicit default, in this case.

flavour:
My +1 for 1 year is for the usecase...i.e. this should be a configurable option for the widget.
I agree that the widget default should be blank

gauravmittal1995:
@nursix , Please Review.
I have set default to blank as u suggested.
Also added a Todo to make an option to set implicit default.

gauravmittal1995:
@nursix Any other suggestion?
@flavour Please review

nursix:
haven't had enough time today to review, sorry. Hopefully tomorrow.

gauravmittal1995:
@nursix Please Review

nursix:
...Tip: a possible alternative to injecting widget options is to bind data nodes to the DOM. However, that's a bit weak since DOM objects often need to be manipulated (or newly created) on the client side, so it could get lost.

gauravmittal1995:
@nursix Please Review

gauravmittal1995:
@nursix I didnt quiet understand the namespacing. Can u help me?

nursix:
Read more at http://api.jquery.com/bind 

Every jQuery UI widget should use its own namespaceID for local event bindings. Examples are in the code, e.g. s3.ui.locationselector.js

gauravmittal1995:
@nursix Please Review

gauravmittal1995:
@nursix , I am sorry i didnt reply to you on irc. Actually i wasn't online. I use IRCcloud and it lets the user be online for 2 hours after inactivity and thats y u saw me as online.

static/scripts/S3/s3.datepicker.js
```diff
@@ -55,3 +55,289 @@ S3.start_end_date = function(start_field, end_field) {
...
+    $.widget('s3.end_date_interval', {
+
+        /**
+         * Default options
+         *
+         * @prop {bool} show_clear_button - Display the clear button next to the selector
+         */
+        options: {
+            show_clear_button: true
+        },
```
gauravmittal1995: I made this option, so that for some instance when the end-date is required we can change this to false.

gauravmittal1995:
@nursix Please Review.

static/scripts/S3/s3.datepicker.js
nursix: I wouldn't hide the clear-button even if the end date is required.

static/scripts/S3/s3.datepicker.js
```diff
@@ -55,3 +55,289 @@ S3.start_end_date = function(start_field, end_field) {
...
+            }
+
+            var buttonPane = $('.ui-datepicker-buttonpane');
+            $('.ui-datepicker-current').hide();
+            $('.ui-datepicker-close').hide();
+
+            var pane_heading = $('<h5>Intervals</h5>');
+            pane_heading.appendTo(buttonPane);
+
+            // @ToDo: DRY (make an array instead of repeating the code and numbering the buttons)
```
nursix: You could action this ToDo, and make the actual intervals (and probably the number of buttons) configurable.

static/scripts/S3/s3.datepicker.js
```diff
@@ -55,3 +55,289 @@ S3.start_end_date = function(start_field, end_field) {
...
+            var btn_1 = $('<a class="ui-datepicker-current ui-state-default ui-priority-primary" type="button">+6 MO</a>');
```
nursix: Obviously, the button labels need internationalization.

nursix:
If you have the time, it would be good to action some of the ToDos.

It could be merged as a first iteration, unless @flavour objects - but I'd expect some improvements rather sooner than later.

flavour:
I can megre mid-way if green-lighted, but better to merge something more complete if possible...I see the missing i18n as a pretty serious issue.

gauravmittal1995:
@nursix , @flavour i18n done.

I will do some of the Todos (the DRY and the implicit default) as another PR.

static/scripts/S3/s3.datepicker.js
```diff
@@ -55,3 +55,289 @@ S3.start_end_date = function(start_field, end_field) {
...
+            options = this.options;
+            self = this;
+            start_element = this.start_element;
+            element = $(this.element);
+            interval = this.interval;
+            default_interval = this.default_interval;
+
+            element.datepicker('option',{
+                yearRange:'c-100:c+100',
+                showButtonPanel:true,
```
flavour: Beware comma after final member of set: Breaks compile as breaks older IE

PR #1018, opened 2014-12-31
SSF Task Subscriptions

hitesh96db:
- Allow 'Members' of Tasks to receive notifications.
- Commenting on tasks automatically subscribes the user to updates.


hitesh96db:
done

hitesh96db:
@flavour done :)

hitesh96db:
@flavour done :)

private/templates/SSF/config.py
```diff
@@ -13,6 +13,7 @@
 from gluon.html import *
 
 from s3 import *
+from applications.eden.private.templates.SSF.controllers import subscriptions
```
nursix: Hardcoded application name - not acceptable. Use relative path instead.

private/templates/SSF/config.py
```diff
@@ -533,12 +537,85 @@ def task_rheader(r):
...
+                return False
+        if r.representation == "s3json":
+            # Check if already a Member ?
+            task_id = r.get_vars.task_id
+            db = current.db
+            mtable = current.s3db.project_member
+            user = current.auth.s3_logged_in_person()
+            query = (mtable.task_id == task_id) & \
+                    (mtable.person_id == user)
+            rows = db(query).select(mtable.id,
```
nursix: Row limit is 80 characters, not 40.

private/templates/SSF/controllers.py
```diff
@@ -17,6 +17,7 @@
 from s3 import S3FilterForm, S3CustomController, S3OptionsFilter
 
 THEME = "SSF"
+T = current.T
```
nursix: Assignment from current to module-global - incorrect.

private/templates/SSF/config.py
```diff
@@ -533,12 +537,85 @@ def task_rheader(r):
...
+    standard_postp = s3.postp
+    def custom_postp(r, output):
+        # Call standard postp
+        if callable(standard_postp):
+            output = standard_postp(r, output)
+        if r.component_name == "member" and r.component_id and \
+            r.http == "DELETE":
+            # Remove Task subscription
+            sub = subscriptions()
+            sub.remove_task_subscription(str(r.id))
```
nursix: I would think that a delete-cascade is more appropriate here (and you do not check for success of the DELETE request).

private/templates/SSF/config.py
```diff
@@ -533,12 +537,85 @@ def task_rheader(r):
...
+    # Custom postp
+    standard_postp = s3.postp
+    def custom_postp(r, output):
+        # Call standard postp
+        if callable(standard_postp):
+            output = standard_postp(r, output)
+        if "item" in output:
+            # Bypass
+            current.response.headers["Content-Type"] = "application/json"
+        elif "success" in output and r.http == "PUT":
```
nursix: Funny success-test - success is actually a variable that can be True or False ;)

nursix: Better make this an onaccept?

private/templates/SSF/config.py
```diff
@@ -767,6 +844,81 @@ def customise_project_task_resource(r, tablename):
...
+    task_id = current.request.args[0]
+
+    if user:
+        query = (mtable.person_id == user) & \
+                (mtable.task_id == task_id)
+        member = db(query).select(mtable.role_id).first()
+        if not member:
+            query = (rtable.role == "Watching") & \
+                    (rtable.deleted == False)
+            role = db(query).select(rtable.id).first()
```
nursix: limitby = (0,1)

private/templates/SSF/config.py
```diff
@@ -533,12 +537,85 @@ def task_rheader(r):
...
+            query = (mtable.task_id == task_id) & \
+                    (mtable.person_id == user)
+            rows = db(query).select(mtable.id,
+                                    limitby = (0, 1))
+            if rows:
+                # Bypass request
+                result = {"bypass": True}
+                item = current.xml.json_message(success=True,
+                                                created=[rows[0].id])
+                result["output"] = {"item": item}
```
nursix: Merge the first and the third statement?

private/templates/SSF/config.py
hitesh96db: Relative path ? I am unable to get the import working.

private/templates/SSF/config.py
hitesh96db: It's actually "status" that's a variable here, that could be "success" or "failure" ;)

private/templates/SSF/config.py
nursix: Yeah - I see, you're trying to import from private/..., which is not possible.

But why is the class in controllers anyway - if it's only used in config?

private/templates/SSF/config.py
nursix: Err...according to the code above, output is a dict with {"item": json_message} - so the condition "success" in output will always fail, no? At least that is how I read this.

private/templates/SSF/config.py
hitesh96db: Fran told me about the possibility of memory leaks when defining classes in executed code, so he told it's better to keep it in places that are loaded as modules..
https://github.com/hitesh96db/eden/commit/21c8b64722f4e1a8ba27d6b4150c2a327e935017#diff-9c39b6a58ebe22321698bb6806a43b3fR1117

private/templates/SSF/config.py
nursix: That is you set:

    result = {"bypass": True}
    item = current.xml.json_message(success=True, created=[rows[0].id])
    result["output"] = {"item": item}

...which makes the REST API bypass the method call - and return result["output"], and that is, to me, a dict.

hitesh96db: output will be a dict only when the request is bypassed...otherwise when using RESTful "DELETE" it's a json string.

private/templates/SSF/config.py
hitesh96db: So, how should I go about this ?

private/templates/SSF/config.py
nursix: If the user is already a member of the task, then the request will always be bypassed even for a DELETE - so it will be a dict. How is a DELETE supposed to succeed?

Further - why return "created" if you bypassed the request? You didn't create anything.

And then - you look for a record with person_id and task_id - but you do not check whether it's deleted or not. You're lucky of course that both are foreign keys, so they would be set to None - but that's not really clean.

Eventually, this will always be bypassed except for the first PUT. Even if you try to DELETE, it will return a "created". Somehow this is a bit dodgy - maybe don't try tricks?

private/templates/SSF/config.py
nursix: That is nice, but you still must not hardcode the appname ;)

nursix: Besides, the risk for memory leaks is fairly small here - you can very well include a class in config.py (as it is done in various other places) as long as you avoid the __del__ method.

The point is that the code to destruct the instance would be inside the exec environment - so the instance can not be destructed after that environment ceased to exist, so it would remain in memory beyond the reach of the outer thread.

But I don't see a __del__ method in subscription?

nursix: I would say, either move the subscription class into config.py - which I give you my word is safe as long as you don't implement a __del__, or generalize it and move it into the model. 

I wouldn't mind a S3TaskSubscription class in s3notify.py either *if* it is properly generalized and thus re-usable across use-cases.

hitesh96db: I don't think there's a general use-case for it yet, few templates use subscriptions, and the use-cases are different. Subscriptions for Tasks is very specific to SSF for now...so maybe leave it as-is ?

nursix: For your understanding what causes the memory leak:

Let your code be this:

    class Example(object):
        def __del__(self):
            print "deleted"
    example = Example()

Now, when you run this in regular Python (like: python test.py), it will print "deleted".

But when you run it as web2py script (i.e. inside restricted), you will not see any output - the instance "example" is never destructed.

The reason is quite simple: the destructor method holds a reference to the exec environment, the exec environment holds a reference to the "example" instance, and the "example" instance holds a reference to the destructor method. So you have a perfect circular reference which prevents any of the three objects from being destructed because the interpreter doesn't know in which order.

To break the cycle, you can add:

    class Example(object):
        def __del__(self):
            print "deleted"
    example = Example()
    example = None

The last statement removes the reference to the "example" instance from the exec environment - thereby breaking the cycle. If you now run this through web2py, it will print "deleted".

When you import a module, the Python interpreter will do this setting to None automatically - which is why this does not happen for imported modules.

However, knowing this, you can avoid the problem - and hence very well put the class inside config.py since you do not have a destructor, or even if you had one, you would know how to break the cycle.

private/templates/SSF/config.py
hitesh96db: The idea behind the bypass was to take care of the situation where a user comments on a task, and when one does so, one becomes a member of the task and becomes subscribed to it. But the 'Watch' Button in the client-side isn't updated unless he/she refreshes the page. So if he/she then clickes 'Watch', he/she is already a member, so there isn't a need to add the user as a Member again, so just bypass, and take the "already-created" member id of the user and pass it to the client-side(which extracts the id from the created variable).

DELETE doesn't go through project/member but through project/task/[id]/member/[id]

private/templates/SSF/config.py
nursix: Leaving as-is is not possible - you must not hardcode the application name in the import statement. No way, sorry.

private/templates/SSF/config.py
hitesh96db: Sorry, it was RESTful "PUT", I got confused...

private/templates/SSF/config.py
hitesh96db: No no, it can be moved to config, no issue at all, I was talking about the S3TaskSubscriptions

hitesh96db: :+1: for the explanation

nursix: You can actually take my word for this - as long as you avoid the above problem, it is safe to have classes like this in config.py. It isn't recommended - but for very specific reasons, which you can easily get around.

If you don't trust my word, then an alternative could be to break the class up into functions and move those into config.py. It is questionable anyway whether this needs to be a class as you always only call a single function of the instance - but I understand that a class looks nicer and is perceived as easier to maintain.

The ultimate solution for this sort of problems would be to import templates as modules (instead of executing them) - and move them into the module path (modules/templates). But that is truly a major framework change and would require hundreds of hours of work - not to even speak of the retraining of everyone who has ever worked on templates, so it is currently rather wishful thinking, or, as we use to say, something for 2.0 ;) We could equally well think of migrating Eden to django, or cherrypy.

nursix: Well, it's up to you whether you want to generalize all the different subscription implementations into a common module with config options - or keep a template-specific one.

But yeah - that's rhetorics, of course ;) sorry if it sounded like a proposal. If you are sane, you will *not* try to generalize that.

private/templates/SSF/config.py
nursix: Certainly, that's how I would see it. Apart from that, I would actually *update* the Button to "Unwatch" ;) Shouldn't be too hard (but not required for this PR).

hitesh96db:
@nursix done

flavour:
The way to import from controllers.py would be to use this trick:
https://github.com/flavour/eden/blob/master/views/generic.xls#L5

Moving to a module not only avoids the potential memory leak (which the original link described clearly as only being an issue with a 'del' method, but still avoiding executing them is good as someone /could/ unwittingly add such a method later) but also about caching (modules are not reloaded every request except in debug mode).
Of course, this isn't critical.

nursix:
Not quite true - imports into the exec environment will be removed when the exec-environment is destructed so there is no caching advantage here ;)

nursix:
...but a module would be compiled only once (whilst config.py is compiled every request), and then byte-code executed - which indeed saves some processing. Question is whether that is faster than the constructed exec - or at all relevant.

The clean, pythonic solution would be to treat config as a module (and the template as a package), instead of executing it, which is possible with some refactoring (we do e.g. handle template menus and layouts as modules rather than executing them).

I think the critical thing here is not to hardcode the application name - the rest may belong into a wider problem area with the current template framework.

nursix:
Looks ok to me.

IMHO we have way too many function and class definitions in config.py - which should ultimately lead to a framework change. Either we move templates into modules/templates thereby allowing direct imports, or we treat templates as packages and config.py as modules (i.e. importing them rather than executing them). The former is easier (and a lot cleaner) - whereas the latter is somewhat "softer", with a lower learning curve and friendlier for downstream projects.

What's your opinion Fran?

nursix:
This is indeed a question of re-compiling config.py on every request - not so much about the (slightly far-fetched) risk for memory leaks. If we look at some templates, then there are thousands of lines of template-specific code in config.py - and that's actually intended to be possible.

flavour:
Friendlier to downstream seems good to me

nursix:
For a gradual shift, though, we could simply introduce modules/templates for template-specific modules, and leave the config.py execution as-is yet gradually moving the classes and functions into the modules folder.

In this case, it would mean to move the subscriptions() class into modules/templates/ssf.py - and then import it in config.py like:

    from templates.ssf import subscriptions

That would be both downstream-friendly yet clean-pythonic - and it would even allow re-using the same functions and classes across templates (something we wanted for some time already).

The learning curve seems low for this solution - the mere settings.xyz = value would remain in config.py, and we have all the time in the world to migrate bit by bit as we find the time, without pressure.

nursix:
If you agree, I would start this (once you have merged this one).

flavour:
The disadvantage here is that it splits a template even more...from 2 folders (private/templates/x & static/themes/x) to 3

nursix:
Very true - but is that a critical problem or just an inevitable trade-off?

flavour:
Some more thinking to avoid the issue would be good rather than rushing to action this as-is.
We already manage imports in controllers/menus from private...can we not use the same trick for config.py?

nursix:
Unfortunately, this would require a refactoring of all config.py's (not impossible, but a fair bit of work).

Since we attempt to access a current-variable (settings), we must move all config.py settings into a callable (either a class, or a function), then import config.py and then execute the callable.

Not sure whether downstreams can deal with such a change (it's not too hard, but at first prevents a simple pull-upgrade) - so introducing an (optional) modules/templates/<template>.py as a start seems a bit nicer.

Alternatively, we could introduce a fallback. If we change the name from config.py into template.py, where config.py is "old-style" (=executable script) whilst template.py is "new-style" (=importable module), then we could look for template.py and if it's not there, then execute config.py as we do now.

New templates (or migrated ones) would then be written in template.py style - whilst old templates would still work.

nursix:
A new-style template should then look like this:

    def config():
        settings = current.deployment_settings

        setttings.xyz.abcdef = "something"

We would then import and execute this like (via a constructed exec):

    import applications.<appname>.private.templates.<template>.config as template
    template.config()

nursix:
With this style, theoretically (if that makes any sense to you yet), you can define template globals and access them from everywhere. For example, if template.py would define a function:

    def example(x, y)

you can later access it like:

    template.example(x, y)

...without a need for a setting. So the hooks could introspect the template dict instead of calling S3Config, which is a fair bit faster (and cleaner).

nursix:
...but of course it makes templating a bit harder since you couldn't just introspect S3Config to find all possible hooks ;) so, this is purely theoretical.

nursix:
But - and this is where it really comes nice...

...you could have more than one config's in your template, e.g. different configs for different orgs. You first execute the common parts, then run sub-configs per org. That would be fantastic e.g. for the IFRC template - a lot less if/if/if to execute for every request just to determine the right config for this particular org.

nursix:
...i.e. less processing in the hooks (e.g. preps) - the prep would already be org-specific.

nursix:
Yup, tested with default - works fine. So, the requirement is to move the template's config into a callable (I used a function as described above). Fallback for old-style configs can be implemented, so that there's no urgency for downstreams to migrate.

nursix:
Anything that speaks against this mode?

flavour:
> We would then import and execute this like (via a constructed exec):
> import applications.<appname>.private.templates.<template>.config as template

This means that we're not caching though since we need to do this via exec()
This is what we're currently doing with menus.
Can we add private/templates to the Python Module Path?
If not, can we avoid this by moving the whole private/templates to modules/templates?

> Unfortunately, this would require a refactoring of all config.py's (not impossible, but a fair bit of       > work).

From what I see this appears pretty trivial to me...I'm certainly happy to do this.

> Not sure whether downstreams can deal with such a change (it's not too hard, but at first prevents > a simple pull-upgrade)
> Alternatively, we could introduce a fallback. If we change the name from config.py into template.py,    > where config.py is "old-style" (=executable script) whilst template.py is "new-style" (=importable           > module), then we could look for template.py and if it's not there, then execute config.py as we do   > now.

This eases migrateability for templates which aren't in trunk but implies changing a lot of docs.
I'd prefer to attempt new style & warn about the need to migrate if not new-style (as I say, the migration seems pretty trivial & I'd be happy to help individuals)

>But - and this is where it really comes nice...
>...you could have more than one config's in your template, e.g. different configs for different orgs.   > You first execute the common parts, then run sub-configs per org. That would be fantastic e.g. for > the IFRC template - a lot less if/if/if to execute for every request just to determine the right config   > for this particular org.

I can certainly see some benefits here, although the per-Org differences aren't always settings.xx so we won't remove all of the prep changes here, but can remove a few.

nursix:
Moving the whole of private/templates into modules/templates would be useful (as it would allow in-template imports), but doesn't solve the need to put the template config into a callable - so this would still be required.

Changing the Python module path from inside the application seems a bit unsafe to me (interfering with the custom importer) - and would prevent the custom importer from auto-reloading. But sure, we can test it.

I would though prefer to stick with the modules path - as that's a lot more logical and doesn't require any additional "tricks". Why are templates in private anyway if they are meant to be in the repository? I never really understood this.

nursix:
Btw - attempting to first import config.py, and then look whether it is old or new style involves a certain risk to the global namespace (because the config.py's we have would import, but execute illegal assignments at the module level).

Better to give new-style configs a new name - and fall back to config.py if no new-style config present.

nursix:
...however, it can be done that way of course. Just a bit risky.

nursix:
But *what* we can do is to combine the migration with the move to modules/templates ;)

If the template is in modules/templates, then we expect a new-style config.py - whereas if it is in private/templates, we would expect old-style.

I appreciate that this requires a lot of docs to be changed ;) happy to help with this (need some guidance, as I am not so familiar with the docs).

But it would allow a gradual transition rather than all-at-once.

flavour:
ok, so I think we're getting ready for a comms to mailing list on a plan...I don't think it needs a BluePrint:
(1) We are moving private/templates to modules/templates so that we can import template-specific modules without needing to use exec => can cache
(2) We are restructuring config.py as a module so that it can be cached & doesn't need to be compiled every request
(3) We will migrate all existing templates in trunk
(4) Non-trunk templates will still work fine through a legacy-support option, although we recommend that they are migrated to the new location/structure (we will provide docs on this)
(5) We will update Docs (Wiki / Book / Slides)
- assistance welcomed :)

Does that look right?

All we then need is a timetable...I suggest we make the change 48 hours after the communication.
I can merge the core code from you, along with a template of your choice to prove it, & then migrate the templates before I push to GitHub (I know I don't need to do /all/ of them but I would rather do so).

F

nursix:
Fine by me - I'll do the core code, and you can expect me to do default and skeleton to document the intended pattern, and potentially EVASS for testing (and as downstream-support). 

48 hours seems fine for me, gives me some time to test it thoroughly and iron out some path dependencies (e.g. prepop, views).

But now you can certainly merge this one now, right? ;) before we forget about it accidentally.

nursix:
@flavour: ok, core code complete - ready to start the migration whenever you're ready.

flavour:
I put the email notification out today. Let me merge you tomorrow morning & work on migrating modules/updating docs through the day

nursix:
Ok!

PR #1019, opened 2014-12-31
Happy New Year :D

hitesh96db:


hitesh96db:
So, Should those be removed ?, I guess that's better, no need to update it every year ;)

nursix:
No no, all good. I wasn't that serious ;)

PR #1020, opened 2015-01-01
[ToDo]: move is_affiliated()

raj454raj:


modules/s3menus.py
```diff
@@ -1439,7 +1439,7 @@ def org():
 
         return M(c="org")(
                     M("Organizations", f="organisation")(
-                        M("Create Organization", m="create"),
+                        M("Create", m="create"),
```
raj454raj: Just to maintain uniformity in all the modules.

modules/s3/s3aaa.py
```diff
@@ -5328,6 +5328,28 @@ def get_owners(self, table, record):
             owner_user = record["owned_by_user"]
         return (realm_entity, owner_group, owner_user)
 
+    # -----------------------------------------------------------------------------
+    def is_affiliated(self):
```
nursix: definitely the wrong place - nothing to do whatsoever with permissions.

Why move it at all? It is only used in req and nowhere else, so it seems specific. And even if not - it should either go into org (most logical to me) or into AuthS3 (still logical), but certainly not S3Permission.

nursix: To be honest, I would leave this in req, and remove the todo. It is only used in the req controller and nowhere else - moving it into a different place will only make debugging harder for no reason.

raj454raj: Yeah didn't made sense to me either but as the Todo suggested. Okay I'll remove Todo.

nursix:
ok :+1:

nursix:
Sometimes (I said that before) a ToDo is a sign for an immature idea - and hence it should make sense to you before you action it. The general rule in open source programming is "change the code if you can improve it" - not change it because you can change it ;)

raj454raj:
Sorry for that :p , I actually thought I'll ask flavour for that on IRC but he wasn't around.

nursix:
I am a bit uncertain about the is_affiliated-extension of the business rule here: what does it mean? Do we only check permitted facilities if the user is affiliated with an organisation? Can't quite make sense of it yet. An explanation would be nice.

My feeling is that there must have been a reason why it had not been done before despite available - if it would have been /that/ simple, then it wouldn't have been a ToDo.

raj454raj:
Okay one doubt: mpr module doesnt have its table definitions in s3db, is it not intended to make one or the module's not in use.

nursix:
mpr is an extra perspective of dvi, not a module of its own.

raj454raj:
Ok. :+1:

nursix:
i.e. it doesn't have its own model, but uses pr and dvi models.

Generally, you can consider controller files as "perspectives" (or use-cases) rather than as "modules" - Eden isn't really modular in that sense. Some models have multiple perspectives, e.g. doc and gis are ubiquitous, supply is used both for req and inv (but has no independent use-case), stats is a meta-model, as is sit. 

And there are also perspectives which extend others, and do not necessarily need any additional model of their own: mpr extends dvi (no extra model needed), vol extends hrm (minimal extra model), ...

raj454raj:
Thanks.

nursix:
NB Are you GCI contestant, or Sahana volunteer? Just wondering why you're trying to action arbitrary ToDo's.

raj454raj:
Sahana Volunteer.

controllers/req.py
```diff
@@ -468,9 +467,10 @@ def prep(r):
             elif type == 3: # People
                 # Limit site_id to orgs the user has permissions for
                 # @ToDo: Make this customisable between Site/Org
-                # @ToDo: is_affiliated()
-                auth.permitted_facilities(table=r.table,
-                                          error_msg=T("You do not have permission for any facility to make a commitment."))
+                if is_affiliated():
```
flavour: I don't think this is what is meant by the ToDo...just hiding the permission check isn't valuable.
I think the issue here is that *once Commitments can be made by an Org rather than a Facility* then is_affiliated can be used to determine whether a user has the ability to commit on behalf of an org or not
The first part is the harder part since the Requests module is currently tied to all Requests/Commitments happening at the Site level...in order to support Inventory/Shipments...but these don't apply as much to Person commitements (people will often travel directly from their homes, even if to a staging area before actual deployment. The staging area is only sometimes their office)

modules/s3menus.py
```diff
@@ -1439,7 +1439,7 @@ def org():
 
         return M(c="org")(
                     M("Organizations", f="organisation")(
-                        M("Create Organization", m="create"),
```
flavour: +1 to this for consistency.
Some templates may decide to be explicit about the create links, but it should be consistent either way.
(Personally I like to reduce the 'noise' of redundant information, but this does depend to some extent on the menu styling & structure as to how explicit we need to be here)

flavour: Done this in another commit...too small to justify a PR ;)

raj454raj: Sure thanks :+1:

PR #1021, opened 2015-01-02
ToDos in dc module

samsruti:
Since this file is not updated in edengci repo, so i am sending a pr to the main repo
Its for GCI task : http://www.google-melange.com/gci/task/view/google/gci2014/5346487999397888

### DIFF
--- a/modules/s3db/dc.py
+++ b/modules/s3db/dc.py
@@ -267,13 +267,15 @@ class DataCollectionModel(S3Model):
         define_table(tablename,
                      self.super_link("doc_id", "doc_entity"),
                      self.dc_template_id(),
-                     # @todo: default to now
-                     s3_date(),
+                     s3_date(
+                             default = "now"
+                            ),
                      self.gis_location_id(),
                      # @todo: default to user root-org
                      self.org_organisation_id(),
-                     # @todo: default to logged-in user
-                     self.pr_person_id(),
+                     self.pr_person_id(
+                                       default = current.auth.s3_logged_in_person(),
+                                      ),
                      s3_comments(),
                      *s3_meta_fields())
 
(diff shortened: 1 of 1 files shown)


nursix:
Sorry, but this is nonsense - the "answer" field needs to be developed to store multi-type data.

I am currently working on this module, which is why these are "ToDos" - but they are to myself, not meant to be actioned. See status-tag: WIP

samsruti:
Oops :( Sorry.. I wasnt completely sure about that .

modules/s3db/dc.py
```diff
@@ -301,9 +303,8 @@ def model(self):
             msg_record_deleted = T("Data Collection deleted"),
             msg_list_empty = T("No Data Collections currently registered"))
 
-        # @todo: representation including template name, location and date
         represent = S3Represent(lookup=tablename,
-                                fields=["date"],
+                                fields=["template_id","location_id","date" ],
```
nursix: Ha ha - of course not the template_id and location_id. That doesn't tell the user anything.

nursix:
I would request this to be left alone for now - I'll remove the status tag once I'm done with this.
Fine if you want to experiment with it, though (which is why it is shared in trunk)

samsruti:
Sorry Sorry , I m trying some other todo

nursix:
No need to apologize - it's not obvious.
But it's an incomplete implementation that needs development - no just "fixes".
I wasn't quite sure whether "answer" should become JSON, but that's the most likely solution - I left it a ToDo, so that I can find it more easily. However, there's a lot more to do here than the ToDos alone - this needs a widget, a representation and reporting capability - so your "fix" isn't really helpful, even if with good intention.
But if you want to experiment with XForms, then this is certainly something you can work with - input and feedback welcome.

nursix:
Tip: best you start with ToDos where you are certain about the meaning.

Sometimes (or rather often) they are not meant to be executed "by the word", but actually require more in-depth thinking and development. If it was that easy, then it wouldn't have been a ToDo ;)

samsruti:
Thanx @nursix  for your suggestions . I will surely implement that :+1:  :D

nursix:
Implement what? :O

samsruti:
Implement your suggestions :P

PR #1023, opened 2015-01-03
Minor changes with pdf codec and a small eyeball test for planned enhancment

graeme-f:
I plan to re-factor the pdf codec so that styling will be easier. To ensure that the change can be properly tested I have added this test along with two sample files from the test. Currently the test will consist of manually verifying the file created matched the file on the repository. Later an automatic diff could be added for this test.

### DIFF
--- a/modules/s3/s3codecs/pdf.py
+++ b/modules/s3/s3codecs/pdf.py
@@ -259,8 +259,7 @@ class S3RL_PDF(S3Codec):
             response.headers["Content-Type"] = contenttype(".pdf")
             response.headers["Content-disposition"] = disposition
 
-        doc.output.seek(0)
-        return doc.output.read()
+        return doc.output.getvalue()
 
     # -------------------------------------------------------------------------
     def get_html_flowable(self, rules, printable_width):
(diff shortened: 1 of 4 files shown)


flavour:
Sounds good :)

graeme-f:
I wasn't certain if they should be on the repo or not. My thought was that putting it there gives a record of what was produced before any changes, obviously I can do that locally but any record of what was produced before any code changed would be lost and so it would make it harder for others to check the changes. So by putting them in the repo I'm trying to make it easier for reviewers but maybe that is not necessary.

flavour:
The PDF files are small enough to include if you wish to have them there and are text not binary, so will diff nicely as-changed.

graeme-f:
Well the PDF is not a true text file, the contents are hidden in an embedded stream, https://github.com/flavour/eden/pull/1023/files#diff-190e5e45744b7f646a827f4c2f96d217R68.
The generator adds a timestamp here: https://github.com/flavour/eden/pull/1023/files#diff-190e5e45744b7f646a827f4c2f96d217R103 and a unique id here: https://github.com/flavour/eden/pull/1023/files#diff-190e5e45744b7f646a827f4c2f96d217R149. All of  which makes an auto compare between the old file and the new non-trivial.

So I might be putting forth arguments against putting it in the repo.

graeme-f:
The separator was not added. The hard-coded '/' was replaced with os.path.sep; which I felt was safer - It was failing when run in Windows. Until I could invest more time, which I would like to do, I didn't want to rewrite this function. To properly check any changes to this function I would need to extend the test, and run it on both Windows and Linux.

nursix:
Just saying that adding OS-specific separators is the actual point of os.path.join ;) so no need to add any separator to the path components ;)

graeme-f:
Hi Fran,

I know that Dominic raised a valid concern but I wasn't planning on any
more changes now. Rather get the tests in place and then I can restructure
the code and as part of that process I will implement and test that change.
So I am happy for the code to be merged. I will then work on the
restructuring as and when time permits (school is back so I need to get
into the swing of that).

Cheers,

Graeme.

PR #1024, opened 2015-01-03
[ToDo]: Link added for fire station corresponding to the logged in user

raj454raj:


views/_popup.html
```diff
@@ -1,4 +1,4 @@
-<div id='popup' class='cf'>
+<div id='popup' class='cf' style='display: table;'>
```
raj454raj: http://demo.eden.sahanafoundation.org/eden/pr/person/93/contacts
Add emergency contact.

views/fire/index.html
```diff
@@ -1,8 +1,13 @@
 {{extend "layout.html"}}
 <div id='home'>
-{{=H2(module_name)}}
+<h2>{{=T("Fire")}}</h2>
 <p>
 {{=T("The Fire Stations Module can be used to record information on Fire Stations.")}}
 </p>
+{{if station_id:}}
+<ul>
+  <li><a href='{{=URL(f="station", args=[station_id, "update"])}}'>Open Fire Station</a></li>
```
nursix: Should probably not be "Open Fire Station", but the name of the particular station - and if at all, then T()

views/_popup.html
nursix: How does that affect other popups? This is by far not the only use-case.
And *if* then this should not be an inline-style but probably widget.css.

views/fire/index.html
raj454raj: Name of particular station :+1:

controllers/fire.py
```diff
@@ -10,10 +10,14 @@
 def index():
     """ Module Homepage """
 
-    # @todo: have a link to the fire station the current user works at
+    response.title = T("Fire")
 
-    module_name = settings.modules[module].name_nice
-    response.title = module_name
+    table = s3db.fire_shift_staff
+    person_id = auth.s3_logged_in_human_resource()
```
nursix: Don't call it person_id if it is a human_resource_id.

nursix: You are not checking whether this returns anything - what if the user isn't logged in?

nursix: ...or not a fire staff member?

controllers/fire.py
```diff
@@ -10,10 +10,14 @@
...
+    rows = db(table.human_resource_id == person_id).select(table.station_id,
+                                                           limitby = (0, 1),
+                                                           )
+    station_id = rows[0].station_id
```
nursix: Use rows.first() not rows[0] - and make sure you check whether this returns any rows. If the logged-in user is not a fire staff member, then this will crash.

controllers/fire.py
raj454raj: For this if station_id will be None and the link wont be shown according to the if condition in views

views/fire/index.html
```diff
@@ -1,8 +1,13 @@
 {{extend "layout.html"}}
 <div id='home'>
-{{=H2(module_name)}}
+<h2>{{=T("Fire")}}</h2>
```
nursix: Why change this?
Now it needs to be changed in multiple places (both the response.title in the controller, and the title in the view) - which makes it harder to customize. Actually the module_name_nice is a deployment setting, so shouldn't be hardcoded.

controllers/fire.py
```diff
@@ -10,10 +10,14 @@
 def index():
     """ Module Homepage """
 
-    # @todo: have a link to the fire station the current user works at
+    response.title = T("Fire")
```
nursix: Why remove the deployment-setting lookup of the module name? (see the other comment on the view)

controllers/fire.py
nursix: But station name will not be None - it will not return a row, and hence crash ;)

controllers/fire.py
raj454raj: :+1:

views/fire/index.html
nursix: Besides: avoid hardcoding the "update" method here - just open the record, so CRUD will decide whether update or read depending on permissions. Not all fire staff members are automatically permitted to update the station record ;)

controllers/fire.py
raj454raj: :+1: ;)

views/fire/index.html
raj454raj: :+1:

views/fire/index.html
raj454raj: Okay considering that, somehow got confused between two labels ;)

raj454raj:
@nursix : done.

raj454raj:
@nursix : done.
Doubt: Whats the difference in these three --> 
1. using limitby(0, 1) and rows[0]
2. using .first()
3. using both limitby(0, 1) and .first()
Thanks.

nursix:
To answer the question:

rows[0] is the same as first() except that it is not fail-safe. If no rows are returned, rows[0] will crash, while first() will return None. So, either you explicitly check len(rows) before using rows[0] - or you use first() (recommended).

Secondly, if you intend to only ever use the first result, you should use limitby=(0, 1) - for efficiency. Otherwise the database will try to find and return /all/ rows, unnecessarily. 

The tightest bottleneck herein is the DAL converting the DB output into Python objects - this takes a lot of processing since values need to be parsed and converted, so any unnecessary row extraction should be avoided.

And as a more general efficiency/performance guideline: never handle any data that you don't need, only ever do just as much as is absolutely required for the operation.

nursix:
That ^ is btw also a /security/ guideline ;) (not touching anything that you don't need)

raj454raj:
@flavour : Can you please merge this.

nursix:
Yes - he can, just give him a few more hours. If you're looking for GCI points, consider this done.

raj454raj:
Well @nursix , I am not a GCI student ;)

nursix:
...ok ok, bear with me - just feeling stalked by this sometimes ;) no offence intended

PR #1025, opened 2015-01-06
[Fix]: Representation of item name in autocomplete widget

raj454raj:
http://127.0.0.1:8000/eden/req/req/1/req_item --> Add Item to request , Autocomplete for Item field not working before. (Giving link of localhost because demo is not working ;) )

### DIFF
--- a/modules/s3/s3widgets.py
+++ b/modules/s3/s3widgets.py
@@ -7440,9 +7440,9 @@ def search_ac(r, **attr):
         output = []
         append = output.append
         for row in rows:
-            record = dict(id = row.id,
-                          label = row[fieldname],
-                          )
+            record = {"id": row.id,
+                      fieldname: row[fieldname],
+                      }
             append(record)
 
     current.response.headers["Content-Type"] = "application/json"


nursix:
Fix at the wrong end - this was just changed into fieldname as "label" isn't standard. Please fix the corresponding controller.

nursix:
If you change it back into "label", it will not work with e.g. org_site_search_ac and org_search_ac. So, either you fix /all/ search_ac's to use "label" (then you can hardcode it) - or you fix the particular search_ac that uses "label" to use the fieldname.

To me, fieldname is more consistent and less fragile.

nursix:
...less fragile because the ac handler allows to extract additional fields - and hardcoding anything would allow name collisions, whereas fieldname would at most "collide" with itself ;)

nursix:
Like this: https://github.com/nursix/eden/commit/26f4d3a7a68844321510f29b7a38a3f70d46ffd9

raj454raj:
Okay :+1:

nursix:
I think the problem here is the inconsistency between all the different search_ac's (implemented by different people). Using the fieldname is more or less intuitive, and obviously how most of the devs interpret the overall concept - so let's just stick with this for consistency reasons.

Thanks however for pointing to the issue, I had stumbled over this some time ago, but apparently didn't see the whole picture then.

nursix:
Shall I revoke my commit, so you can update this PR?

nursix:
...this would give you a trunk commit ;)

raj454raj:
@nursix : done.

nursix:
Err...not quite the same thing, is it?

Should be like:

    record = {"id": row.id,
                   fieldname: row[fieldname],
                   }

...and remove the JS changes.

raj454raj:
Hold on: This time I need to apologize sorry.

nursix:
Apologizing again? :D

raj454raj:
@nursix :done.

nursix:
Good :+1: ready to merge.

PR #1026, opened 2015-01-06
Subscriptions config

hitesh96db:
Allow users to change the Subscription settings, i.e, frequency and e-mail format.


modules/templates/SSF/controllers.py
```diff
@@ -149,4 +151,81 @@ def __call__(self):
...
+
+        user = current.auth.user.pe_id
+        stable = current.s3db.pr_subscription
+        formstyle = current.deployment_settings.get_ui_formstyle()
+        query = (stable.pe_id == user) & \
+                (stable.deleted != True)
+        row = db(query).select(stable.id,   
+                               stable.frequency,
+                               stable.email_format,
+                               limitby=(0, 1))
```
nursix: Forgot the .first()?

modules/templates/SSF/config.py
```diff
@@ -1123,6 +1125,111 @@ def custom_prep(r):
...
+    def custom_render(resource, data, meta_data, format=None):
+        """
+            Custom Method to pre-render the contents for the message template
+
+            @param resource: the S3Resource
+            @param data: the data returned from S3Resource.select
+            @param meta_data: the meta data for the notification
+            @param format: the contents format ("text" or "html")
+        """
+        rows = data["rows"]
```
nursix: data.rows is fine

modules/templates/SSF/config.py
```diff
@@ -1123,6 +1125,111 @@ def custom_prep(r):
...
+            rfields = data["rfields"]
+            output = {}
+            new, upd = [], [] 
+            
+            # Standard text format
+            labels = []
+            append = labels.append
+
+            for rfield in rfields:
+                if rfield.selector == created_on_selector:
```
nursix: This is a bit too much I'd say - colname for "created_on" is always <tablename>.created_on ;) really no need to introspect it.

nursix: No really :D sorry - but now you're getting carried away a bit :P Technically excellent, but completely unnecessary.

|grin|

nursix:
Looks fine to me :+1:  good to merge.

nursix:
But out of curiosity - is there prior art for this slightly-over-the-top inteospection? If so, then it should perhaps be defused a bit. It's a bit like putting on a hat with a crane ;)

modules/templates/SSF/controllers.py
hitesh96db: Is there a need for it ?, I'll be checking for a row existance anyways, so I could just use row[0] without trying to catch an error.

modules/templates/SSF/config.py
hitesh96db: Prior art :), I'll change it...

modules/templates/SSF/controllers.py
nursix: If you add .first() to the select, then you can just use row instead of row[0] - using row[0] looks a bit cumbersome, and is somewhat inconsistent with the standard way of doing it.

modules/templates/SSF/config.py
nursix: Prior art? Where? :D

modules/templates/SSF/controllers.py
nursix: Wants to say: the standard way of doing it is:

    row = db(query).select(...., limitby=(0, 1)).first()
    if row:
        value1 = row.field1
        value2 = row.field2

If you do instead:

    row = db(query).select(..., limitby=(0, 1))
    if row:
        value1 = row[0].field1
        value2 = row[0].field2

...then you repeat the index lookup, and probably confuse others since "row" doesn't quite suggest that it is a Rows instead and not a Row.

modules/templates/SSF/config.py
hitesh96db: https://github.com/flavour/eden/blob/master/modules/s3/s3notify.py#L562

hitesh96db: So I'll change that to "~.created_on"

nursix: I actually meant you can assume that created_on_colname will always be "<tablename>.created_on", so you do not need to find it by using "~.created_on" as selector. Just define it as:

    created_on_colname = "%s.created_on" % resource.tablename

No? No Introspection required.

nursix: But I appreciate that GitHub comments has hidden away the "tablename" :D so you probably misunderstood my comment. I try again Python-style: "%(tablename)s.created_on"

modules/templates/SSF/config.py
```diff
@@ -1123,6 +1125,111 @@ def custom_prep(r):
...
+            @param resource: the S3Resource
+            @param data: the data returned from S3Resource.select
+            @param meta_data: the meta data for the notification
+            @param format: the contents format ("text" or "html")
+        """
+        rows = data["rows"]
+ 
+        if format == "text":
+            
+            created_on_selector = resource.prefix_selector("created_on")
```
nursix: Here you define the created_on_selector, but not the created_on_colname. Yet you could forget the selector since at this point you already know the colname!

    created_on_colname = "%s.created_on" % resource.tablename

No need to find it by scanning the rfields. Instead you can later skip this rfield by checking:

    if rfield.ftype != "id" and rfield.colname != created_on_colname:

;) That's what I meant. A tiny little bit too much introspection trying to find the created_on field by it's selector instead of the (constant) column name.

modules/templates/SSF/config.py
hitesh96db: Will "created_on" be in every table ? I guess the reason it was put there was to consider new records created as well, so only if a "created_on" field is present can we say something about new records...
 That's what I understand from this: https://github.com/flavour/eden/blob/master/modules/s3/s3notify.py#L571

modules/templates/SSF/config.py
```diff
@@ -1123,6 +1125,111 @@ def custom_prep(r):
...
+            last_check_time = meta_data["last_check_time"]
+            rfields = data["rfields"]
+            output = {}
+            new, upd = [], [] 
+            
+            # Standard text format
+            labels = []
+            append = labels.append
+
+            for rfield in rfields:
```
nursix: Also: here you loop over the rfields to extract the labels and colnames, and further down you loop over the result again to extract the values from the rows. I don't think you need to do it like that - instead just change the second loop to loop over rfields directly.

Surely you would have the "rfield.colname" lookup inside the inner loop - but then that isn't really slower than the tuple extraction in the "for", and eventually you save a whole loop by not running this extra preparation loop. And you make it a lot less complicated ;) so easier to debug to understand for others.

nursix: Just mentioning that - but it's not in the way for a merge, of course.

Generally, striving after the simplest possible way to do something is good (and in Python, the most intuitive way is also often the fastest). 

Complications may not harm functionality - but this is open source code, so it's always good to think of others who may want work with your code later, and make life easier for them.

Also - remember that people could (and will!) copy your style. If you make it complicated, so will others - but then it could lash back against you when you try to debug their code ;) Hence: if you have a choice, then "simpler" is "better".

hitesh96db: :+1:

modules/templates/SSF/config.py
nursix: That is true - even if created_on is in the table, it may not be in rfields (and hence not in the rows either), and then you can not distinguish new and updated records. That's a bit the meaning behind this.

But still doesn't create a need to introspect the column name of "created_on". If there is a created_on field, it will always have the column name "%(tablename)s.created_on", so this can be simplified.

Also note that generic code sometimes needs to do a little more introspection so that it can be used for varying use-cases - but in specific use-cases you can make assumptions. I think in this case it's just fine to assume that created_on is in the rfields.

I'd say - leave it as you have it ;) this should not hold up the merge. Technically it's totally fine.

nursix: I see that created_on_colname is used as a flag for whether the created_on field is among the rfields, so that this doesn't need to be checked for every row separately. I think this idea is good in general - but it isn't too obvious and has a bit of efficiency side-effects (especially since it loops over all the fields).

You could do it like this instead:

    created_on_colname = "%s.created_on" % resource.tablename
    has_created_on = any(rfield.colname == created_on_colname for rfield in rfields)

...and then use:

    if has_created_on:

instead. That makes it a bit more obvious - and it exits early yet provides the flag this needs. Then have the inner loop like:

    for rfield in rfields:
        if rfield.colname != created_on_colname:
            append_column(rfield.label, row[rfield.colname])

nursix: But as I said: it's only about style, and whether or not people can easily understand the code. Technically, your solution is totally correct and acceptable.
I also appreciate that you copied this from a generic case - and that this generic case is a bit over the top. So, not your "fault" ;) but of course something you /could/ improve. If you wanted to and had time to waste...

hitesh96db: I'll leave this for my next PR ;)

nursix: Of course, np.

PR #1027, opened 2015-01-07
start_date and end_date widget working for project/activity

raj454raj:
I still think the fieldname of project_activity should be changed to start_date instead of date(consistency!)

### DIFF
--- a/controllers/project.py
+++ b/controllers/project.py
@@ -548,6 +548,15 @@ def activity():
         return True
     s3.prep = prep
 
+    def postp(r, output):
+        if r.interactive:
+            if not r.component:
+                s3.jquery_ready.append(
+'''S3.start_end_date('project_activity_date','project_activity_end_date')''')
+
+        return output
+    s3.postp = postp
+
     return s3_rest_controller(csv_template = "activity",
                               hide_filter = False,
                               rheader = s3db.project_rheader,


controllers/project.py
```diff
@@ -548,6 +548,15 @@ def prep(r):
         return True
     s3.prep = prep
 
+    def postp(r, output):
+        if r.interactive:
+            if not r.component:
+                s3.jquery_ready.append(
+'''S3.start_end_date('project_activity_date','project_activity_end_date')''')
```
nursix: Ideally, the injection would look like:

    $('#project_activity_end_date').selectInterval({startDate: 'project_activity_date',intervals: [6,12,24,60],default:12,selectOnLoad:true});

(well - just making up options here, but these are the options I would see as necessary).

No S3.*, no globals, just a local instance on that particular pair of fields. To achieve that, make it a jQuery function, or even better a jQuery UI widget (using the widget factory, very easy to do).

Should be injected from the S3DateWidget, so that it can be re-used for other cases, and would even work inline. Selectors should be introspected (so that it automatically uses the inline prefix), rather than hardcoded.

The JS seems to have gone completely now? Maybe you forgot to add it?

raj454raj: start_end_date function is used for that disabling some dates. Infact this is used everywhere else why need a seperate widget (when we have a method that already does that).
Please Correct me.

nursix: Making it a datepicker "plugin" would allow this to be injected like:

    $('#project_activity_end_date').datePicker({...options...})
                                   .selectInterval({...options...});

If you make it a jQuery UI widget, you'll be able to access the context element as:

    var inputField = $(this.element);

...which makes it very easy to inspect it. And you can access the corresponding datepicker instance like:

    var instance = $(this.element).datepicker('instance');

...e.g. to get at the namespaceID, manipulate options and created elements, etc...

All the local data would be properly encapsulated, so you can have any number of "selectInterval" instances in the same form (which is required for inline forms and deduplicator).

raj454raj: @nursix : More uniform would be making the fieldname of project_activity as 'start_date' (Why keep only the label 'Start Date').
Anyway I still didn't understand why you dont want to call start_end_date similar to this:
https://github.com/flavour/eden/blob/master/controllers/project.py#L246
> Maybe you forgot to add it?

Please explain.

nursix:
Example of a jQuery UI widget using widget factory:
https://github.com/nursix/eden/blob/master/static/scripts/S3/s3.ui.hierarchicalopts.js#L12

...and how it is injected:
https://github.com/nursix/eden/blob/master/modules/s3/s3widgets.py#L6285

There are many more examples in static/scripts/S3 and static/scripts/ui - some are less, some are more complex. Datepicker itself is a UI widget. The principle is always the same, so there's plenty of examples...

nursix:
Example of a jQuery function (not a UI widget, but still well encapsulated):
https://github.com/nursix/eden/blob/master/static/scripts/S3/S3.js#L1282

...and how it is injected:
https://github.com/nursix/eden/blob/master/modules/s3db/req.py#L638

This wouldn't have the this.element context, though - so the widget pattern seems better for this case.

raj454raj:
So all injections like these should be changed accordingly:
https://github.com/flavour/eden/blob/master/controllers/project.py#L246

controllers/project.py
nursix: So, this is unrelated to your other work on start/end date selection?

nursix: S3.start_end_date is certainly fine here - I just mistook this commit as related to your other work (very similar commit messages and names).

raj454raj: O okay actually you thought it was gauravmittal95's commit(https://github.com/flavour/eden/pull/1013)? So do I need to change anything now? @nursix

nursix: Really sorry - I mixed this up - was just looking at another PR with a very similar commit message (and topic), and got confused over the apparent "change in direction". Collateral damage, really - too many parallel threads, too much chasing.

Your commit is just fine and good to merge.

raj454raj: Okay thanks - well again I am asking this doubt shouldn't the fieldname of project_activity be 'start_date' instead of just 'date' (Anyway label is "Start Date") @nursix ?

nursix: The discussion about renaming "date" into "start_date" is not new, but here is what stopped it from happening:

There are /a lot/ of live production data that would need to be migrated, and /a lot/ of related XSLT and other scripts...so this is not "just" to rename the field - it would require a major operation, for very little benefit especially on the user side (where the effect of it is equal to zero).

raj454raj: True that! Okay :+1:

nursix: Hence: heads up when you first create a model. It's a bit like jumping from a skyscraper - after you've jumped, you can not "just" change your mind between the 48th and 47th floor and return to the roof.

raj454raj: HAHA ;). Since you are not on IRC I am troubling you here:
http://127.0.0.1:8000/eden/hrm/staff/summary --> menus -> search by skills
What is this for?
Something's really wrong with this --> It says create skill inside it gives another option for create skills, sometimes after filling the form it gives "record not found", sometimes insert takes place in hrm_competency. @nursix

raj454raj: Shouldn't "Competency" be made a different menu item and have options to create and search independently?

nursix: Post-production field name changes always cause major trouble - and can disrupt critical workflows. 

We had that once in a production system where suddenly the recipient fields in a messages table were all blank - thereby making the whole system completely useless. Luckily we had a backup so we could reconstruct the data, but the service was however down for a whole day.

You really need a better reason than code cosmetics to do that, and even if you have better reasons, it is usually better to introduce a new, independent field and leave the old field in place (e.g. hidden) for some time so that production systems have a chance to migrate without loosing critical data.

And you must adapt a lot of code, and have plenty of fallbacks to handle existing data in the old field (and issue deprecation warnings)...and so forth. A lot of things to take care of, not something you can do en-passant.

Most of the time, that's not worth it - especially since we have the ability to set arbitrary field labels.

nursix: Let me check this "search by skills" thing - I get back to you.

raj454raj: :+1:

nursix: Obviously it should say "Search by Competency" not "Search by Skills". It is meant to find a staff member with a certain competency level for a certain skill.

The inner "Create Skill" is to define a skill - the outer "Create Skill" is actually a "Create Competency", i.e. it assigns a competency level to a person for a particular skill.

The "Record not found" error is indeed weird - most likely that is related to a filter inside the controller. Do you want to investigate that or shall I?

raj454raj: Why does the insert takes place(shouldnt be right the option is search by competency).
How about adding a new menu item Competency and sub-options -- create and search

raj454raj: Well @nursix : I'll check and get back to you(Hopefully with a PR) :)

nursix: This is the problem:
https://github.com/flavour/eden/blob/master/controllers/hrm.py#L565

The "Person" selector allows you to select any kind of person - even if not type==1 (Staff). So you can add a competency record for a person of type==2 or no type at all. Then, after form submission, it would try to redirect to the competency record you just created, but since it is not of type 1 as the controller requires it, it can not be found, hence "Record not found".

Two possible ways to solve that:
1) Do not apply the filter if you have a record ID in the request (r.id)
2) Do not allow the selection of persons who are not type 1

IMHO, both should be done.

Secondly, we do not have pure "Search" pages - almost every "Search" page will also allow you to add a new record (standard). It may be useful to make the "Create" button a bit less prominent at this point (e.g. by moving it to the right, like done in CRMT), so that it is perceived as an additional option, and doesn't cause confusion - but completely removing it is not what we want to do.

Third: no - there should not be a separate competency menu item, because that is actually not a separate workflow. The standard workflow here is to add a competency on the "Skills" tab in the Staff record itself (compare these two forms and you will see the logic in it).

In fact, for this particular case, we may even want to remove the "Create" button from the search page - but because it would be invalid, or for UX reasons, but rather because it isn't the intended workflow. But it is beyond me to make that call - this is something the users should come up with.

I would fix the above two points for now, and then leave it at that.

raj454raj: Okay thanks :+1:

raj454raj:
@gauravmittal1995: ^ I think the above two comments will help you

controllers/project.py
raj454raj: @nursix : Can I make these changes?

nursix: Sure.
I will then put the Create-button in a nicer place. Like this:
ss22

raj454raj: Okay :+1: thanks :)

flavour: From the user-perspective, I think this is 'Search by Skill' - more people will understand this than 'Search by Competency'...of course internally where we're filtering a Person list by Skill, we'll do so using the hrm_competency link table.

raj454raj: How about "Search by Skill/Competency"? ;)

flavour: Seems noisy to me & an extra string to translate

raj454raj: :+1: okay. So I'll keep "Search by skill" in the menu and competency inside?

nursix: Absolutely.

A bit confusing it is to have a "Create Skill"  inside a "Create Skill", however - so how to make clear that in the first case it is the "Skill of a person", and the second case it is "the Skill" in general?

No better idea at this point, though - and neither the time nor the *skills* to fight with labels :D

nursix: Sorry, my comment was in reply to Fran, not your question. (leapfrogging a bit)

nursix: Suggest not to change any labels, but just fix the filtering issue for now as discussed.

raj454raj: Well just about to ask that. Okay!

nursix: E.g. using a different controller (S3PersonAutocompleteWidget takes "controller" as parameter) => "hrm" in the one case, and "vol" in the other. Or you allow S3PersonAutocompleteWidget to take an additional parameter so that it filters by HR type (optional).
I would go for the latter, but the former is very ok too.

nursix: Note that using a different controller (and then using that in pr_search_ac to determine what to filter for) may be a lot easier, but way less flexible.

Allowing the specification of URL filters for the Ajax URL would give us the ability to filter by *anything* we ever need, since pr_search_ac is RESTful and thus processing URL filters - including the ability to change these filters dynamically at the client side ;)

nursix: E.g. selecting a driver for a fire engine => only show people who have a driver license... (just making it up, but doesn't look too silly).
If you can't add this to the autocomplete widget, then I would - it's a fairly small change for a BIG effect.

nursix: NB if you add a URL filter to the widget, nothing needs to be changed in pr_search_ac ;) URL filters are processed by the REST API before the search_ac is called.

gauravmittal1995:
@nursix  and @raj454raj : Thanks :)

PR #1028, opened 2015-01-07
Showing only required rows in dropdown for search by skills

raj454raj:
@nursix : done.


nursix:
Ouch - that's what I call an ugly hack ;)

Seriously?

nursix:
...after all the pointers I gave you :( bit disappointing. You can well better than that.

static/scripts/S3/s3.autocomplete.js
```diff
@@ -585,7 +593,8 @@
                 $.ajax({
                     url: real_input.data('url'),
                     data: {
-                        term: request.term
+                        term: request.term,
+                        cont: controller
```
nursix: Instead of adding controller to the data, you could add "&~.human_resource.type=1" to the url, and remove the server-side processing of the "cont" option.

nursix: Even better would be if you allow the URL filter to be passed in from the widget. Then we are free to define whatever we need, wherever we need it.

raj454raj: :+1:

raj454raj: @nursix : I'll do as you told. I need to understand filters properly. Thanks a lot for your help.

nursix: Well, 

the general flow of RESTful request processing in Eden is:

    client => controller => S3Request (via s3_rest_controller) => method handler

The S3Request interface creates a filtered S3Resource instance for the requested URL, e.g. a request to pr/person generates an instance of the pr_person resource. 

By default, this resource is filtered by accessibility (only the records which are not deleted) and by permissions (only the records the user is permitted to access) - and you can specify additional filters in the request URL, e.g. you could specify:

    /pr/person?~.first_name=Dominic

...to filter only for person records where the first name is "Dominic". Multiple filters are possible, and you can specify filters along any thinkable join axis, e.g. components, link tables, foreign keys.

The filtered resource is then passed on to the method handler. In this case, the method handler is pr_search_ac, and it has the signature:

    pr_search_ac(r, **attr)

The "r" parameter is the S3Request instance, and r.resource is the filtered S3Resource instance generated by this request. When you select from this filtered resource like:

    data = resource.select(...)

...then this select will apply all the filters (accessibility, permissions, URL filters) *plus* the filter that is set by pr_search_ac itself.

So, if you pass in the additional filter for "human_resource.type" via the URL, there is nothing about that you need to process in pr_search_ac - r.resource already has it when it arrives in pr_search_ac.

If you pass this URL filter to the widget, e.g. by setting:

    field = s3db.hrm_competency.person_id
    field.widget = S3PersonAutocompleteWidget(ajax_filter="~.human_resource.type=1")

...and then let S3PersonAutocompleteWidget pass this filter on to the s3.autocomplete.js script, and let the autocomplete script add it to the Ajax URL that calls back to pr_search_ac, then you have a perfect way to specify additional filters for (person-)autocompletes.

Of course, the vol/competency controller would specify a different filter ;)

But with that, you can do more than filtering by "human_resource.type=1", e.g. you can also specify things like "human_resource.organisation_id$root_org=3" (to filter by members of the same organisation as the logged-in user), ...or whatever else you might need.

NB the whole filter form framework works this way - by adding filter queries to Ajax URLs. E.g. when you type something into the filter form in the org/organisation page, then the filter script will add it as a URL filter to the Ajax URL of the datatable, and then just Ajax-refresh that datatable.

The filter form can handle multiple "targets", e.g. in summary pages: you can have a datatable, a map and a pivot table report - and the filter form would update all their Ajax URLs at once, so that you can switch between the different views without loosing your filters.

URL filters work for all URLs in a REST controller - including XML exports - so its a universal framework.

raj454raj: Well thanks a lot @nursix , seriously this encouraged me. Thanks. I will send the changes tomorrow(If you dont mind)

raj454raj:
@nursix : done.

raj454raj:
Yes!

nursix:
Yes, thanks.

@flavour: good to merge.

PR #1029, opened 2015-01-08
DRMP one line bug fix and NYC menu nav bar css fix

samsruti:
The CSS fix for NYC template.
Browser : Google Chrome (Zoom level set to default)
OS : Windows 7
Screenshot:
Before the fix:
nyc

After the fix:
nyc_after

### DIFF
--- a/modules/templates/DRMP/config.py
+++ b/modules/templates/DRMP/config.py
@@ -14,7 +14,7 @@ from gluon.html import *
 from gluon.storage import Storage
 from gluon.validators import IS_EMPTY_OR, IS_NOT_EMPTY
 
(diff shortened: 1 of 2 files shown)


flavour:
The CSS fix here seems to be needed with Admins using small-width screens - as this is happening for you because there are too many top-level menu items for the width of your screen.& so it wraps.
This issue wouldn't happen for nopn-Admins (as they have less top-level menu items) or for users with wider screens.
Other (normal) users will see a negative from this change as the menu buttons will now be smaller/less easy to read.
The whole NYC template is being revised imminently anyway with less modules activated & the theme will restart from default (i.e. Foundation) so I don't see it as critical to fix.

Thanks for the DRMP catch, which I applied manually :)

PR #1031, opened 2015-01-09
Re-write using S3Method handler

raj454raj:
I'll complete another 3 re-writes once you have your views on this!
This function takes care of download button in this : http://demo.eden.sahanafoundation.org/eden/survey/template/1/translate

### DIFF
--- a/controllers/survey.py
+++ b/controllers/survey.py
@@ -110,9 +110,11 @@ def template():
                 s3_action_buttons(r)
                 s3.actions.append(dict(label=str(T("Download")),
                                        _class="action-btn",
-                                       url=URL(c=module,
-                                               f="templateTranslateDownload",
-                                               args=["[id]"])
+                                       url=r.url(method = "translate_download",
+                                                 component = "translate",
+                                                 component_id = "[id]",
+                                                 representation = "xls",
+                                                 )
                                        ),
                                   )
                 s3.actions.append(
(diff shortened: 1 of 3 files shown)


raj454raj:
This function takes care of download button in this : http://demo.eden.sahanafoundation.org/eden/survey/template/1/translate

nursix:
...or actually survey_TranslateDownload ;) I mistyped.

To explain the above: the difference is that now the method runs through the survey/template controller instead of a separate survey/translate controller ;) No additional lookup of the master record required, nor an additional ACL for the other controller.

Further, the method is bound specifically to this combination of records (template<=>translation) - so it would be possible to restrict access to translations of restricted (e.g. unpublished/unapproved) templates - without any additional controller logic.

nursix:
That has wider implications, btw - the survey module doesn't really implement this, but normally you wouldn't want half-ready or obsolete templates to be downloaded and used (e.g. via mobile apps), so this requires a logic to restrict public access.

The new survey module "dc" which I am currently working on does implement a logic to suppress templates in public form lists (not in trunk yet), so that the mobile client doesn't see templates which are not ready for use yet, or not valid anymore.

And of course we do not want the translations of unapproved templates to be downloadable either.

controllers/survey.py
```diff
@@ -110,9 +110,9 @@ def postp(r, output):
                 s3_action_buttons(r)
                 s3.actions.append(dict(label=str(T("Download")),
                                        _class="action-btn",
-                                       url=URL(c=module,
-                                               f="templateTranslateDownload",
-                                               args=["[id]"])
+                                       url=r.url(method = "translate_download",
+                                                 component = "",
+                                                 id = template_id)
```
raj454raj: Can you explain the difference.

flavour: template_id = all hardcocded to that template
"[id]" = use the ID of that row
- this is an action button for a dataTable Row, so each row should have a different ID

raj454raj:
@nursix : 
```
r.url(method="translate_download", component = "", component_id = "[id]")
```
gives /eden/survey/template/2/translate_download and 
```
r.url(method="translate_download", component = "translate", component_id = "[id]")
```
/eden/survey/template/2/translate/%5Bid%5D/translate_download
Former is what we want right?
And in s3db, in survey_template model
```
set_method("survey", "template",
                   method = "translate_download",
                   action = survey_TranslateDownload,
                   )
```
There is no component attribute that set_method takes so that we can pass "translate" in it?

raj454raj:
@nursix : done.
As you said I can access the survey_translate record by self.record, but by that we can only get the record_id. So still one db lookup. Any suggestions?

nursix:
Yes, you're right - self.record is the record ID, not the record.

Reason for that is that the framework doesn't know which fields you need in the method handler, so rather than extracting everything, it just gives you the record ID. But at least that makes the lookup very simple - use self.record to extract the survey_translate record (be sure to check that it is not None before you run the lookup).

The main advantage here is that you run the method through the same controller as the master request, not that you "save" a db lookup.

raj454raj:
Yeah thanks. Any other modifications?

nursix:
No time at the moment to review - later.

raj454raj:
Sure! :+1:

raj454raj:
@nursix : Done for export_all_responses() +  some minor fixes

raj454raj:
So,
++ s3_unicode  and 
-- T.lazy = False ;) Thanks

raj454raj:
```
sheet = book.add_sheet(T("Responses", lazy = False))
```
and 
```
sheet.write(row,col,s3_unicode(widgetObj.fullName()))
``` 
?
Will take care of names :+1:

raj454raj:
@nursix : done. :+1: . I will send another S3Method in next PR. I know this getting messed up to review.

raj454raj:
@nursix : Any other modifications?

raj454raj:
Yeah just a sec I'll send these small changes and for sure I'll be sending a PR for refactoring once I finish moving to S3Method. My next PRs would be two more S3Method in this module followed by a PR for  refactoring!!! :+1:

PR #1032, opened 2015-01-12
SSF Bug Fix

hitesh96db:
Filtering doesn't work with some fields here.
http://sunflower.sahanafoundation.org/eden/project/project/1/task 
For eg: milestone.

### DIFF
--- a/modules/templates/SSF/config.py
+++ b/modules/templates/SSF/config.py
@@ -844,10 +844,19 @@ def config(settings):
                                                                fields = [("", "project_id")],
                                                                multiple = False,
                                                                ))
+            
             crud_form = S3SQLCustomForm(*crud_fields)
-
+            get_config = s3db.get_config
+            list_fields = get_config(tablename, "list_fields")
+            filter_widgets = get_config(tablename, "filter_widgets")
+            # Remove 'Assigned to' field
+            list_fields.remove("pe_id")
+            custom_filter_widgets = [widget for widget in filter_widgets \
+                                                if "pe_id" not in widget.field ]
             s3db.configure(tablename,
                            crud_form = crud_form,
+                           list_fields = list_fields,
+                           filter_widgets = custom_filter_widgets,
                            )
 
     settings.customise_project_task_resource = customise_project_task_resource


flavour:
This will fix that URL but break the more common (at least for me) /project/task/x URLs
We need to detect function & change accordingly.
NB One of the changes was already in a function != project

nursix:
Ahm - it should actually work just fine for project/task/x? ("task" prefix is automatically detected as master if it matches the master alias).

hitesh96db:
Yes, ^ I tried that, it seems to work fine

nursix:
It should, because "~.selector" is equal to "alias.selector" is alias is master.

flavour:
ok, so I learnt something...but one of the fixes still seems wrong since it fixes for project/project/task yet is behind a function!= "project"?

nursix:
Question is though whether you really need to prefix it at all - normally, when you configure the filters for a table, it will always be the master table...so why the need for an explicit prefix here?

nursix:
Just *if* you prefix it, then ~.* and master.* or no prefix is all the same thing.

nursix:
But I still don't see the point of prefixing in a filter widget configuration (there is actually a reason why you do not need that - the table is defined by s3db.configure(tablename, ...))

hitesh96db:
But then filter forms won't work when viewed in components like project/project/id/task ?

hitesh96db:
Because these filters are meant to be resolved with the master "task" table and not "project"

nursix:
Ahm - actually not: filter widgets are resolved against the target resource, not the master resource - and should automatically be prefixed. This would only fail if the request goes to the wrong URL (i.e. ajax URL misconfigured).

nursix:
I'll check it. Seems like the wrong fix (but it shouldn't harm, however).

nursix:
Yeah - as I suspected it: your fix is "correct", but in the wrong place. 

What you do here should actually happen automatically - namely: we do pass in the component alias to the filter form so that it can prefix any non-prefixed selector in filter widgets correctly for the current component:

https://github.com/flavour/eden/blob/master/modules/s3/s3crud.py#L1180

But in this case of "task_milestone.milestone_id", the selector /does/ already have a prefix (="task_milestone"), so it was skipped - which is a bit insufficient logic: of course this must check whether it is the /right/ prefix, not just any prefix.

I pushed a fix for that.

nursix:
So, @flavour : no need to learn anything new ;) filter widgets do not need a component prefix.

flavour:
So this PR is deprecated?

hitesh96db:
No, wait...the other part, that is, removing the "Assigned to" field is required

hitesh96db:
I'll remove the "top" part of the commit, i.e, the "task" part

hitesh96db:
@flavour done

nursix:
Yes. deprecated after you merged the fix I pushed.

PR #1033, opened 2015-01-13
SSF Tweaks

hitesh96db:
- Prevent User from receiving notifications for his/her comments.
 - (Sunflower - 'Watch') Bug fix - I assumed the url to be of the form "task/id" but it can also be  "task?task.uid=urn:uuid=...."

### DIFF
--- a/static/themes/SSF/js/role.js
+++ b/static/themes/SSF/js/role.js
@@ -20,7 +20,7 @@ $(function (){
 
             url = S3.Ap.concat('/project/member.s3json');
             type = 'PUT';
-            if (task_id<=curl.length) {
+            if (task_id<curl.length) {
                 url += '?task_id=' + curl[task_id];
             }
             data_json = {'$_project_member':[{'$k_task_id':{'@resource':'project_task',
--- a/modules/templates/SSF/config.py
+++ b/modules/templates/SSF/config.py
@@ -549,6 +549,8 @@ def config(settings):
             After DB I/O for Project Member
             - Subscribe User to Task updates
         """
+        if task_id is None:
+            task_id = str(form.vars.task_id)
         sub = TaskSubscriptions()
         sub.add_task_subscription(task_id)
 
@@ -855,7 +857,6 @@ def config(settings):
                                                 if "pe_id" not in widget.field ]
             s3db.configure(tablename,
                            crud_form = crud_form,
-                           list_fields = list_fields,
                            filter_widgets = custom_filter_widgets,
                            )
 
@@ -1392,6 +1393,8 @@ class TaskSubscriptions(object):
                           label=current.T("Updates"))
                      ]
         self.rfilter = "comment.task_id__belongs"
+        # Remove comments created by the user
+        self.exclude = ["comment.created_by__ne", str(current.auth.user.id)] 
         # Get current subscription settings resp. from defaults
         self.subscription = self.get_subscription()
         subscription = self.subscription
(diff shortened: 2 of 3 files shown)


static/themes/SSF/js/role.js
```diff
@@ -20,7 +20,7 @@ $(function (){
```
flavour: Would be good to see a minified version of this file

hitesh96db:
@nursix Not related to this commit, But I need some clarification on this:
https://github.com/flavour/eden/blob/master/modules/s3/s3rest.py#L1482
the transform expects a full path ? or a relative path(i.e, from request.folder) ?
If it expects a full path(which I think it does), Is it good(or safe) to expose the directory structure publicly in the url, I think I've read that it isn't good to do so(security issues) ?

nursix:
The "transform" variable doesn't take a file path but a URL. This can be used to specify an external transformation stylesheet, i.e. one that is *not* located on the Sahana server.
For internal stylesheets, you should actually put the stylesheet into static/formats and then just use the respective format extension (like .odt, .have).

hitesh96db:
@nursix What if it needs to be template specific ? 
For eg: I want to add more fields in task rss feeds

nursix:
You can override r.XSLT_PATH in the prep, e.g.:

    if r.representation == "rss":
        r.XSLT_PATH = "modules/templates/ssf/rss"

...and put your custom stylesheets into that folder.

But at this point I don't quite see why it would be template-specific? What fields were you thinking of?

nursix:
(saying that because so far Sunflower is the only application using project/task.rss - so it could be a precedent)

hitesh96db:
right now, Task feeds only include the description and title, so I wanted to add more fields such as status, project, and tag...

hitesh96db:
@nursix Is it fine if it goes in static/formats/rss ?

nursix:
I think those can very well be in static/formats/rss!

What I would recommend, however, is to move any resource-specific XSLT in the export.xsl into a sub-stylesheet, and then import it into the main rss (I've done similar things in s3csv/hrm/person.xsl). 

This makes it more maintainable - and if we really wanted to move that into the template at some point, we don't need to rewrite everything.

hitesh96db:
Okay :)

hitesh96db:
@nursix 
How does importing prevent rewriting ?
say, I import "project.xsl" in "export.xsl", that href would be w.r.t that folder, so how would you reuse it or prevent re-writing everything if moved in the template ?

flavour:
Status, project & Tag all seem very generic things to be in the general export.trss

nursix:
Very easy to see: having resource-specific xsl (i.e. the  respective xsl:template's) in sub-stylesheets which are then imported into main export.xsl makes it easy to exchange just those sub-stylesheets rather than having to dig through a mega-monster-everything-in-one-file stylesheet with hundreds of embedded templates.

So, if we would want to replace the generic project/task RSS - then we could just move your project_task.xsl into the sunflower template - and write a new default one, rather than having to write a new all-in-one export.xsl.

nursix:
That is btw the main purpose of xsl:include and xsl:import - the ability to modularize stylesheets.

Since RSS is in no way limited to only a handful of resources, but can theoretically be applied to each and everything - and yet every resource has specific needs, we are likely to end up with one xsl:template per resource.

Hence, having only the root template ("s3xml") and the fallback in the master stylesheet, but the resource-specific parts in separate stylesheets makes it a lot easier to maintain.

I don't mind that to be per module, however - i.e. hrm.xsl, project.xsl, org.xsl ... to avoid over-fragmentation (and some parts may be re-usable between resources of the same module anyway). But avoiding a behemoth like person.xsl would be very much what I'd recommend.

nursix:
@flavour - yes, I would think so. They seem generic enough, nothing template-specific there.

flavour:
+1 to module-specific format extensions. Per resource isn't required, but a behemoth isn't desired either.

PR #1034, opened 2015-01-13
[S3Method]: Download assessment form(.xls and .rtf) working

raj454raj:
I have removed the code for "from PyRTF import something" and replaced by "from gluon.contrib.pyrtf import *" because now web2py has pyrtf module inbuilt. No need to check whether user have it installed or not!!! YAY !!!


modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+                try:
+                    baseObj.__dict__[element] = mergeObjectDiff(baseObj.__dict__[element],
+                                                                value,
+                                                                zeroObj.__dict__[element])
+                except:
+                    if zeroObj.__dict__[element] != value:
+                        baseObj.__dict__[element] = value
+            return baseObj
+
+        COL_WIDTH_MULTIPLIER = 240
```
raj454raj: Didnt know how to improve this!!! Any Idea?

modules/s3db/survey.py
```diff
@@ -83,17 +83,25 @@
...
 
 from gluon import *
+from gluon.contenttype import contenttype
+from gluon.contrib.pyrtf import *
+
 try:
     from gluon.dal.objects import Row
 except ImportError:
     # old web2py
     from gluon.dal import Row
```
nursix: Unused?

nursix: Nah - there is still one use-case. Maybe should be moved into that function.

raj454raj: :+1:

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+            filename = "%s.rtf" % series.name
+            contentType = ".rtf"
+
+        else:
+            r.error(405, current.ERROR.BAD_METHOD)
+
+        # If output is none => returned from series_export_spreadsheet
+        try:
+            output.seek(0)
+        except AttributeError:
```
nursix: This looks like it's obscuring the actual error - very bad style.

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        s3db = current.s3db
+        db = current.db
+        table = s3db.survey_series
+        vars = current.request.post_vars
+        series = r.record
+
+        if not series.logo:
+            logo = None
+        else:
+            if "Export_Spreadsheet" in vars:
```
nursix: We do not normally use case-sensitive URL vars, nor does a RESTful method handler actually need method vars. It would be better to bind this method to two separate methods (=2x set_method with the same action, then check r.method inside the handler)

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+                                "survey",
+                                "logo",
+                                "%s.%s" % (series.logo, ext)
+                                )
+            if not os.path.exists(logo) or not os.path.isfile(logo):
+                logo = None
+
+        # Get the translation dictionary
+        langDict = dict()
+        lang = vars.get("translationLanguage", None)
```
nursix: Same here - camel case URL var. Very ugly and fragile (URLs should not be case-sensitive)

modules/s3db/survey.py
raj454raj: Then redirect from that function to survey/series with session.error ?

modules/s3db/survey.py
nursix: First of all: naming convention! No camelCase please, but underscore.

nursix: Secondly - try to remove any unnecessary local variables, and move them closer to where they are used.

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        for section in sectionList:
+            sectionName = survey_T(section["name"], langDict)
+            rules = survey_getQstnLayoutRules(template_id,
+                                                section["section_id"])
+            layoutRules = [sectionName, rules]
+            layout.append(layoutRules)
+        widgetList = s3db.survey_getAllWidgetsForTemplate(template_id)
+        layoutBlocks = LayoutBlocks()
+
+        ######################################################################
```
nursix: Not our normal style for comments.

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+# -----------------------------------------------------------------------------
+class series_ExportFormatted(S3Method):
+    """
+        Download a Spreadsheet which can be filled-in offline & uploaded
+    """
+
+    # -------------------------------------------------------------------------------------
+    def series_prepare_matrix(self, series_id, series, logo, langDict, justified=False):
+        """
+            Helper function for apply_method
```
nursix: No parameter description in docstring

modules/s3survey.py
```diff
@@ -2538,7 +2553,12 @@ def writeToRTF(self, ss, langDict):
             This will just display the grid name, following this will be the
             grid child objects.
         """
-        from PyRTF import Paragraph, Cell, B, BorderPS, FramePS
+        from gluon.contrib.pyrtf.Elements import Paragraph, \
+                                                 Cell, \
+                                                 B
+        from gluon.contrib.pyrtf.PropertySets import BorderPS, \
```
nursix: Line limit is 80 characters, not 60

modules/s3db/survey.py
raj454raj: Two S3Method classes like this:
```
class series_ExportSpreadsheet(S3Method):
```
 and 
```
class series_ExportWord(S3Method):
```
:+1:

modules/s3db/survey.py
raj454raj: Yeah ;)

modules/s3survey.py
nursix: Do we actually need all those different writeToRTF's? Or can this be DRY'd? It seems to me that much of this could be parametrized, and hence a lot of code removed.

modules/s3db/survey.py
raj454raj: Changing names in this module would be my next PR?

nursix: Maybe consider switching to easyxf

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        borderR2 = xlwt.Borders()
+        borderR2.right = xlwt.Borders.MEDIUM
+
+        borderB1 = xlwt.Borders()
+        borderB1.bottom = xlwt.Borders.THIN
+        borderB2 = xlwt.Borders()
+        borderB2.bottom = xlwt.Borders.MEDIUM
+
+        alignBase = xlwt.Alignment()
+        alignBase.horz = xlwt.Alignment.HORZ_LEFT
```
nursix: Repeated lookup of xlwt.Alignment.HORZ_LEFT, consider local var "HORZ_LEFT"

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        alignWrap.vert = xlwt.Alignment.VERT_TOP
+        alignWrap.wrap = xlwt.Alignment.WRAP_AT_RIGHT
+
+        shadedFill = xlwt.Pattern()
+        shadedFill.pattern = xlwt.Pattern.SOLID_PATTERN
+        shadedFill.pattern_fore_colour = 0x16 # 25% Grey
+        shadedFill.pattern_back_colour = 0x08 # Black
+
+        headingFill = xlwt.Pattern()
+        headingFill.pattern = xlwt.Pattern.SOLID_PATTERN
```
nursix: Same - repeated lookup

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        book = xlwt.Workbook(encoding="utf-8")
+        output = StringIO()
+
+        protection = xlwt.Protection()
+        protection.cell_locked = 1
+        noProtection = xlwt.Protection()
+        noProtection.cell_locked = 0
+
+        borders = xlwt.Borders()
+        borders.left = xlwt.Borders.DOTTED
```
nursix: xlwt.Borders.DOTTED xlwt.Borders.DOTTED xlwt.Borders.DOTTED xlwt.Borders.DOTTED 

A thousand times...

modules/s3db/survey.py
raj454raj: :+1:

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+                the zeroObj and place them in the baseObj
+            """
+
+            elementList = newObj.__dict__
+            for (element, value) in elementList.items():
+                try:
+                    baseObj.__dict__[element] = mergeObjectDiff(baseObj.__dict__[element],
+                                                                value,
+                                                                zeroObj.__dict__[element])
+                except:
```
nursix: Except without exception class - no catch-all excepts please.

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        boxR1 = xlwt.XFStyle()
+        boxR1.borders = borderR1
+        boxR2 = xlwt.XFStyle()
+        boxR2.borders = borderR2
+        boxB1 = xlwt.XFStyle()
+        boxB1.borders = borderB1
+        boxB2 = xlwt.XFStyle()
+        boxB2.borders = borderB2
+        styleList = {}
+        styleList["styleTitle"] = styleTitle
```
nursix: Move these assignments closer to where the variable is defined.

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        styleInstructions.protection = protection
+        styleInstructions.alignment = alignWrap
+        styleBox = xlwt.XFStyle()
+        styleBox.borders = borders
+        styleBox.protection = noProtection
+        styleInput = xlwt.XFStyle()
+        styleInput.borders = borders
+        styleInput.protection = noProtection
+        styleInput.pattern = shadedFill
+        boxL1 = xlwt.XFStyle()
```
nursix: Uses only border styles => move closer to the border definitions

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+            return finalStyle
+
+        # -------------------------------------------------------------------------
+        def mergeObjectDiff(baseObj, newObj, zeroObj):
+            """
+                function to copy all the elements in newObj that are different from
+                the zeroObj and place them in the baseObj
+            """
+
+            elementList = newObj.__dict__
```
nursix: There is truly a nicer way to do this.

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        T = current.T
+        try:
+            import xlwt
+        except ImportError:
+            print >> sys.stderr, "ERROR: xlwt module is needed for importing spreadsheets"
+            return
+
+        import math
+        # -------------------------------------------------------------------------
+        def wrapText(sheet, cell, style):
```
nursix: I'd recommend to move out the inline-functions. Inline function definitions are useful if the exact definition depends on a condition - but it seems that these functions are constant, even static (no self-use!), so they should be moved out and made staticmethods.

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+            row = cell.row
+            col = cell.col
+            try:
+                text = unicode(cell.text)
```
nursix: s3_unicode please

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+            except UnicodeDecodeError:
+                text = cell.text
+            width = 16
+            # Wrap text and calculate the row width and height
+            characters_in_cell = float(width-2)
```
nursix: "Ouch" (technically) - "Hmm" :S (style-wise) - "OH!" (for the idea to actually parametrize it, even if not done eventually)

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+            if cell.merged():
+                try:
+                    sheet.write_merge(cell.row,
+                                      cell.row + cell.mergeV,
+                                      cell.col,
+                                      cell.col + cell.mergeH,
+                                      text,
+                                      style
+                                      )
+                except Exception as msg:
```
nursix: Catch-all except - should be avoided. What exactly can go wrong - and is there actually a valid state after that?

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+        styleList["boxL1"] = boxL1
+        styleList["boxL2"] = boxL2
+        styleList["boxT1"] = boxT1
+        styleList["boxT2"] = boxT2
+        styleList["boxR1"] = boxR1
+        styleList["boxR2"] = boxR2
+        styleList["boxB1"] = boxB1
+        styleList["boxB2"] = boxB2
+
+        sheet1 = book.add_sheet(T("Assessment"))
```
nursix: Here starts the actual function - which is a few miles away from the def. Very hard to read and debug :/

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+
+        series_id = r.id
+        if not series_id:
+            r.error(405, current.error.BAD_METHOD)
+
+        # Load Model
+        s3db = current.s3db
+        db = current.db
+        table = s3db.survey_series
+        vars = current.request.post_vars
```
nursix: Avoid "vars" - call it either "get_vars" or "post_vars" or "request_vars" ("vars" is a Python keyword)

nursix:
I'm sorry - there's at least a thousand things here that should be changed, I simply can't comment on everything at once. We wouldn't ever get to an end.

So, let's stage this a little bit...
Step 1: move out the inner functions, make staticmethods where possible
Step 2: move the style definitions into a function that returns the style list. Should probably be a lazy property, since the style-list doesn't ever change.
Step 3: Fix naming convention wherever easily possible (e.g. local vars) - make sure variable names are clear (Check: do you understand from the name what the variable constitutes?)
Step 4: Look for opportunities to DRY. IMHO at least half of the code looks like repititions

...and then let's review this again.

raj454raj:
@nursix : Okay thanks for review. Working on it!!! :+1:

raj454raj:
@nursix : We want this yeah?
https://github.com/flavour/eden/pull/1034#discussion_r22879038

modules/s3db/survey.py
nursix: Eeek - no!!!

If output.seek(0) raises an AttributeError, then something has gone wrong before. So why make it raise a completely different exception here instead of catching and handling the error where it happens?

This is provoking an exception in order to catch it! We should never get here unless output does have a seek().

raj454raj: It happens here https://github.com/flavour/eden/pull/1034/files#diff-c87b764fac9b35c325082f7c2cc850f4R3714
So do you suggest to redirect from here

nursix: Iiirrgh - and that doesn't even return anything :/

I would suggest to move this check out of the subfunction and into apply_method (at the top), and check and raise the error before even attempting to do anything else. Even better would be to also remove the related action buttons (+log a warning).

modules/s3db/survey.py
```diff
@@ -3563,4 +3572,523 @@ def apply_method(self, r, **attr):
...
+
+    # -----------------------------------------------------------------------------
+    def apply_method(self, r, **attr):
+        """
+            Download a Spreadsheet which can be filled-in offline & uploaded
+
+            @param r: the S3Request
+            @param attr: controller arguments
+        """
+
```
nursix: Here is where you should check for xlwt - because if it's not present, every subsequent action is futile.

If you move the action button definition into a function, you can even hide those if xlwt is not present.

raj454raj: :+1:

modules/s3db/survey.py
raj454raj: Got You!!!

modules/s3db/survey.py
nursix: No, can be the same method handler class - just bound to two different methods ;)

Here's the logic:

    set_method("my", "table", method="first", action=MyHandler)
    set_method("my", "table", method="second", action=MyHandler)

That way, it can be called either with /my/table/1/first or /my/table/1/second. When you get into apply_method, you can check which:

    def apply_method(self, r, **attr):
        if r.method == "first":
            # called with /first
        elif r.method == "second":
            # called with /second

Quite simple - no get_vars acrobatics.

nursix: It is btw best practice to move the actual method into a subfunction:
    def apply_method(self, r, **attr):
        if r.method == "first":
            return self.first(r, **attr)
        elif r.method == "second":
            return self.second(r, **attr)

    def first(self, r, **attr):
        """ Subfunction for /first """

    def second(self, r, **attr):
        """ Subfunction for /second """

...and then call commons from there.

In principle, all methods within the same set of functionality would typically be built into the same method handler class - you do not need to create a new method handler class every time (doesn't even make sense DRY-wise).

Maybe consider even to merge in the other two if that has advantages DRY-wise.

nursix: Some examples:

Primarily URL-controlled:
https://github.com/flavour/eden/blob/master/modules/s3/s3crud.py

Partially HTTP-controlled:
https://github.com/flavour/eden/blob/master/modules/s3/s3merge.py#L45

Pure REST:
https://github.com/flavour/eden/blob/master/modules/s3/s3hierarchy.py#L52

modules/s3db/survey.py
raj454raj: @nursix : Can you help in this.

raj454raj:
@nursix : Done. :+1:

modules/s3survey.py
```diff
@@ -1410,23 +1410,51 @@ def writeToMatrix(self,
         #    return (row+self.xlsMargin[1]+mergeLV+mergeWV, col+self.xlsMargin[0]+max(mergeLH,mergeWH))
 
     # -------------------------------------------------------------------------
-    def writeToRTF(self, ss, langDict):
+    def writeToRTF(self, ss, langDict, paragraph=None, para_list=[], question_name=""):
```
nursix: +1 to DRY this.

I would rename the base function as "_writeToRTF", and the specific caller functions into "writeQuestionToRTF" etc., rather than overwriting the base method and calling super() every time (yeah - unfortunately Python doesn't support overloading).

The base function _writeToRTF can then probably be staticmethod (so self-access as far as I can see?), which makes it a bit more efficient.

Else - good work.

modules/s3db/survey.py
```diff
@@ -3563,4 +3585,557 @@ def apply_method(self, r, **attr):
...
+            @param matrix: The matrix of cells
+            @param matrixAnswers: Matrix of Answers
+            @param logo: The logo of the file
+        """
+
+        # Already checked availabitlity of xlwt in the caller function
+        import xlwt
+        import math
+
+        COL_WIDTH_MULTIPLIER = 240
```
nursix: As I mentioned it before - I think all these configurations should probably be in a subfunction, so that the actual function starts after the docstring ;) otherwise one has to scroll and scroll and scroll until the actual function appears.

nursix:
Ok - I think if you clean up the pseudo-overloading of writeToRTF, then this first stage of cleanups is done. Looks a lot better already, doesn't it?
Before you proceed with the next round of improvements, consider maybe having a couple of unit tests to make sure your refactoring doesn't break anything. Strategically, this would be super-smart.

modules/s3db/survey.py
```diff
@@ -3563,4 +3585,557 @@ def apply_method(self, r, **attr):
...
+            sheet1.insert_bitmap(logo, 0, 0)
+
+        sheet1.protect = True
+        sheetA.protect = True
+        for i in range(26):
+            sheetA.col(i).width = 0
+
+        sheetA.write(0, 26,
+                     s3_unicode(T("Please do not remove this sheet")),
+                     styleHeader)
```
raj454raj: @nursix : Couldn't do this https://github.com/flavour/eden/pull/1034#discussion_r23083418 because styleHeader is used here so. There is a bad way to return styleHeader and styleList both from the function. But then this seems to be better than that! What say?

nursix: Ahm - styleHeader == style_list["styleHeader"] so where is the problem?

raj454raj: Okay my bad!!!

raj454raj:
>Looks a lot better already, doesn't it?

Really good!
Will try to write unit_tests, but believe me no previous experience in doing this.

raj454raj:
@nursix : S3QuestionTypeDateWidget, S3QuestionTypeStringWidget etc do not have writeToRTF defined so when this was called the base class function was called. If now we change the names, we need to define a function in the class something like this - 
```
    def writeQuestionToRTF(self, ss, langDict):
        return S3QuestionTypeAbstractWidget._writeToRTF(self, ss, langDict)
``` 
Is it fine?

modules/s3survey.py
```diff
@@ -1566,6 +1587,17 @@ def display(self, **attr):
             attr["_maxlength"] = length
         return S3QuestionTypeAbstractWidget.display(self, **attr)
 
+    # -------------------------------------------------------------------------
```
nursix: Consider the following (schematic):

    class BaseClass(object):
        @staticmethod
        def _complexMethod(param,kwarg1=None,kwarg2=None):
            pass # do something

        def wrapper(param):
            """ Wrapper for _complexMethod to simplify the call-pattern """
            return self._complexMethod(param)

    class SubClassA(BaseClass):
        # This class uses the standard wrapper (=no subclass-implementation of wrapper required)

    class SubClassB(BaseClass)
        def wrapper(param):
            """ 
                This class uses more complex call signature of _complexMethod, hence overwriting
                the default wrapper to pass in an instance attribute.
            """
            return self._complexMethod(param, kwarg2=self.attribute)

That way, the wrapper method needs to be implemented in the subclass only when it uses a more complex signature of _complexMethod - otherwise the simple default wrapper() of the base class will do.

Generally, of course, the wrapper would only be useful if called from outside the instance (otherwise it would unnecessarily complicate the flow).

nursix: ...or look at it like this: the wrapper is good to build an abstract interface, where the outside caller doesn't need to distinguish between the subclasses, but calls the instance method with always the same call signature (just "param"). The instance itself would then "know" which other parameters to add before calling the actual _complexMethod.

This is an acceptable reason for a wrapper - otherwise wrappers are problematic.

nursix: web2py's DIV class (+subclasses) make a good example for this structure ;)

raj454raj: Okay thanks :), sending changes in some time.

raj454raj: Oh! _writeToRTF uses self.fullName(but self is not an argument here). So _writeToRTF cant be staticmethod ? @nursix

raj454raj:
@nursix : done. Further suggestions? Unit tests or something else

raj454raj:
@nursix : What should be the target of the unit tests - both the file formats xls and rtf getting downloaded?
> next round of improvements

? Whats next, I'll start working on it.

nursix:
So, stage 1 and 2 are done - next would be 3.

But before working on 3, you should indeed have some unit tests. Unit tests test functions in isolation, i.e. you call a function with mockup-parameters and verify the results. Ideally, you check both success and failure response.

A schematic example. Say you have a function like this:

    def myfunc(a, b):
        return a / b

Then a unit test could be like this:

    def test_myfunc():
        result = myfunc(4, 2)
        if result != 2:
            raise AssertionError("Wrong result: should have been 2, but was %s." % result)

So you're writing another function to call the function that is to be tested, with parameters for which you know the result, in order to prove that the function really returns the expected result.

Typically, you should have at least one test call for each distinct type of case - e.g. for myfunc() this would additionally be the cases where either parameter is 0.

To make writing unit tests easier, Python provides a library "unittest" that supports more complex testing strategies and assertion checking. See modules/unit_tests - there are various examples of how unit tests could look like.

Important is that you don't ever use the tested function itself to determine which results you should expect ;) because then the unit test doesn't prove anything except for the fact that two identical calls deliver the same results.

Now, in order to write unit tests for this module, you need to look at each function and understand what results it should deliver for which input - and try to find a way to inspect and verify the result. Then identify the assertions (=expected results) for each type of case, and encode it as a test case.

E.g., get_style_list() should always return a dict() with certain members - that is an easy assertion, and quick to code. writeToRTF() is more complicated, but can still be introspected.

You do not need to test the overall functionality - all we need to know is whether the parts still do after the refactoring what they did before, the results are still correct, and no new bugs have been introduced (=regression testing).

raj454raj:
Okay thanks :+1: will work on it.

raj454raj:
Sorry, got some merging problem will send a new PR!

raj454raj:
@nursix : Doubts -
1. How do I call a S3Method class(series_ExportFormatted) without request object while writing tests. I mean how do I pass r and attr parameters to the apply_method function.
2. How do I create a temporary widget in tests by this -
```
        qtable = s3db.survey_question
        record = Storage(name = "Test for string widgets",
                         type = "String",
                         code = "72H-68")
        question_id = qtable.insert(**record)
```
but the problem here is every time I run the tests it inserts multiple times. Shall do a select first and then see if the record is there or not?
Thanks.

nursix:
You can just create a request:

    r = S3Request(...)

From there you can either call the method:

    method = survey_ExportFormatted()(r)

(note that you do not call apply_method directly, but execute the method)

Or, if you have the resource and method configured correctly in the request, just execute the request:

    output = r()

However, either way that is a bit too high level for a unit test.

nursix:
For the second question: 

First of all, unit tests should not permanently change the database status - if they need to create, update or delete records, they should rollback these changes at the end of the test.

Secondly, unit tests should ideally not depend on any particular database status - they should create all objects they need (and remove them again). The only exception is authorization for the test suite itself - this must be pre-defined (chicken-and-egg problem).

As long as you do not need to execute complex cascade operations (e.g. hierarchical deletion or something like that), a simple db.rollback at the end of each test case (=tearDown) should do the trick.

raj454raj:
Okay got you thanks a lot. :+1:

PR #1035, opened 2015-01-13
Small Fix

samsruti:
1. http://127.0.0.1:8000/eden/project/task
When we click on the data table header like "Created On" , the elements of that column got shifted towards right cloumn.
 2. Fix for http://sunflower.sahanafoundation.org/eden/project/task/373

### DIFF
--- a/modules/s3db/hrm.py
+++ b/modules/s3db/hrm.py
@@ -165,6 +165,8 @@ class S3HRModel(S3Model):
                                  ),
                      *s3_meta_fields())
 
+        label = T("Create a New Department")
+        tooltip = T("If you don't see the Department in the list, you can add a new one by clicking link 'Create Department' ")
         label_create = T("Create Department")
         crud_strings[tablename] = Storage(
             label_create = label_create,
@@ -193,7 +195,9 @@ class S3HRModel(S3Model):
             sortby = "name",
             comment = S3AddResourceLink(c="vol" if group == "volunteer" else "hrm",
                                         f="department",
-                                        label=label_create),
+                                        label=label_create,
+                                        title=label,
+                                        tooltip=tooltip),
             )
 
         configure("hrm_department",
@@ -450,17 +454,10 @@ class S3HRModel(S3Model):
                                           readable = False,
                                           writable = False,
                                           ),
-                     Field("org_contact", "boolean",
-                           label = T("Organization Contact"),
-                           represent = s3_yes_no_represent,
-                           readable = False,
-                           writable = False,
-                           ),
                      Field("site_contact", "boolean",
                            label = T("Facility Contact"),
-                           represent = s3_yes_no_represent,
-                           #represent = lambda opt: \
-                            #(T("No"), T("Yes"))[opt == True],
+                           represent = lambda opt: \
+                            (T("No"), T("Yes"))[opt == True],
                            ),
                      s3_comments(),
                      *s3_meta_fields())
@@ -2401,9 +2398,7 @@ class S3HRSkillModel(S3Model):
                              label = T("Date Received")
                              ),
                      s3_date("end_date",
-                             # @ToDo: Change implicit default to explicit default with interval of 12 months
-                             start_field = "hrm_credential_start_date",
-                             default_interval = 12,
+                             # @ToDo: Automation based on deployment_settings, e.g.: date received + 6/12 months
                              label = T("Expiry Date")
                              ),
                      *s3_meta_fields())
@@ -3087,16 +3082,15 @@ class S3HRSkillModel(S3Model):
                           _title="%s|%s" % (T("Competency Rating"),
                                             T("Level of competency this person has with this skill.")))
         if current.deployment_settings.get_hrm_skill_types():
-            script = \
+            s3.js_global.append('''i18n.no_ratings="%s"''' % T("No Ratings for Skill Type"))
+            s3.jquery_ready.append(
 '''$.filterOptionsS3({
  'trigger':'skill_id',
  'target':'competency_id',
  'lookupResource':'competency',
  'lookupURL':S3.Ap.concat('/%s/skill_competencies/'),
- 'msgNoRecords':'%s'
-})''' % (controller, T("No Ratings for Skill Type"))
-            comment = TAG[""](comment,
-                              S3ScriptItem(script=script))
+ 'msgNoRecords':i18n.no_ratings
+})''' % controller)
 
         return comment
 
@@ -5591,7 +5585,8 @@ def hrm_training_organisation(row):
     return current.messages["NONE"]
 
 # =============================================================================
-def hrm_rheader(r, tabs=[], profile=False):
+def hrm_rheader(r, tabs=[],
+                profile=False):
     """ Resource headers for component views """
 
     if r.representation != "html":
@@ -5782,11 +5777,6 @@ def hrm_rheader(r, tabs=[], profile=False):
         else:
             id_tab = None
 
-        if settings.get_hrm_use_address():
-            address_tab = (T("Address"), "address")
-        else:
-            address_tab = None
-
         if settings.get_hrm_salary():
             salary_tab = (T("Salary"), "salary")
         else:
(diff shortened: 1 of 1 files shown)


modules/templates/default/config.py
```diff
@@ -1069,12 +1069,12 @@ def config(settings):
         #   module_type = 10,
         #   #access = "|DVI|",      # Only users with the DVI role can see this module in the default menu & access the controller
         #)),
-        #("mpr", Storage(
-        #   name_nice = T("Missing Person Registry"),
-        #   #description = "Helps to report and search for missing persons",
-        #   restricted = True,
-        #   module_type = 10,
-        #)),
+        ("mpr", Storage(
```
somayjain: What is this for?

modules/templates/SSF/views/index.html
```diff
@@ -18,7 +18,7 @@
         </div>
 
         <!-- Login/Register, Report a Problem buttons -->
-        <div class='row'>
+        <div class='row-home'>
```
flavour: Do we really want this to NOT be a 'row'? (That has clear meaning in Foundation at least)
Perhaps just add a 2nd class to distinguish in CSS - e.g. .row.home

modules/templates/SSF/config.py
```diff
@@ -743,7 +743,7 @@ def customise_project_task_resource(r, tablename):
             msg_record_created = None,
             msg_record_deleted = None
         )
-        if r.interactive:
+        if r.interactive or r.representation == "aadata":
```
flavour: This is a fix fior something else?

flavour: Looks like it's to ensure the list_fields in aadata match interactive.

modules/templates/SSF/views/index.html
samsruti: After this change
.row {
    margin: 0px;
} 
when i opened the index page then the alignments of homepage elements gets distorted . So i added the row-home(according to the foundation css)

nursix: SSF template is not using foundation, and row-home is not a foundation class.

nursix: But row is a bootstrap class ;)

static/themes/SSF/style.css
```diff
@@ -42,3 +42,11 @@
     background-image: url(../../img/sunflower_spin_throbber.gif);
     background-repeat: no-repeat;
 }
+
+.row {
```
nursix: I wouldn't do that (at least not globally) - better follow bootstrap.

modules/templates/SSF/views/index.html
samsruti: i mean to make the alignment of datatable , forms i changed 
.row {
margin: 0px;
} 
since this made the wrong position of  homepage elements (buttons ,etc) so i changed the row class to row-home as that problem is only occuring in the index page so i changed only in the index file

static/themes/SSF/style.css
samsruti: but without doing this the alignment of forms and data tables are not in correct order as they should be aligned with the Sunflower logo.
So i used this globally.

hitesh96db: The solution I was thinking of was to define SSF specific html files...Since no html files are defined right now, it takes the foundation ones, and hence, the classes are applied in a wrong way.

modules/templates/SSF/views/index.html
nursix: I understand - but it's still not the right fix.

The problem is rather that this is using the default _list_filter.html, which is a foundation grid layout - and bootstrap doesn't understand the "columns" class (bootstrap is using span-x). 

Hence, the right fix would be to implement a suitable _list_filter.html for bootstrap using span-x classes instead of columns.

Overriding bootstrap in order to deal with a foundation layout brings up the question: why use bootstrap then?

samsruti: Thanx.. :+1: 
When i changed _list_filter.html for bootstrap using span-x classes , only the data table get aligned.
So i have to do that for others also ?

nursix: Not "change" _list_filter.html - you need a custom one in the SSF template.

This one would work:
```
<div class='row'>
 <div class='span-12'>
  {{try:}}{{=H2(title)}}{{except:}}{{pass}}
  {{try:}}{{=s3db.ocr_buttons(r)}}{{except:}}{{pass}}
 </div>
</div>
<div class='row'>
 <div id='rheader' class='span-12'>
  {{try:}}{{=XML(rheader)}}
 </div>
</div>
<div id='component' class='row'>
{{except:}}
 </div>
</div>
<div class='row'>
{{pass}}
 <div class='span-12'>
  <div class='row'>
   <div id='map' class='span-12'></div>
  </div>
  <div class='row'>
   <div class='span-12'>
    <div id='list-btn-add'>
     {{try:}}{{showadd_btn=showadd_btn}}{{except:}}{{showadd_btn=None}}{{pass}}
     {{if showadd_btn:}}{{=showadd_btn}}{{hide_listadd = True}}
     {{else:}}{{try:}}{{=buttons["add_btn"]}}{{except:}}{{pass}}{{hide_listadd = False}}{{pass}}
    </div>
    <div id='list-add'>
     {{try:}}{{=H3(addtitle)}}{{except:}}{{pass}}
     {{try:}}{{=addheader}}{{except:}}{{pass}}
     {{include "key.html"}}
     <div class='form-container'>
      {{try:}}{{=form}}{{except:}}{{pass}}
     </div>
    </div>
   </div>
  </div>
  <div class='row'>
   <div class='span-12'>
    {{try:}}{{if list_filter_form:}}
    <div id='list-filter' class='form-container'>
     {{=list_filter_form}}
    </div>
    {{pass}}
    {{except:}}{{pass}}
   </div>
  </div>
  <div class='row'>
   {{try:}}{{ltype = list_type}}{{except:}}{{ltype = "datatable"}}{{pass}}
   {{try:}}{{items = items}}
   {{if ltype == "datalist":}}
    <div id='dl-container' class='span-12'>
     {{=items}}
    </div>
   {{else:}}
    <div id='table-container' class='span-12'>
     {{=items}}
    </div>
   {{pass}}
   {{except:}}{{pass}}
  </div>
 </div>
</div>
<div class='row'>
 {{if s3.rfooter:}}
 <div id='rfooter' class='span-12'>
  {{=XML(s3.rfooter)}}
 </div>
 {{pass}}
</div>
{{if ltype == "datalist":}}
{{include "dataLists.html"}}
{{else:}}
{{include "dataTables.html"}}
{{pass}}
{{if r.http == "POST" or not hide_listadd:}}
{{script = '''$('#list-add').show()
$('#show-add-btn').hide()'''}}
{{s3.jquery_ready.append(script)}}
{{pass}}
```

nursix: But hey...this is a very crude translation of the default _list_filter.html into bootstrap. Rather than that, you may actually want to make one that fits neatly into the Sunflower design ;)

samsruti: Wait I have done something ;)

static/themes/SSF/style.css
nursix: Right what I'm saying @hitesh96db - overriding bootstrap.css in order to make it work with non-bootstrap layouts somehow defeats the original idea of using bootstrap.

But of course, alternatively you can throw out bootstrap and use foundation instead ;)

samsruti:
I saw that but when i opened the admin page of SSF template that page also work fine .

flavour:
+1 to making a decision to move SSF template from Bootstrap -> Foundation, although that's bigger than this PR as a decision ;)

hitesh96db:
@nursix I had a doubt, so asked you here: https://github.com/flavour/eden/pull/1033
Sort of unrelated to this or that PR ;)

nursix:
Yeah - "working fine" is relative. I think you'll take that back on mobile ;)

The point with grid classes (regardless which of the frameworks) is to be responsive, i.e. have it adapt to screen size.

PR #1036, opened 2015-01-14
Module-specifc.xsl - RSS

hitesh96db:


nursix:
Excellent! Many thanks. Tested with Sunflower I assume?

nursix:
+1 to merge

nursix:
Now one could say that base.xsl could be moved back into the master stylesheet ;) but hey...

PR #1037, opened 2015-01-15
Explicit Default for s3.end_date_interval

gauravmittal1995:
1) Created an option for explicit default in s3.end_date_interval
2) Using S3.start_end_date in hrm_credential


gauravmittal1995:
@nursix, @flavour  Please Review

PR #1039, opened 2015-01-15
SSF: Comments RSS, Subscribe task author to updates

hitesh96db:
- Comments rss - template-specific ?
- Task author should automatically be subscribed to task updates

### DIFF
--- a/modules/templates/SSF/config.py
+++ b/modules/templates/SSF/config.py
@@ -648,9 +648,7 @@ def config(settings):
 
         query = (mtable.task_id == task_id) & \
                 (mtable.deleted == False) & \
-                (mtable.person_id == user) & \
-                (rtable.deleted == False) & \
-                (rtable.id == mtable.role_id)
+                (mtable.person_id == user)
 
         user_role = db(query).select(mtable.id,
                                      mtable.uuid).first()
@@ -660,17 +658,6 @@ def config(settings):
 
         if user_role:
             data["id"] = str(user_role.id)
-
-        # Need UUID to insert via 's3json'
-        # UUID of 'Watching' Role
-        query = (rtable.role == "Watching") & \
-                (rtable.deleted == False)
-        role = db(query).select(rtable.uuid).first()
-        try:
-            data["role_uuid"] = role.uuid
-        except:
-            data["role_uuid"] = "null"
-            current.log.error("Task Roles not prepopulated")
 
         return data
 
@@ -724,6 +711,20 @@ def config(settings):
                 interest = TD(button)
 
         return interest
+
+    # -----------------------------------------------------------------------------
+    def custom_project_task_create_onaccept(form):
+        """
+            - Subscribe author to updates
+            - Add author as a Member
+        """
+
+        user = current.auth.s3_logged_in_person()
+        task_id = form.vars.id
+        current.s3db.project_member.insert(task_id=task_id,
+                                           person_id=user)
+        sub = TaskSubscriptions()
+        sub.add_task_subscription(str(task_id))
 
     # -----------------------------------------------------------------------------
     def customise_project_task_resource(r, tablename):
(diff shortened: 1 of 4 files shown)


modules/templates/SSF/config.py
```diff
@@ -858,6 +886,8 @@ def customise_project_task_resource(r, tablename):
             s3db.configure(tablename,
                            crud_form = crud_form,
                            filter_widgets = custom_filter_widgets,
+                           create_onaccept = lambda form: \
+                                custom_project_task_create_onaccept(form, create_onaccept)
```
nursix: That is not how we normally do it. All callbacks in configure can be lists, e.g.

    create_onaccept = [custom_project_task_onaccept, create_onaccept]

...and will then be executed in list order.

modules/templates/SSF/config.py
```diff
@@ -726,6 +726,33 @@ def task_member_option(data,
...
 
     # -----------------------------------------------------------------------------
+    def custom_project_task_create_onaccept(form, create_onaccept):
+        """
+            - Subscribe author to updates
+            - Add author as a Member
+        """
+
+        # Standard onaccept
+        create_onaccept(form)
```
nursix: Not that way, no - make the onaccepts a list in configure.

modules/templates/SSF/config.py
```diff
@@ -940,6 +970,28 @@ def link(self, k, v, row=None):
...
+
+        # Custom Prep
+        standard_prep = s3.prep
+        def custom_prep(r):
+            # Call standard prep
+            if callable(standard_prep):
+                result = standard_prep(r)
+                if not result:
+                    return False
+            if r.get_vars.format == "msg" or r.representation == "rss":
```
nursix: Why test this after the standard prep?

Further - what is the difference between r.get_vars.format == "msg" and r.representation == "msg"? Answer: r.get_vars.format=="msg" will only detect ?format=msg - whereas r.representation=="msg" will detect both .msg extension as well as ?format=msg.

So, this should be:

    if r.representation in ("msg", "rss"):

modules/templates/SSF/config.py
```diff
@@ -940,6 +970,28 @@ def link(self, k, v, row=None):
...
+                return True
+            return False
```
nursix: However, if you intend to return False for all other formats anyway, then you can equally well skip the standard prep.

modules/templates/SSF/config.py
```diff
@@ -726,6 +726,33 @@ def task_member_option(data,
         return interest
 
     # -----------------------------------------------------------------------------
+    def custom_project_task_create_onaccept(form, create_onaccept):
+        """
+            - Subscribe author to updates
+            - Add author as a Member
```
nursix: Stupid questions: Say - Fran creates a new task (he gets subscribed and is made a member). Then, I update the task description to better reflect the problem.
=> Does Fran remain subscribed as the task author?
=> Do I get subscribed as a new member (or as author?)?
=> Is that correct?

nursix: Or other funny task actions: say, I update the task status from "closed" to "reopened". Will I become the new "task author"? Will I get subscribed to the task? Will I become a member? Not? Why not? Will I? Why?

Yeah...I know, now I'm becoming a PITA.

hitesh96db: @nursix 
Fran creates a new task  - you update it..
Does Fran remain subscribed as the task author?
--- Yes, but note, there's no author role yet..
Do you get subscribed as a new member (or as author?)? 
--- No, but if you're updating it, that would mean you're interested in the task(?), So.....Click the 'Watch' button :D ?
The idea here was to keep the author subscribed to the task, so that,
- if a user comments, there's someone to answer/reply to that
- if the task is edited, the author knows about it ;)

modules/templates/SSF/config.py
hitesh96db: :+1:

modules/templates/SSF/config.py
nursix: Ok, understood.

hitesh96db:
done

hitesh96db:
@flavour done

hitesh96db:
@flavour done

PR #1040, opened 2015-01-16
Minor fixes in date and form submission + cleanup

raj454raj:
s3_menu_postp is not being called anywhere!

### DIFF
--- a/controllers/dvi.py
+++ b/controllers/dvi.py
@@ -9,42 +9,6 @@ if not settings.has_module(module):
     raise HTTP(404, body="Module disabled: %s" % module)
 
 # -----------------------------------------------------------------------------
-def s3_menu_postp():
-    # @todo: rewrite this for new framework
-    menu_selected = []
-    body_id = s3base.s3_get_last_record_id("dvi_body")
-    if body_id:
-        body = s3db.dvi_body
-        query = (body.id == body_id)
-        record = db(query).select(body.id, body.pe_label,
-                                  limitby=(0,1)).first()
-        if record:
-            label = record.pe_label
-            response.menu_options[-3][-1].append(
-                [T("Candidate Matches for Body %(label)s") % dict(label=label),
-                 False, URL(f="person",
-                            vars=dict(match=record.id))]
-            )
-            menu_selected.append(
-                ["%s: %s" % (T("Body"), label),
-                 False, URL(f="body", args=[record.id])]
-            )
-    person_id = s3base.s3_get_last_record_id("pr_person")
-    if person_id:
-        person = s3db.pr_person
-        query = (person.id == person_id)
-        record = db(query).select(person.id, limitby=(0, 1)).first()
-        if record:
-            name = s3db.pr_person_id().represent(record.id)
-            menu_selected.append(
-                ["%s: %s" % (T("Person"), name),
-                 False, URL(f="person", args=[record.id])]
-            )
-    if menu_selected:
-        menu_selected = [T("Open recent"), True, None, menu_selected]
-        response.menu_options.append(menu_selected)
-
-# -----------------------------------------------------------------------------
 def index():
     """ Module's Home Page """
 
--- a/controllers/hms.py
+++ b/controllers/hms.py
@@ -11,33 +11,6 @@ if not settings.has_module(module):
     raise HTTP(404, body="Module disabled: %s" % module)
 
 # -----------------------------------------------------------------------------
-def s3_menu_postp():
-    # @todo: rewrite this for new framework
-    if len(request.args) > 0 and request.args[0].isdigit():
-        newreq = dict(from_record="hms_hospital.%s" % request.args[0],
-                      from_fields="hospital_id$id")
-        #selreq = {"req.hospital_id":request.args[0]}
-    else:
-        newreq = dict()
-    selreq = {"req.hospital_id__ne":"NONE"}
-    menu_selected = []
-    hospital_id = s3base.s3_get_last_record_id("hms_hospital")
-    if hospital_id:
-        hospital = s3db.hms_hospital
-        query = (hospital.id == hospital_id)
-        record = db(query).select(hospital.id,
-                                  hospital.name,
-                                  limitby=(0, 1)).first()
-        if record:
-            name = record.name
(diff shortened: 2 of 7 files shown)


controllers/dvi.py
```diff
@@ -9,42 +9,6 @@
     raise HTTP(404, body="Module disabled: %s" % module)
 
 # -----------------------------------------------------------------------------
-def s3_menu_postp():
-    # @todo: rewrite this for new framework
```
nursix: Missed the "todo" here? Not "remove this" but "rewrite this"

raj454raj: But the function isnt called anywhere.

raj454raj: Primarily I was trying for the todo but didnt find where was this used.

modules/s3db/inv.py
```diff
@@ -4071,13 +4071,11 @@ def model(self):
                            #widget = S3InvBinWidget("inv_adj_item")
                            ),
                      # Organisation that owned this item before
-                     organisation_id(name = "old_owner_org_id",
```
nursix: Can't just drop the field name here.

modules/s3db/inv.py
```diff
@@ -4071,13 +4071,11 @@ def model(self):
...
-                                     label = T("Current Owned By (Organization/Branch)"),
+                     organisation_id(label = T("Current Owned By (Organization/Branch)"),
                                      ondelete = "SET NULL",
-                                     writable = False),
```
nursix: That's intentional.

controllers/pr.py
```diff
@@ -8,35 +8,6 @@
 resourcename = request.function
 
 # -----------------------------------------------------------------------------
-# Options Menu (available in all Functions' Views)
```
nursix: This case can indeed be dropped.

controllers/hms.py
```diff
@@ -11,33 +11,6 @@
     raise HTTP(404, body="Module disabled: %s" % module)
 
 # -----------------------------------------------------------------------------
-def s3_menu_postp():
```
nursix: +1, I think this can be removed.

controllers/dvi.py
nursix: I think we want to retain this functionality - but probably as action button.

modules/s3db/inv.py
raj454raj: Why show the Create Organization option then after submitting the popup it goes to this url
http://127.0.0.1:8000/eden/inv/adj_item_new_owner_org_id/options.s3json?field=organisation_id
which indeed is not defined

controllers/dvi.py
raj454raj: What is the functionality?

controllers/hms.py
raj454raj: :+1:

controllers/vol.py
```diff
@@ -549,6 +549,12 @@ def postp(r, output):
                 # Set the minimum end_date to the same as the start_date
                 s3.jquery_ready.append(
 '''S3.start_end_date('hrm_experience_start_date','hrm_experience_end_date')''')
+
+            elif r.component_name == "identity":
+                # Set the minimum valid_until to the same as the valid_from
+                s3.jquery_ready.append(
+'''S3.start_end_date('pr_identity_valid_from','pr_identity_valid_until')''')
```
nursix: I wonder whether that needs a separate controller-action each and every time? It would be a lot better to place this into the s3_date constructor, and then apply it in the respective model - otherwise we need to repeat this a thousand times.

raj454raj: +1 To the idea will work on that. Will this be merged or I shall close it?

modules/s3db/inv.py
raj454raj: @nursix : ^

nursix: It's writable=False, and should therefore not have any AddResourceLink (set comment=None). But it should still be writable=False.
The URL is wrong due to an invalid caller-setting, which again is caused by this being a read-only field. So, simply remove the AddResourceLink here, but leave the writable=False in place.

raj454raj: Awesome! thanks :)

controllers/vol.py
nursix: I would like to stop this here before merging even more of those injections. Better to have a DRY solution in s3_date and apply in the respective model.
Here for example you target pr_identity - but pr_identity is used in multiple places. Shall we really inject this from multiple controllers rather than only once in the model?

controllers/dvi.py
nursix: Providing an action item to see the candidate matches for the current body.

modules/s3db/inv.py
nursix: However, you *must* retain the field names - these are two different fields, hence need two different field names.

controllers/dvi.py
nursix: It currently not called because the menu framework has changed.
However, it should not be in the menu anymore - but rather near the record (either as action button, or e.g. as a tab in the record view, or both).

controllers/vol.py
raj454raj: @nursix : Yeah got the idea!!! So, only the form submission fixes and removing s3_menu_postp from hms and dvi not from pr.py should remain in the PR

raj454raj: @nursix ^

flavour: +1 to Model (I thought exactly the same when merging the last one), although this should be as an S3ScriptItem so only gets added if rendered, not just by running the model...

raj454raj: Okay I'll close this PR and merge the minor form fix in my other PR, @nursix s3_menu_postp should be as is yeah? Will consider that @flavour :)

nursix: I'm not saying it "should be as-is" - I'm just saying that we would want to retain the functionality and re-implement it using current functionality and design rather than just removing it. How exactly we do that - that's a separate question. I don't think it should still be in the menu.

raj454raj: @nursix :Yeah ;) I meant specific to this PR. Can you review my other PR if possible?

PR #1042, opened 2015-01-19
Deployment Settings for interval in S3.end_date_interval

gauravmittal1995:
Please let me know which ones need explicit default.
Any other suggestions?

### DIFF
--- a/modules/s3db/survey.py
+++ b/modules/s3db/survey.py
@@ -1633,6 +1633,8 @@ class S3SurveySeriesModel(S3Model):
                                   ),
                           s3_date("end_date",
                                   label = T("End Date"),
+                                  start_field = "survey_series_start_date",
+                                  default_interval = 1,
                                   ),
                           #self.super_link("source_id", "doc_source_entity"),
                           *s3_meta_fields())
--- a/modules/s3db/asset.py
+++ b/modules/s3db/asset.py
@@ -938,6 +938,8 @@ class S3AssetTelephoneModel(S3Model):
                           # @ToDo: Validation to ensure not before Start Date
                           s3_date("end_date",
                                   label = T("End Date"),
+                                  start_field = "asset_telephone_usage_date",
+                                  default_interval = 1,
                                   ),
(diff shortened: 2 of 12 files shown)


gauravmittal1995:
@flavour Please review

flavour:
Travis CI is failing for some reason.
In general I'm very wary about this - seems like we're adding a lot of complexity for little value?
We already have too many deployment_settings really...I suspect these will be very unlikely to be set & so are just more settings that people have to wade through & try to understand when they're trying to configure their system.
Just because we can, doesn't necessarily mean that we should?

gauravmittal1995:
@flavour Removed the deployment Settings. Should all these have implicit default? or some of them should have explicit defaults??

gauravmittal1995:
@flavour Also is there any important instance in one of these where a deployment setting may come useful?

modules/s3db/gis.py
```diff
@@ -242,6 +242,8 @@ def model(self):
                     ),
             s3_date("end_date",
                     label = T("End Date"),
+                    start_field = "gis_location_start_date",
+                    default_interval = 12,
```
flavour: I don't think this has a default_interval at all...Locations change based on local politics

modules/s3db/survey.py
```diff
@@ -1633,6 +1633,8 @@ def model(self):
                                   ),
                           s3_date("end_date",
                                   label = T("End Date"),
+                                  start_field = "survey_series_start_date",
+                                  default_interval = 12,
```
flavour: A survey_series is normally a very brief time...default_interval might be 1 month

gauravmittal1995:
@flavour Done the changes, Also added some more instances of start and end date.
Please review

flavour:
Am still wary about looking for too many of these - rather keep it as a useful tool we can easily configure now.
If the issues highlighted are fixed then it can be merged, thanks :)

gauravmittal1995:
@flavour  Done.

PR #1043, opened 2015-01-19
[DateWidget]: Disabling previous dates before selected start date

raj454raj:
@flavour : Done.

### DIFF
--- a/modules/s3db/project.py
+++ b/modules/s3db/project.py
@@ -185,10 +185,11 @@ class S3ProjectModel(S3Model):
                      self.project_status_id(),
                      # NB There is additional client-side validation for start/end date in the Controller
                      s3_date("start_date",
-                             label = T("Start Date")
+                             label = T("Start Date"),
                              ),
                      s3_date("end_date",
-                             label = T("End Date")
+                             label = T("End Date"),
+                             start_field = "project_project_start_date",
                              ),
                      # Free-text field with no validation (used by OCHA template currently)
                      Field("duration",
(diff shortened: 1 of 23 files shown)


modules/s3db/survey.py
```diff
@@ -1633,6 +1633,7 @@ def model(self):
                                   ),
                           s3_date("end_date",
                                   label = T("End Date"),
+                                  start_field = "survey_series_start_date",
                                   ),
```
gauravmittal1995: @raj454raj , I have already done this in my PR (https://github.com/flavour/eden/pull/1042), please remove this as it will cause merge conflicts with mine. And also some other options are needed to be added.

modules/s3db/project.py
```diff
@@ -1706,6 +1708,7 @@ def model(self):
                      s3_date("end_date",
                              #empty = False,
                              label = T("End Date"),
+                             start_field = "project_beneficiary_date",
                              ),
```
gauravmittal1995: @raj454raj , I have added project_project, project_activity, project_beneficiary in https://github.com/flavour/eden/pull/1042 . Please refer to this PR and remove these, as it can cause merge conflicts. Thanks :)

modules/s3db/survey.py
raj454raj: Okay I'll remove that after review, the idea was to add start_field to ALL the places where the pair occurs so... (Though I dont think merge conflict for 2 lines matter much ;) )

flavour:
Travis CI is failing.
Other than that, this doesn't seem to be using the new widget options, so I guess should be rebased to see / use that?

PR #1044, opened 2015-01-20
[DateTimeWidget]: Disabling previous dates

raj454raj:
Both for S3DateWidget and S3DateTimeWidget

### DIFF
--- a/modules/s3/s3fields.py
+++ b/modules/s3/s3fields.py
@@ -1307,27 +1307,27 @@ def s3_date(name="date", **attr):
 
     if "widget" not in attr:
         # Widget Options
-        widget_option = {}
+        widget_opts = {}
 
         if "start_field" in attr:
-            widget_option["start_field"] = attr["start_field"]
+            widget_opts["start_field"] = attr["start_field"]
             del attr["start_field"]
 
         if "default_interval" in attr:
-            widget_option["default_interval"] = attr["default_interval"]
+            widget_opts["default_interval"] = attr["default_interval"]
             del attr["default_interval"]
 
         if "default_explicit" in attr:
-            widget_option["default_explicit"] = attr["default_explicit"]
+            widget_opts["default_explicit"] = attr["default_explicit"]
             del attr["default_explicit"]
 
         if future:
-            widget_option["future"] = future
+            widget_opts["future"] = future
 
         if past:
-            widget_option["past"] = past
+            widget_opts["past"] = past
 
-        attr["widget"] = S3DateWidget(**widget_option)
+        attr["widget"] = S3DateWidget(**widget_opts)
 
     f = S3ReusableField(name, "date", **attr)
     return f()
@@ -1343,6 +1343,7 @@ def s3_datetime(name="date", **attr):
             widget = "date" (in addition to usual meanings)
             past = x hours
             future = x hours
+            start_field = "selector" for start field
 
         @ToDo: Different default field name in case we need to start supporting
                Oracle, where 'date' is a reserved word
@@ -1370,18 +1371,26 @@ def s3_datetime(name="date", **attr):
         attr["represent"] = lambda dt: S3DateTime.date_represent(dt,
                                                                  utc=True)
 
+    # Widget Options
+    widget_opts = {}
+
+    if "start_field" in attr:
+        widget_opts["start_field"] = attr["start_field"]
+        del attr["start_field"]
+    else:
+        start_field = None
+
     if "widget" not in attr:
-        if past is None and future is None:
-            attr["widget"] = S3DateTimeWidget()
-        elif past is None:
-            attr["widget"] = S3DateTimeWidget(future=future)
-        elif future is None:
-            attr["widget"] = S3DateTimeWidget(past=past)
-        else:
-            attr["widget"] = S3DateTimeWidget(past=past, future=future)
+
+        if future:
+            widget_opts["future"] = future
+
+        if past:
+            widget_opts["past"] = past
+
     elif attr["widget"] == "date":
         if past is None and future is None:
-            attr["widget"] = S3DateWidget()
+
             requires = IS_DATE(
                     format=current.deployment_settings.get_L10n_date_format()
                 )
@@ -1390,7 +1399,8 @@ def s3_datetime(name="date", **attr):
             current_month = now.month
             if past is None:
                 future = int(round(future/744.0, 0))
-                attr["widget"] = S3DateWidget(future=future)
+                widget_opts["future"] = future
+
                 future_month = now.month + future
                 if future_month <= 12:
                     max = now.replace(month=future_month)
@@ -1411,7 +1421,8 @@ def s3_datetime(name="date", **attr):
                     )
             elif future is None:
                 past = int(round(past/744.0, 0))
-                attr["widget"] = S3DateWidget(past=past)
+                widget_opts["past"] = past
+
                 if past < current_month:
                     min = now.replace(month=current_month - past)
                 else:
@@ -1432,7 +1443,9 @@ def s3_datetime(name="date", **attr):
             else:
                 future = int(round(future/744.0, 0))
                 past = int(round(past/744.0, 0))
-                attr["widget"] = S3DateWidget(past=past, future=future)
+                widget_opts["future"] = future
+                widget_opts["past"] = past
+
                 future_month = now.month + future
                 if future_month <= 12:
                     max = now.replace(month=future_month)
@@ -1513,6 +1526,8 @@ def s3_datetime(name="date", **attr):
         else:
             # Default
             attr["requires"] = IS_EMPTY_OR(requires)
+
+    attr["widget"] = S3DateTimeWidget(**widget_opts)
 
     f = S3ReusableField(name, "datetime", **attr)
     return f()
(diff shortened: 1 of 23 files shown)


raj454raj:
@flavour : done :+1:

modules/s3/s3fields.py
```diff
@@ -1514,6 +1527,8 @@ def s3_datetime(name="date", **attr):
             # Default
             attr["requires"] = IS_EMPTY_OR(requires)
 
+    attr["widget"] = S3DateTimeWidget(**widget_opts)
```
flavour: Is this over-riding explicit attr["widget"]?

modules/s3/s3fields.py
```diff
@@ -1370,18 +1371,26 @@ def s3_datetime(name="date", **attr):
...
                                                                  utc=True)
 
+    # Widget Options
+    widget_opts = {}
+
+    if "start_field" in attr:
+        widget_opts["start_field"] = attr["start_field"]
+        del attr["start_field"]
+    else:
+        start_field = None
```
flavour: Can we not remove the else = None here?

modules/s3/s3widgets.py
```diff
@@ -1353,10 +1353,20 @@ def __call__(self, field, value, **attributes):
...
+            start_end_date_script = \
+'''S3.start_end_date('%(start_field)s', '%(selector)s')''' % \
+                dict(start_field = start_field, # Field selector for start_date field
+                     selector = selector)       # Field selector for end_date_field
+
+            if start_end_date_script not in jquery_ready:                     # Prevents loading twice when form has errors
+                jquery_ready.append(start_end_date_script)
+
+        # Set auto updation of end_date based on start_date if start_field attr are set
+        default_interval = self.default_interval
```
flavour: Why bring this into top-level scope since only used once - just use self.default_interval the one time it is used.

flavour:
Good, nearly there :)

modules/s3/s3widgets.py
```diff
@@ -1511,6 +1521,20 @@ def inject_script(self, field, value, **attributes):
...
         jquery_ready = s3.jquery_ready
         language = current.session.s3.language
+        # Start field selector
+        start_field = opts.get("start_field", None)
+
+        if start_field:
+
+            # Set the minimum end_date to the same as the start_date
+            start_end_date_script = \
+'''S3.start_end_date('%(start_field)s', '%(selector)s')''' % \
```
nursix: Does this work with datetimepicker?
That is because the S3DateTimeWidget uses datepicker only if explicitly requested.

modules/s3db/event.py
```diff
@@ -216,6 +216,7 @@ def model(self):
                                  ),
                      s3_datetime("end_date",
                                  label = T("End Date"),
+                                 start_field = "event_event_start_date",
```
nursix: I'm not sure that this makes sense everywhere. Looks like you're applying it all over the place without enough information about the requirements - which could actually confuse users rather than helping them.

nursix: We also pay (in the sense of "processing time") for the script injection without any actual requirement.

nursix:
Really - having this as an option for the widget(s) is fine, but applying without need, just because we can apply it, is rather counter-productive (in terms of efficiency), and can be harmful where users don't expect it.

I'd recommend to focus on the original use-case and do a good documentation so that it can easily be configured for other cases if/when needed - and leave unrelated cases untouched.

nursix:
And......by the way: S3DateTimeWidget has such functionality built-in, there is really no need to introduce yet another option for it.

modules/s3/s3widgets.py
nursix: Actually, there is already an option for S3DateTimeWidget to link the fields like this. Your addition does interfer with this, and should actually not be introduced here.

modules/s3db/hrm.py
```diff
@@ -1498,8 +1500,8 @@ def model(self):
                              ),
                      s3_date("end_date",
                              label = T("End Date"),
-                             widget = S3DateTimeWidget(hide_time=True,
-                                                       set_max="hrm_salary_start_date",
+                             widget = S3DateTimeWidget(hide_time = True,
+                                                       set_max = "hrm_salary_start_date",
```
nursix: This ^ is how it is done in S3DateTimeWidget. "set_max" and "start_field" do basically the same thing. So, you do not need to introduce a new way to do it.

nursix: But as you can easily see - you can well remove the widget = S3DateTimeWidget for this particular case, and use s3_date with start_date (i.e. regular datepicker, not datetimepicker without time). 

Because...this was using S3DateTimeWidget only for the reason that the min/max option did not exist in S3DateWidget, and now that this option exists in S3DateWidget, this is not necessary anymore.

raj454raj: Okay :+1:

modules/s3/s3widgets.py
raj454raj: I get that :)

modules/s3db/event.py
raj454raj: @nursix : What do you suggest for this then?

nursix:
I think you should revert the change in S3DateTimeWidget (as that already has such an option). 

Instead we could expose the existing option as parameter for s3_datetime() - but that's a different issue.

modules/s3db/event.py
raj454raj: What are the models for which this should be applied? This would be tough to decide. I am not able to understand why is it a problem wherever this occurs? @nursix

nursix: I'm suggesting that there isn't necessarily always a problem that requires such a functionality, and that sometimes the problem can not be solved so easily.

This script does not work with inline-forms (or otherwise forms which work with different field names), since you're hardcoding the selectors - and hence applying it indiscriminately everywhere leads to unpredictable side-effects.

It is good to have the option, but I don't think that applying it just *everywhere* is the right thing to do. In most of these cases, people have either not entered any end-date before the start-date simply because they understood as much by themselves - nor has that had any negative impact so far.

I'd be more careful with this - it's not universally applicable, and users are not generally stupid and destructive.

raj454raj:
Need change lot of files for the merge conflicts. Will be sending new PR. Sorry!

modules/s3db/event.py
nursix: @raj454raj - the problem is that this option assumes a particular HTML element id. That will work with default forms, but we are increasingly using custom pages with custom forms and custom widgets - so that we often end up with different HTML element ids.

In those cases, the start_end_date will not work properly and can even have side-effects on other page elements, or other scripts that are injected into the page.

So, what I mean is that you need to check very carefully on a case-by-case basis (and across all the different pages in all templates which load that same form) whether it actually works without undesirable side-effects - rather than applying it indiscriminately everywhere.

That applies for *every* widget application, not just this one - it is a mistake to conclude from one case to another without explicit validation against requirements. 

The fact that this script has been around for so long, yet has only been used in a few occasions doesn't suggest negligence but rather that there hasn't been the requirement to use it more commonly - or, in fact, the script is not as universally applicable as it seems. Since it doesn't easily adapt to form variations (especially due to hardcoded element IDs), I think it indeed isn't so universal and flexible, let alone failsafe.

raj454raj: True! But do we want the developer to add the scripts in every controller wherever the model was used - This was the problem we discussed previously. Checking by case-by-case would be really tedious task to do. So what do you expect from my next PR for revision to this one(If required)? @nursix

nursix: Good question :$ sorry for not being more constructive...

No, we do not want this to be injected in controllers, but rather have this to be a model option - but as it says "option" it means "where suitable" not "everywhere as standard".

What I would expect after what I have said is:
a) that you improve the start_end_date script so it would work both in standard forms as well as in inline-forms without need to alter the setting.
b) that you investigate each case that you identified for whether any template uses a custom form, and if so, then adapt that template so the script does not interfere with that custom form (this will be very little to change, but a lot to investigate indeed).
c) that you include the caveats in the documentation of the feature
d) that you consider how this can generally be introspected rather than having the element IDs hardcoded (that is possible! Just *think* about it ;) )

nursix: Maybe start in reverse order ;)

raj454raj: @nursix : ;) thanks :+1: will look into it.
Sorry for asking you here, can my other PR be merged followed by further refactoring?
 You didnt told me what are the next changes I have to do. I'll be writing unit_tests for this but it will take time.

nursix: I'm sorry - I'm not keeping track of PRs, so I don't know what "your other PR" is.

If something isn't answered yet, then that is simply /because/ you have submitted multiple PR's - and since I can only deal with so and so many parallel threads (sorry, I'm only human), it is very likely I will only look at one at a time. Close one, and I will look at another ;)

raj454raj: Yeah sorry for not providing the link. Actually I didnt wanted to refer it from this. But anyway https://github.com/flavour/eden/pull/1034

raj454raj: @nursix ^

nursix: Am I too slow responding? 20 minutes is really very tight to respond to the question "what next", don't you think?

raj454raj: Sorry!

PR #1045, opened 2015-01-20
S3Represent for dc_collection and other Todos

gauravmittal1995:
Representation for dc_collection (template, location, date)
default to logged in person and root_org in dc_collection

### DIFF
--- a/modules/s3db/dc.py
+++ b/modules/s3db/dc.py
@@ -251,6 +251,7 @@ class DataCollectionModel(S3Model):
 
         T = current.T
         db = current.db
+        auth = current.auth
 
         crud_strings = current.response.s3.crud_strings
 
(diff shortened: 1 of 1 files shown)


nursix:
Thanks for the effort, but sorry - it's not /that/ easy. 

The location must use the location-represent ;)

For the other changes, as the status-tag says: WIP (work in progress), I am currently working on this module, but the requirements development is incomplete - so the todos can't be actioned yet (it's an upcoming project, not used yet).

I'd merge the logged-in user default, but not the other changes just yet.

nursix:
Note that the data model may change significantly - so polishing isn't worthwile yet. I intentionally left these as todo's to not put unnecessary efforts into things which may vanish in subsequent versions.

gauravmittal1995:
@nursix I see ... Okay :) , so should i just give the default logged in person and root_org?

nursix:
Actually, I have already done that in my story branch locally - so no need for you to do that.

Again - many thanks for the effort. I know that branch is overdue - but has been delayed due to other priorities.

nursix:
That is btw the reason why I put in that status-tag (it's not so obvious, I know) - I have a local story branch which is a number of commits ahead, hence other work would necessarily conflict unless I can coordinate it.

Just consider all todos in this particular module as either done or obsolete - I think I'll push that soon (not today though).

gauravmittal1995:
sure!! :)

PR #1047, opened 2015-01-21
Register Bug-fix

hitesh96db:
The register form doesn't work with "foundation_inline" as the formstyle. So, To-Do = Fix ;)

### DIFF
--- a/modules/s3/s3aaa.py
+++ b/modules/s3/s3aaa.py
@@ -1072,15 +1072,24 @@ Thank you"""
         labels, required = s3_mark_required(utable)
 
         formstyle = deployment_settings.get_ui_formstyle()
+        REGISTER = T("Register")
+        buttons = [INPUT(_type="submit", _value=REGISTER),
+                   A(T("Login"),
+                     _href=URL(f="user", args="login"),
+                     _id="login-btn",
+                     _class="action-lnk",
+                     ),
+                   ]
         current.response.form_label_separator = ""
         form = SQLFORM(utable,
                        hidden = dict(_next=request.vars._next),
                        labels = labels,
                        separator = "",
                        showid = settings.showid,
-                       submit_button = T("Register"),
+                       submit_button = REGISTER,
                        delete_label = messages.delete_label,
                        formstyle = formstyle,
+                       buttons = buttons,
                        )
 
         # Identify form for CSS & JS Validation
@@ -1089,23 +1098,6 @@ Thank you"""
         if js_validation:
             # Client-side Validation
             self.s3_register_validation()
-
-        login_link = A(T("Login"),
-                       _href=URL(f="user", args="login"),
-                       _id="login-btn",
-                       _class="action-lnk",
-                       )
-
-        # @ToDo: Probe formstyle rather than hardcode options here
-        formstyle_name = deployment_settings.ui.get("formstyle")
-        bootstrap = formstyle_name == "bootstrap"
-        foundation = formstyle_name == "foundation"
-        if bootstrap:
-            form[0][-1].append(login_link)
-        elif foundation:
-            form[0][-1][0][1].append(login_link)
-        else:
-            form[0][-1][0].append(login_link)
 
         # Insert a Password-confirmation field
         for i, row in enumerate(form[0].components):


nursix:
Hmm - but the register form should/does not use "foundation_inline". So, no fix needed.

However, this is definitely the cleaner way to inject the buttons - so fine to merge it, provided you have *tested* it across all templates. Have you?

nursix:
Don't get me wrong - it is much more likely that the old version fails than the new one ;) and yet I have to ask whether it is tested because it is such a critical and prominent feature.

The devil is in the details here: if you inject buttons like this, they will be wrapped in an additional DIV (unlike in the old version) - and that can break a bunch of CSS rules, especially in the pixel-pusher bootstrap themes.

hitesh96db:
I tested all the formstyles on a template(SSF), not this formstyle on all templates. This ticket showed up while I was trying to get rid of Bootstrap and use foundation instead, for Sunflower.
Initially I thought of a different solution:
- use row.element search for a INPUT tag with type "submit" which can be done in this "for loop" <a href='https://github.com/flavour/eden/pull/1047/files#diff-490b78a7772eee613a7f16277d7ffe14R1103'>here</a>
- then append the login_link to its parent....

Would this solution be more stable ?
But then I looked at how the login form was operating and changed my idea...

nursix:
No, your solution here is both the intended standard method /and/ the most appropriate solution. I had done the same change in the same way before, but then backed out for CSS problems in bootstrap themes.

You do not need to test the "foundation_inline" formstyle on all templates, no.

What I mean is that you need to test the button injection with all templates with *their* respective formstyles to see whether the template-specific CSS for this particular form doesn't break. And there is form-specific CSS!

---
On the other note:

To switch Sunflower to foundation should be as easy to start with as to just use the default theme with the default formstyles, with just a custom homepage. Then, later (if necessary) branch off and customize it.

...although the idea of "eating our own dogfood" is to make Sunflower feed back into mainline Sahana, which is obviously the less realistic the more custom stuff it uses - so a custom theme is actually counter-productive.

I think the main barrier is that people think they should retain the specific look&feel of the SSF theme - in spite of switching to foundation as framework. But that doesn't seem to make much sense - overriding the framework-CSS only to make it look like another framework - effectively replicating bootstrap theme on top of foundation. What's the point of switching frameworks then?

To make it harder?

nursix:
You do actually not need to test the foundation_inline formstyle for the registration form at all - it is simply neither intended nor used for that form.

But you do need to test the change with all other formstyles that *are* used for this form, and check that the template-specific CSS for *this particular form* doesn't break.

nursix:
...and note that your tests should *not* result in you backing out this change, but actually in fixing the respective CSS ;)

hitesh96db:
Okay....
Btw I'm making use of the CRMT css files and layout files for Sunflower, which looks much better than the default theme.So for now, I just changed the css.cfg to include the CRMT css files, which I should move into SSF as well ?

nursix:
Whether CRMT looks better is a question of taste ;) that judgement is not universal.

Technically, CRMT is a mess - and not recommendable for re-use. Too much custom stuff, and extremely fragile form layouts. The default theme is much more robust - and introspective, hence flexible.

Of course, certain details could be picked from CRMT - the colouring, the gutting, the button style. 
But else, the CRMT theme is a point solution and hence incomplete - it does not support all features in Eden. Certain features do not work at all due to layout conflicts.

So I can only recommend you to refrain from taking it wholesale .

As I said before: I would start with the default theme, and then cherry-pick the nice parts. And eventually you must really ask the question: where is the point with "eating our own dogfood" if Sunflower uses a custom (=non-default) theme?

nursix:
And be careful with "fashionable" - it was only a few months ago when the current SSF theme was declared *the* "most beautiful" Sahana theme in the world.

Now it's CRMT that is traded as the top of beauty, but very soon we'll have more designer work (on another theme), and then CRMT will go out of fashion again - and could well be called "ugly" just like you call the default theme ugly now.

nursix:
Technically, however, the default theme is the top of the line - the best supported, best integrated and most robust theme. And that is what I recommend it for, not for it's beauty ;)

flavour:
+1 to SSF template using the default theme (base, can still have a few extras in an additional local style.css) as the best way to eat dogfood.
This /can/ be used as a testing ground to port in some speciifc CRMT theme features which are desired, but as Dominic says, the CRMT theme as a whole has some major issues preventing this being used as default....apart from the non-universal taste issue:
* Forms have lots of manual width settings, so non-introspective => not suitable for default template => not suitable for SSF template
* The Summary page layout has the filters only visible on the Table tab, not the Report or Map tabs.

hitesh96db:
@nursix Tested it with all the templates, and things only get better when 'Register' and 'login' are inside a div - they are displayed inline.

nursix:
Ok, good to merge then :+1:

PR #1048, opened 2015-01-22
Added S3.end_date_interval widget option to S3DateTime

gauravmittal1995:
Adding of S3.end_date_interval widget to S3DateTime.
This is not the complete code
Right now, only the implicit and explicit default interval option is working. 
The interval buttons are not displayed (probably getting overridden by datetimepicker's button panel).

@nursix Suggestions?

### DIFF
--- a/modules/s3/s3fields.py
+++ b/modules/s3/s3fields.py
@@ -1343,6 +1343,9 @@ def s3_datetime(name="date", **attr):
             widget = "date" (in addition to usual meanings)
             past = x hours
             future = x hours
+            start_field = "selector" for start field
+            default_interval = x months from start date
+            default_explicit = Bool for explicit default
 
         @ToDo: Different default field name in case we need to start supporting
                Oracle, where 'date' is a reserved word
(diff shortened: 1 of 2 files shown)


nursix:
Doesn't make sense to me, sorry - S3DateTimeWidget is to select a time point to the minute, so not typically used to select month-long (or any pre-determined) intervals. I think you're getting carried away a little bit.

This doesn't have a use-case and only complicates the widget class unnecessarily, so -1 to this.

gauravmittal1995:
OK

nursix:
Don't feel too bad about it ;) You have no idea how many of my ideas I had to back out in order to de-complicate code.

But there is an intrinsic logic to that: any piece of code that is used very frequently will also be maintained - bugs get detected and fixed, documentation is kept up-to-date, the code is regularly upgraded to new versions of 3rd party libraries...all that stuff.

Code that is never used, won't be maintained either - bugs will go undetected, versions won't be upgraded, documentation goes out of date, and eventually it will turn into a rich source of regression issues and maintenance hurdles, so we do well not to introduce it in the first place. 

Moreover, it can very well happen that by the time a use-case for that code finally turns up, the functionality is so terribly out-of-date that it needs to be re-written. Or we have forgotten about it altogether. And exactly that has happened many times before, so requiring an actual use-case has become somewhat a matter of policy - though I am still relatively negligent about that. Fran's a lot tougher when it comes to "potentially useful".

gauravmittal1995:
@nursix I understand. I have a question:
what of this todo? https://github.com/flavour/eden/blob/master/modules/s3db/pr.py#L3193#L3197
Should there be a autofill widget??

PR #1049, opened 2015-01-23
Repeatability

AnirudhTiwari:
Tests are made repeatable by executing the following steps in order:

1. A copy of the database is made during the test build as the original database might change while the test runs.
2. The original database is restored from the copy during the test tear down, thus making the tests repeatable. 

Note: The test repeatability flag is off by default and can be turned on in the 000_config.py file. Also, the repeatability feature is currently only implemented for sqlite and postgres database.

### DIFF
--- a/modules/s3cfg.py
+++ b/modules/s3cfg.py
@@ -631,6 +631,12 @@ class S3Config(Storage):
 
         return self.base.get("allow_testing", True)
 
+    def get_base_allow_repeatability(self):
+        """
+            Allow tests to be repeatable
+        """
+        return self.base.get("allow_repeatability", False)
+
     def get_base_migrate(self):
         """ Whether to allow Web2Py to migrate the SQL database to the new structure """
         return self.base.get("migrate", True)
--- a/modules/templates/000_config.py
+++ b/modules/templates/000_config.py
@@ -65,6 +65,8 @@ settings.base.migrate = True
 # ?debug=1
 settings.base.debug = False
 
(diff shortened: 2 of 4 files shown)


nursix:
There's a couple of problems here:

1) For ATDD, you would typically run only a few individual tests - perhaps just a single one - many times. That is exactly the scenarion where you need repeatability, but copying the whole database just to execute a single test is really unreasonable processing effort and contradicts the requirements for tests to be fast.

2) A deployment setting doesn't make sense to me here - it should rather be a call parameter for the test suite (which defaults to True).

3) There is little point in copying the local database if the target URL points to a remote server.

4) No rollback for the case where the test suite is interrupted - but for test runs, especially for ATDD, this is quite a common scenario. If I see something failing, I will ctrl-c it early.

5) No support for MySQL, but this is required as it is one of our standard databases (I am developing with MySQL, for instance).

6) Smoke tests don't need this - it would only unnecessarily slow them down.

7) OS commands and file system access can fail for many reasons (permissions, duplicate filenames, capacity limits...) but you're hardly handling any error at all.

...and of course commits should generally be squashed together into one unless they really target separate issues.

I'm not convinced of this solution in general - I've had long discussions with Arnav S. about that before. 

The ideal would be to find a way to roll back transactions after individual test cases, but I appreciate that this may be very hard. So I could accept that as an interim solution if the problems named above can be solved.

nursix:
Besides: this does entirely ignore our developers on Windows.

PR #1051, opened 2015-01-24
Minor tweaks

raj454raj:


raj454raj:
Okay what about datetime is it fine?

PR #1052, opened 2015-01-24
Filter Widgets for event_site, event_asset, event_human_resource

gauravmittal1995:


gauravmittal1995:
@nursix  Please Review

modules/s3db/event.py
```diff
@@ -2017,6 +2016,24 @@ def model(self):
...
+                                        ],
+                                       label = T("Search"),
+                                       ),
+                          S3OptionsFilter("status",
+                                          label = T("Status"),
+                                          hidden = True,
+                                          ),
+                          ]
+
+        lf = list_fields
```
nursix: Still don't get the point with these additional variables - they don't seem necessary?

modules/s3db/event.py
```diff
@@ -2017,6 +2016,24 @@ def model(self):
...
+                       "site_id",
+                       "status",
+                       ]
+
+        filter_widgets = [S3TextFilter(["site_id$name",
+                                        ],
+                                       label = T("Search"),
+                                       ),
+                          S3OptionsFilter("status",
+                                          label = T("Status"),
```
nursix: There is no need to provide a label if the field has a label. In fact, it is often better to stick with the field label.

modules/s3db/event.py
```diff
@@ -2030,20 +2047,27 @@ def model(self):
                                                                        ],
                                                              ),
                                         )
+
+            lf.extend(("allocation.budget_id",
```
nursix: The might become a bit messy, tbh - there can be multiple budget allocations for a site, and then it'll be hard to see which allocation belongs to which start date, or which end date belongs to which start date. Shouldn't be separate fields - rather have a represent for the allocation.id and only list those as a single field.

modules/s3db/event.py
```diff
@@ -2030,20 +2047,27 @@ def model(self):
...
+                       "allocation.start_date",
+                       "allocation.end_date",
+                       "allocation.daily_cost"
+                       ))
+
+            fw.extend((S3DateFilter("allocation.start_date",
```
nursix: Questionable that too - why would we want to filter sites by budget allocation start date? More likely that we would want to filter by sponsoring organisation (=budget owner), or by whether there is budget allocated to the site at all.

nursix:
I think this needs proper use-case research before hacking in just anything. The budget-related filters for event-sites don't quite make sense to me yet.

Filtering by site name and status is fine, however - that's more obvious.

Also, the fact that the deployment has budgets enabled doesn't necessarily mean that is used for event-sites (it's an obvious use-case, but not a mandatory one). We may need to develop the requirements for this linking first.

The budgeting module has basically just been salvaged, and the requirements for it's use are still rather rudimentary. We do need practitioner input on that.

modules/s3db/event.py
gauravmittal1995: @nursix Sorry, i dont understand.... Represent allocation.id as what?

gauravmittal1995:
@nursix So should we create a deployment setting for the use of budget in event-site???

nursix:
As I say: before we go and implement just anything, we should conduct a proper use-case research, and develop appropriate requirements for this. Right now this looks very much like a "shoot first, ask later" strategy.

We can of course speculate how this case might be, but I seriously doubt that this will lead to a practically useful feature set. It would be much better to get some domain input first (i.e. from potential users).

modules/s3db/event.py
nursix: Obviously represent the allocation.id as what it is: a budget allocation. That might include the time interval, the sponsor and the total allocated budget - or it could also be the budget name (although I doubt that makes a lot of sense) plus start/end dates. I don't know.

As I keep stressing: the requirements for this case are incomplete/unclear, we do need input from potential users. Trying to implement anything without proper understanding of the use-case and its requirements makes very little sense (and is a waste of time, tbh).

At the moment, we can do very little here.

gauravmittal1995:
@nursix I see, so i will keep the list fields same as before and add only the filters for site_id and status. And add a todo to add filter widget for budget filtering

gauravmittal1995:
@nursix Removed the filter widget for the budget, still displaying all the fields as it were (including budget). Added Todos for budget filters.
Also added filter for event_human_resource and event_asset
Please review

gauravmittal1995:
@nursix Hey, So this code still needs re-factoring and also the widgets for budget, but logic wise, is this the implementation that u wanted?

PR #1053, opened 2015-01-25
Refactoring survey module

raj454raj:


raj454raj:
Also should I make get_style_list a staticmethod?

raj454raj:
@nursix : What's step 3 ?

raj454raj:
@nursix: Please review

raj454raj:
@nursix : Wrote tests for series_ExportFormatted. What is step 3. Please reply so that I can start working on it.

nursix:
Referring to our previous conversation - step 3 was to eliminate camelCase names and other non-standard coding styles.

nursix:
Coding convention: function/variable names are all-lowercase with underscore as word separator, class names start uppercase and use CamelCase for word separation (this does not apply for the mandatory module prefix "survey_", however), (pseudo-)constants are all-uppercase with underscores as word separator (LIKE_THIS).

Variable/function/class names should be self-explanatory and indicate the purpose of the variable, cryptic abbreviations like "sftgv" aren't useful.

Further, you can use pylint to evaluate code quality, a rating above 7.5 is desirable.

The unit tests help you to ensure you don't break things when refactoring the code (provided that they are specific enough). You should run them after every little step - over and over again. Consider edge-cases and unhandled failures.

Finally, conduct some testing through the GUI (sometimes all unit tests pass, yet the user-facing functionality is broken).

nursix:
Also, remember: "Be destructive with testing, and constructive with error handling".

raj454raj:
@nursix : Okay just a request, can changing the names done in another PR(Please) - Because this PR is tested with almost all the functionalities and is working. And what about get_style_list? (Static or not).

nursix:
Your call - no self-use implies static, normally.

If you want to do a separate PR for the cleanup then fine - but I haven't reviewed this one yet (I am actually trying to work, so I don't have the time right now). I can probably do so later tonight, and until then, in fact, you could do a little refactoring already.

nursix:
Or, you can help me with something (doing some research) - but we should take that off this conversation.

raj454raj:
Okay sure, where then?

nursix:
Email - send me an email so I get your email address, and can reply to it

modules/s3db/survey.py
```diff
@@ -271,10 +277,12 @@ def model(self):
         tablename = "survey_section"
         define_table(tablename,
                      Field("name", "string", length=120,
+                           label = T("Name"),
```
nursix: That's actually the default label. If you don't set a label, web2py will take the name, capitalize and translate it. So, not needed, actually.

modules/s3db/survey.py
```diff
@@ -271,10 +277,12 @@ def model(self):
...
                            notnull = True,
                            default = "",
                            ),
                      Field("description", "text", length=500,
+                           label = T("Description"),
```
nursix: Same here.

modules/s3db/survey.py
```diff
@@ -1027,10 +1031,21 @@ def question_onvalidate(form):
 
         from xml.sax.saxutils import unescape
 
-        metadata = form.vars.metadata
-        if metadata != None:
+        vars = form.vars
```
nursix: Use form_vars not vars (vars is a Python keyword)

modules/s3db/survey.py
```diff
@@ -1599,17 +1614,17 @@ def model(self):
         tablename = "survey_series"
         self.define_table(tablename,
                           Field("name", length=120,
-                                default = "",
                                 label = T("Name"),
+                                default = "",
```
nursix: Sort options alphabetically - hence "default" before "label"

modules/unit_tests/s3db/__init__.py
```diff
@@ -1,3 +1,6 @@
 from pr import *
 from org import *
 from vulnerability import *
+from inv import *
```
nursix: You just added these tests to the test suite without actually checking whether they are valid?

modules/unit_tests/s3db/__init__.py
```diff
@@ -1,3 +1,6 @@
...
+from survey import *
+from hrm import *
```
nursix: Same here - are these test cases valid?

modules/unit_tests/s3db/survey.py
```diff
@@ -0,0 +1,116 @@
...
+from s3survey import DataMatrix
+
+# =============================================================================
+class ExportFormattedTests(unittest.TestCase):
+    """ Test series_ExportFormatted """
+
+    # -------------------------------------------------------------------------
+    def setUp(self):
+        """
+            Overrides setUp method of TestCase class
```
nursix: Note "overrides" - "implements"! It's an abstract method that is *meant* to be implemented in test cases.

modules/unit_tests/s3db/survey.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        current.auth.override = True
+        db = current.db
+        s3db = current.s3db
+        ttable = s3db.survey_template
+        stable = s3db.survey_series
+
+        # Insert a test record for survey_template
+        record = Storage(name = "Test template",
```
nursix: Can use dict, Storage not needed (generally should try to avoid using Storage unless you really need it)

static/styles/ui/fgtimepicker.css
```diff
@@ -32,7 +32,7 @@
 /* anchors for clickable cells */
 .ui-timepicker-table td a {
     display: block;
-    padding: 0.2em 0.3em 0.2em 0.5em;
+    padding: 0.2em 1.5em 0.2em 0.5em;
```
nursix: Not sure what this is for? Better don't mix in unrelated CSS changes - or provide a screenshot for the change.

modules/unit_tests/s3db/survey.py
```diff
@@ -0,0 +1,116 @@
...
+                                                  logo)
+        self.assertNotEqual(output, None)
+        
+    # -------------------------------------------------------------------------
+    def testGetStyleList(self):
+        """
+            Test get_style_list()
+        """
+        style_list = self.S.get_style_list()
+        self.assertTrue(isinstance(style_list, dict))
```
nursix: That is very little to verify the method, isn't it?

modules/unit_tests/s3db/survey.py
```diff
@@ -0,0 +1,116 @@
...
+                                                               dict(),
+                                                               justified = True)
+        self.assertTrue(isinstance(matrix, DataMatrix))
+        self.assertTrue(isinstance(matrixAnswers, DataMatrix))
+
+        
+        output = self.S.series_export_spreadsheet(matrix,
+                                                  matrixAnswers,
+                                                  logo)
+        self.assertNotEqual(output, None)
```
nursix: A lot of effort for so little Assertion - is that really enough to verify the method?

modules/unit_tests/s3db/survey.py
```diff
@@ -0,0 +1,116 @@
...
+                      args = [self.s, "export_formatted_spreadsheet"],
+                      )
+
+        logo = None
+        (matrix, matrixAnswers) = self.S.series_prepare_matrix(self.series_id,
+                                                               r.record,
+                                                               logo,                                 
+                                                               dict(),
+                                                               justified = True)
+        self.assertTrue(isinstance(matrix, DataMatrix))
```
nursix: Not testing failure response in any way, and very little verification. Basically just checking whether it doesn't crash? For that though, you must at least test all possible cases (look at the "if" branches in the target function).

nursix:
Looks like good progress - but the unit tests are rather weak to really detect regression. As a minimum you must test for failure handling, and touch each (relevant) "if" branch at least once.

You shouldn't add unrelated test cases to the suite - unless you have validated them. I had originally removed them because I found them invalid - and I don't see that you changed anything that would make them valid now?

You're mixing in unrelated changes now - which makes it a bit difficult to verify. You should either add some information about those to the PR description, or maybe even put them in a separate PR (though I see they are fairly small, so maybe not worth a separate commit, but having some background what they are about would be very helpful).

static/styles/ui/fgtimepicker.css
raj454raj: screenshot from 2015-01-30 00 22 15

raj454raj: screenshot from 2015-01-30 00 22 43

raj454raj:
Will be adding more tests. Thanks for the review.

modules/unit_tests/s3db/inv.py
```diff
@@ -50,7 +45,7 @@ def run_suite(*test_classes):
 if __name__ == "__main__":
 
     run_suite(
-        StatsTests,
+        InvTests,
```
raj454raj: But still there are no tests written here.

raj454raj:
Hey @nursix need some help - 
There's a bug that I found while I was working on survey module,
Add a template(Make sure you enter some value in the name of the questions field). Make a series using this template. Now if I enter some data in "Enter Completed Assessment" -> it adds the answers in the survey_answer table, BUT survey_answer.question_id is not the actual question_id that we actually added in survey_question while creating the template. It is just sequential numbers 2,3,4... Because of which there are problems whose consequences are bad.
So what I found is here -
https://github.com/flavour/eden/blob/master/modules/s3db/survey.py#L2890
form_vars contains question_id as 2,3,4 ... which is not required (It should be something in 1500s - according to present database prepop)
So there is no where in the form that we are passing question_ids(Only codes are made ids of the fields) here -
https://github.com/flavour/eden/blob/master/modules/s3db/survey.py#L736
So we need a way so that we can get correct question_ids in answer_onaccept. Please help.

nursix:
I'd need to investigate this, but don't have the time right now. Maybe later tonight?

nursix:
If it comforts you: finding such critical bugs is one of the reasons for the refactoring, so that makes your work successful. Well ok - of course we also need to fix them...sure, sure...

raj454raj:
So what I am telling is - I didnt even started refactoring the module. This PR actually contains the working code of downloading .xls and .rtf which is not in the main trunk right now. I don't think tests should be stopping this PR from merge(Since we need working code before refactoring).
> Well ok - of course we also need to fix them

So I think I will first make the module working(As it really isn't) and refactoring can follow. Shall I remove the tests from this PR? @nursix

nursix:
Hmm, there's a gap in your logic here: without the tests, how do you know it's working code?

I would say exactly the other way around: we first need the tests, then the working code - and then we can start refactoring while using the same tests to verify the changes.

nursix:
Or to make that a little clearer: we may have to insist on evidence that your code here actually works, so (conclusive) test cases are actually required before merge.

raj454raj:
:+1:

nursix:
I've investigated the issue with the question ID, and it turns out that this is a major flaw in the overall design of this module:

When you generate a template, it does automatically insert some standard questions:
https://github.com/flavour/eden/blob/master/modules/s3db/survey.py#L409

The question codes for these standard questions are hard-coded: STD-WHO, STD-DATE etc.

When you submit the answers for these questions, it would first write them together with the question code into a temporary CSV file, and then import them:
https://github.com/flavour/eden/blob/master/modules/s3db/survey.py#L2766

The import would find the question through the question code, using the deduplicator:
https://github.com/flavour/eden/blob/master/modules/s3db/survey.py#L1062

However, since the codes are hardcoded, they would be the same in all templates that have been created through the GUI! Consequently, all answers would always be linked to the first question with the respective code, even if that's not in the corresponding template. Correctly, the question match would additionally require the template_id.

But this is just the tip of the iceberg - if you look a little closer, then there are hundreds of issues in this part of the process, ranging from DB writes during GET over invalid data not being caught, forms not respecting formstyle, errors neither being caught nor reported...to completely non-conformity with coding standards and a terrible interlocking of process elements.

The latter means: this only looks like separate functions, but actually it's a single monstrous sequence of incomplete steps with processing dependencies that span from the very end to the very top (i.e. even a valid partial result can only be achieved by executing non-consecutive steps, which basically defeats the purpose of splitting into subfunctions).

Fundamental flaws like this one can therefore not easily be "fixed" - the domino-effect of this code structure will force you to more or less rewrite the whole thing.

...which I can only encourage you to do, though.

nursix:
i.e. the  issue with the question_id is that this uses a pseudo-key which is not actually a key for the question table and therefore ambiguous. Re-design required.

PR #1054, opened 2015-01-25
Filter Widgets for event_event_type and event_event

gauravmittal1995:


modules/s3db/event.py
```diff
@@ -265,33 +272,43 @@ def model(self):
         # Which levels of Hierarchy are we using?
         levels = current.gis.get_relevant_hierarchy_levels()
 
+        filter_widgets = [S3TextFilter(["name",
+                                        ],
+                                       label = T("Search"),
+                                       )
+                          ]
+
+        fw = filter_widgets
```
nursix: What the point with this additional variable?

modules/s3db/event.py
```diff
@@ -265,33 +272,43 @@ def model(self):
...
+                                    levels = levels,
+                                    label = T("Location"),
+                                    hidden = True,
+                                    ),
+                   # @ToDo: Filter for any event which starts or ends within a date range
+                   S3DateFilter("start_date",
+                                label = T("Date"),
+                                hide_time = True,
+                                hidden = True,
+                                input_labels = {"ge": "From", "le": "To"}
```
nursix: That's the default

modules/s3db/event.py
```diff
@@ -265,33 +272,43 @@ def model(self):
...
+
+        fw = filter_widgets
+
         if hierarchical_event_types:
-            filter_widgets = [S3HierarchyFilter("event_type_id",
-                                                label = T("Type"),
-                                                #multiple = False,
-                                                ),
-                              ]
+            fw += [S3HierarchyFilter("event_type_id",
```
nursix: Using append() is faster than +=

modules/s3db/event.py
gauravmittal1995: Sorry, i dont understand ... what is default?

nursix: These input labels.

modules/s3db/event.py
```diff
@@ -265,37 +272,48 @@ def model(self):
...
 
         filter_widgets.extend((S3LocationFilter("event_location.location_id",
                                                 levels = levels,
                                                 label = T("Location"),
+                                                hidden = True,
                                                 ),
                                # @ToDo: Filter for any event which starts or ends within a date range
                                S3DateFilter("start_date",
-                                            label = None,
+                                            label = T("Date"),
```
gauravmittal1995: @nursix I see, So should i remove the whole label option, or leave it as T("Date") ??

nursix: Huh? No - I said these input_labels are default, so they do not need to be specified.

gauravmittal1995:
@nursix  I have a question:
what of this todo? https://github.com/flavour/eden/blob/master/modules/s3db/pr.py#L3193#L3197
Should there be a autofill widget??

modules/s3db/event.py
gauravmittal1995: @nursix , Oh sorry, Didnt realise that! That was already there in the code, so i though u were reviewing the part which i changed ... will remove the input_labels

gauravmittal1995:
@nursix Removed the input_label, Please Review

modules/s3db/event.py
gauravmittal1995: @nursix Done

PR #1055, opened 2015-01-27
Bug fix

hitesh96db:
I tried:
        current.menu.options.disable(tag="project")
        current.menu.options.enable(tag="deployment") 
as suggested here: https://groups.google.com/forum/#!topic/sahana-eden/krBF4-q7Nho
But it gave a ticket, so this is a fix for it.


hitesh96db:
Or
<pre>
if item.authorized and (item.enabled is None or item.enabled):
    enabled = True
    visible = True
else:
    enabled = False
    visible = False
</pre>

hitesh96db:
Or
Should it use item.enabled and item.visible instead of enabled and visible ?

modules/s3layouts.py
```diff
@@ -231,6 +231,9 @@ def layout(item):
         elif item.enabled is None or item.enabled:
             enabled = True
             visible = True
+        else:
+            enabled = False
+            visible = False
```
nursix: Let's be reasonable:

    if item.authorized and (item.enabled or item.enabled is None):
        enabled = visible = True
    else:
        enabled = visible = False

;)

nursix:
NB I used enabled /and/ visible despite that's not actually needed here, because this layout was also intended as an example for how granular things can be controlled.

One could for example have items disabled *yet* visible (i.e. discoverable for everybody, but accessible only for authorized users or when certain system conditions are met), e.g. having a menu item for "Find People" without any people in the database - the menu item should still be there to inform the user that this function exists, but it can be disabled while there are no records.

I'd leave this in place for future use, even if it looks a bit superfluous right now.

modules/s3layouts.py
hitesh96db: :D

nursix:
(I'm telling you this because S3Navigation can also be used to control e.g. buttons or button groups in the Sunflower homepage/dashboard/menu - e.g. "My Tasks", which might be empty or the user not logged in, and hence disabled, yet should be visible to inform users that this functionality exists and can be used if...).

hitesh96db:
Makes sense :+1:

hitesh96db:
Added one more fix.

hitesh96db:
@flavour Please review.

PR #1056, opened 2015-01-27
Sunflower | Bootstrap => Foundation

hitesh96db:
Update: http://test.sunflower.sahanafoundation.org/
I've introduced a few styling changes(not significant ones) that differ from the default
- the side nav & the inline component color is a bit lighter.
- Added a background-color to the comments, it's hard to notice otherwise
- Added a bit of styling to the content header
- the rest is for the homepage

Thoughts ?


nursix:
Hard to tell without screenshots

modules/templates/SSF/config.py
```diff
@@ -24,38 +25,29 @@ def config(settings):
...
-    settings.ui.formstyle_row = "bootstrap"
-    settings.ui.formstyle = "bootstrap"
-    settings.ui.filter_formstyle = "table_inline"
-
-    # Uncomment to disable responsive behavior of datatables
-    # - Disabled until tested
-    settings.ui.datatables_responsive = False
+    
+    # UI Settings
+    settings.ui.formstyle = "foundation"
```
nursix: That's the default

modules/templates/SSF/config.py
```diff
@@ -24,38 +25,29 @@ def config(settings):
...
+    settings.ui.filter_formstyle = "foundation_inline"
```
nursix: That too.

modules/templates/SSF/config.py
```diff
@@ -24,38 +25,29 @@ def config(settings):
...
+    settings.ui.datatables_responsive = True
```
nursix: That too.

hitesh96db:
Side nav color, content header color difference
default
demo1
sunflower
test1
Inline Components color
default:
demo2
Sunflower + (comments background-color)
test3

modules/templates/SSF/css.cfg
```diff
@@ -1,47 +1,46 @@
...
+zocial/zocial.css
+../themes/default/gis.css
+../themes/foundation/map.css
+../themes/default/mobile.css
+../themes/default/print.css
+d3/nv.d3.css
+../themes/default/report.css
+../themes/default/survey.css
+../themes/default/style.css
+../themes/foundation/foundation-icons.css
```
nursix: I would recommend to stick with font-awesome.

Foundation icons have only a fraction of what font-awesome provides - and foundation icons are a bit harder to recognize.

Personally, I don't like icons at all but prefer pure text. In my understanding (which comes mostly from medical devices), icons are hieroglyphs and thus the user has to learn a hieroglyph script to navigate the application, instead of just using the language he already speaks fluently. And secondly, there is hardly any consistent let alone standard icon "language" - every application uses its own interpretation, and that makes it rather harder to work with icons than with text.

But if it by all means have to be icons, then font-awesome is still the better choice.

modules/templates/SSF/views/deployment_page.html
```diff
@@ -1,17 +1,17 @@
...
 
- <a class='btn btn-primary' href='/eden/project/project/{{=r.id}}' id='edit-btn'>Open</a>
- <a class='delete-btn btn btn-primary' href='/eden/project/project/{{=r.id}}/delete' id='delete-btn'>Delete Deployment</a>
- <a class='btn btn-primary' href='/eden/project/project?sector.name=None,Deployment' id='list-btn'>List Deployments</a>
+ <a class='button action-btn' href='/eden/project/project/{{=r.id}}' id='edit-btn'>Open</a>
+ <a class='button delete-btn' href='/eden/project/project/{{=r.id}}/delete' id='delete-btn'>Delete Deployment</a>
+ <a class='button action-btn' href='/eden/project/project?sector.name=None,Deployment' id='list-btn'>List Deployments</a>
  <div class='row'>
-  <div class='span6'>
+  <div>
```
nursix: Not making use of any columns-classes? Seems a bit strange.

hitesh96db:
@nursix btw you answered only one of the questions asked about the navigation ;), Let me repeat --

As soon as the filter comes off the url, the "highlighted" option in the Main menu changes from "Project" to "Deployment" even if I am in the "Projects" page, because there's no filter to tell it what page it is in - that's how I defined the MM() for Projects and Deployments - the two are distinguished based on the filter.

How can I resolve this ?

modules/templates/SSF/views/index.html
```diff
@@ -1,80 +1,79 @@
 {{extend "layout.html"}}{{theme=response.s3.theme}}
-<style>body{background:none;}div.container.main-container{width:100%!important;padding-left:0;}
-@media(min-width:1200px){div.container.main-container{width:1200px!important;}}</style>
 
-
-<section id='intro-section' style='padding-bottom:20px;'>
-  <div class='container'>
+<div class='row intro-row homepage-row'>
+  <div class='large-2 columns'>
```
nursix: "large" makes only sense if you have a different "medium" or "small". 

The principle is to design "mobile first", i.e. configure for small and/or medium, and then *add* an alternative large-setting if necessary - but not have only "large".

modules/templates/SSF/views/deployment_page.html
hitesh96db: I left the deployment pages as it is for now, I was told it has to be redesigned, so I didn't think about modifying it, I just removed the bootstrap classes.

static/themes/SSF/style.css
```diff
@@ -36,7 +58,84 @@
...
+}
+#wiki-develop {
+    font-size: 20px;
+    color: teal;
+}
+.about-row p {
+    font-size: 1.25rem;
+}
+.about-row h2 > i {
+    font-size: 40px;
```
nursix: Foundation typically using rem not px - which has advantages with proportional scaling (fixed pixel settings don't scale at all! rem does!)

static/themes/SSF/style.css
```diff
@@ -9,24 +29,26 @@
...
 }
-#feed-control span {
-	color: #676767;
-	font-size: 11px;
-	margin: 10px;
-	padding: 4px;
+.side-nav li > a:not(.button) {
+    background-color: #ebebeb;
+    color: #5d5d5d;
+    padding: 3px;
```
nursix: Should really be rem not px so it scales proportionally with the fonts. px is absolute, rem is relative.

nursix:
Not sure I understand your question re: filter coming off the URL...can you explain, or point to the respective code?

hitesh96db:
For eg:
http://test.sunflower.sahanafoundation.org/eden/project/project?sector.name=None%2CProject
In this page, "Project" in the Main menu is highlighted.
Click on 'open'
Now, "Deployment" gets highlighted even though you're in the projects page.

nursix:
It looks fine in general - but before it gets merged you should probably send another email to the mailing list (because your previous email introduced a different design, so people may not expect this one here).

Either point to this PR, or attach some screenshots, clarify that you have abandoned your previous proposal and this is your new draft - then give people a couple of days to comment and discuss before we merge this.

Ok?

hitesh96db:
Yes, that sounds good.

hitesh96db:
@nursix as long as the filter stays in the url, i.e, ?sector.name=None,Project/Deployment, the Main menu shows the correct part being highlighted, I guess it's because of this:
https://github.com/hitesh96db/eden/commit/a5014026b827f06f9e8ed421f942fa4bbdee49bc#diff-d11854e3068b88cd7e7c34112ff346c9L79 ?
For eg:
http://test.sunflower.sahanafoundation.org/eden/project/project?sector.name=None%2CProject
In this page, "Project" in the Main menu is highlighted.
Click on 'open'
Now, "Deployment" gets highlighted even though you're in the projects page.

nursix:
Ok - got you.

This is the automatic URL matching - S3Navigation matches the request URL against the item URL to determine which item to highlight. If two items match (like in this case), it picks the first match.

This is difficult to resolve because the paths are identical (project/project in either case). The best thing you can do is to set it manually at the point where you set the side menu. Tag the one menu item as "project", the other one as "deployment", and then set:

    if ...condition for deployment...:
        topnav = current.menu.main
        topnav.findall("deployment")[0].selected = True
        topnav.findall("project")[0].selected = False

This overrides the automatic detection for those items.

nursix:
well, plus the corresponding "else", of course ;) I think I don't need to say that, do I?

hitesh96db:
Nah ;), Thanks.

nursix:
Sure, if the URL still has the filter, then the "Projects" item is the better match (it's tight, but still...).

But the URL project/project/11 only matches up to "project/project" with either item URL - so 100% equal, hence it picks the first match. You can only manually override this.

There is an alternative, though (that is, generally speaking, not a recommendation) - you could also define a separate controller. E.g.:

    def project():
        return s3_rest_controller("project", "project")

    def deployments():
        return project()

Both controllers do basically exactly the same thing - but they have different URLs, and hence would match different menu items. But as I say - not a recommendation, but just a possibility.

nursix:
Btw - the shadow for the headline box contradicts the shadows in the forms. You have the light source at the top right, but the forms have the light source from above - which looks a bit strange.

I would remove the headline box shadow.

nursix:
Can't have one shadow on the page in another direction than other shadows on the same page ;)

You have a big desire for some artistic extravagance :D but I do have an artistic eye (I've been painting for money some time in my life).

nursix:
Uhm - and I see that the test instance doesn't yet have the latest default/style.css. Maybe you should update that before pointing to it (it has better support for mobile).

nursix:
And fewer colours on the homepage would be better IMHO. Too many shades which don't harmonize.

Try to reduce noise: fewer colours (or shades of the same colour), fonts and button styles, maintain consistency. Don't make button labels bold, they are already in a prominent box. Ideally remove the icons, they don't really help here.

Well, ok - so much for now as feedback on the design ;)

nursix:
Nah - ok, one more: rounded corners for buttons are out :D 

But at least they are inconsistent with the rest of the theme. Either you have rounded corners everywhere (like here on GitHub = bootstrap), or not at all.

hitesh96db:
@nursix I tried --
<pre>
if ...condition for deployment...:
    topnav = current.menu.main
    topnav.findall("deployment")[0].selected = True
    topnav.findall("project")[0].selected = False
</pre>
But it didn't work, i.e, if I set project to False, it still remains selected. 
I believe it's because of this - 
https://github.com/flavour/eden/blob/master/modules/s3/s3navigation.py#L480,
the value gets overwritten and set to True.
Maybe we can add a check there to see if it has been set before - to either True or False ?

nursix:
I have a better idea, I'll implement it then get back to you. Hold off this PR for noew.

hitesh96db:
Okay

nursix:
Yup, a little patience please, though (it's weekend, so I have some family appointments first). Will work on it later in the evening.

hitesh96db:
Take your time, no urgency here.

nursix:
Pushed. Once merged into trunk, you can do as follows:

    if ...condition for deployment...:
        current.menu.main.select("deployment")
    else:
        current.menu.main.select("project")

...though with the menu items in project=>deployment order, you can probably omit the else-branch.

hitesh96db:
Thanks :)

modules/templates/SSF/config.py
```diff
@@ -24,38 +25,29 @@ def config(settings):
...
     settings.auth.opt_in_default = True
     # Uncomment this to default the Organisation during registration
     settings.auth.registration_organisation_default = "Sahana Software Foundation"
-
     # Always notify the approver of a new (verified) user, even if the user is automatically approved
     settings.auth.always_notify_approver = True
-
-    # Base settings
-    settings.base.system_name = T("Sahana Sunflower: A Community Portal")
-    settings.base.system_name_short = T("Sahana Sunflower")
```
michaelhowden: Why are you removing these?

modules/templates/SSF/views/layout.html
```diff
@@ -58,103 +50,45 @@
...
+   </div>
+   {{pass}}
+  </div>
  </div>
 {{pass}}
-{{include "bootstrap_js.html"}}
 {{include "scripts.html"}}
+{{include "foundation.js.html"}}
 </body>
 </html>
```
michaelhowden: Why is a different modules/templates/SSF/views/layout.html needed? How is this different from modules/templates/default/views/layout.html? Could you remove the SSF layout and allow it to default to the default layout?

static/themes/SSF/style.css
```diff
@@ -1,3 +1,23 @@
...
+    padding: 36px 24px; } }
+@media only screen and (min-width: 64.063em) {
+  .homepage-row {
+    padding: 48px 100px; } }
+
+#homepage-title {
+    font-family: Trebuchet MS;
+}
+
+table.embeddedComponent,
```
michaelhowden: Suitable for default?

static/themes/SSF/style.css
```diff
@@ -9,24 +29,26 @@
...
     background-image: url(../../img/sunflower_spin_throbber.gif);
 }
-#logo {
-	width: 55px;
-	position: relative;
-	top: -19px;
+
+#content h2 {
+    background-color: #ebebeb;
+    font-family: Lato,"Helvetica Neue",Helvetica,Roboto,Arial,sans-serif;
```
michaelhowden: Why the different font?

static/themes/SSF/style.css
```diff
@@ -9,24 +29,26 @@
 .map_loader {
     background-image: url(../../img/sunflower_spin_throbber.gif);
 }
-#logo {
-	width: 55px;
-	position: relative;
-	top: -19px;
+
+#content h2 {
```
michaelhowden: Suitable for default?

static/themes/SSF/style.css
michaelhowden: Suitable for default?

nursix: -1 controlled by base foundation setting in default

static/themes/SSF/style.css
nursix: Definitely not.

static/themes/SSF/style.css
nursix: Yes, submit to edenthemes though so that all default-based templates can benefit

static/themes/SSF/style.css
nursix: Generally, wherever possible, foundation settings should be used for default, instead of extra CSS to override them. However, the higher contrast was intended, so -1 to this change in default.

nursix: Although...hang on - this is side-nav not top-nav, so no foundation setting. Sorry, /me confused.

nursix: Note that the original side-nav colour was a bit lighter, but then intentionally darkened for better contrast.

Since there is very little point to keep changing colours forth and back, I've decided to put all main colours into the branding package (and also to ensure that they are used consistently throughout). This branding package is a bunch of SCSS vatiables that a derivate theme can easily change to match its brand colours, basically looking like this:

https://github.com/nursix/edenthemes/blob/master/EVASS/scss/_colors.scss

For the default configuration I've used a colour wheel to determine an optimal yet simple triple of colours - however, the side-nav was set to be too light and should be darker. Then other people come and say make it lighter again...as I say: forth and back and forth and back. No real point with this.

Personally, I would use different shades of green - and I'm convinced I can find a bunch of people who would like it. And some who would find it disgusting. That's how it is, and why it is neutral grey by default.

static/themes/SSF/style.css
```diff
@@ -36,7 +58,84 @@
 #assign-container {
     margin-left: 1%;
 }
+.comment-box {
+    background: none repeat scroll 0 0 #f5f5f5;
```
hitesh96db: Should I add this to default/style.css ?

modules/templates/SSF/css.cfg
hitesh96db: Can I add new ones to icons.css or should it be specific to SSF ?

nursix: New ones? You mean icons which do not exist in font-awesome? Can you give an example?

static/themes/SSF/style.css
nursix: If you are SCSS-literate, I would prefer to get this as PR for edenthemes, this one in foundation/scss/_widgets.scss.

Otherwise it can be added to style.css, and I will then backport to edenthemes.

nursix: https://github.com/nursix/edenthemes

...from there I can propagate it to all default-theme based themes, and also parametrize the background-color so it can be customized easily.

modules/templates/SSF/config.py
```diff
@@ -692,17 +690,17 @@ def task_member_option(data,
         else:
             _id = data["id"]
             if _id == "null":
-                icon_class = "icon-eye-open"
+                icon_class = "fi-eye"
```
nursix: I'd recommend to make use of ICON instead of hardcoding icon classes. This would use an abstract icon name, which can then be configured in config.py.

Example:

Instead of:

    A(I(" ", _class="fi-eye-open"), ...)

you would do:

    A(ICON("watch"), ...)

And then in config.py:

    settings.ui.custom_icons = {"watch": "fi-eye-open"}

Then, if you need to change the icon class (e.g. because you changed the icon set), you adapt this setting rather than the code. E.g. to switch back to font-awesome, you change this into:

    settings.ui.custom_icons = {"watch": "icon-eye-open"}

That makes it easier to maintain and re-use, and you have all icon classes in a central place.

nursix: Note that for ICON definitions that are used in core code (i.e. across templates), a default icon class for the abstract name should be configured in modules/s3/s3widgets.py (ICON class).

In templates config.py, you /could/ hardcode icon classes - but that makes it a bit harder to switch (or upgrade) the icon set and to maintain consistency with core code, of course depending on how many icons the template uses.

Using ICON, the base icon set can be switched with a simple setting:

    settings.ui.icons = "foundation"

...with no need to change any icon classes, except those which you have defined in custom_icons.

nursix: However, this particular icon doesn't seem to be any helpful - it's not very explanatory, nor familiar, and could easily be misinterpreted as action to open ("view") the item, and since it's not repeated in related contexts, it doesn't really have any purpose whatsoever.

modules/templates/SSF/css.cfg
hitesh96db: Font-awesome icons are used by writing 'icon-envelope' which is defined in icons.css, right ? How can I access all the font-awesome icons, apart from adding them in icons.css ?

modules/templates/SSF/config.py
hitesh96db: It doesn't ? I thought this icon was the easiest to understand, an "eye" next to "Watch" :D - like how Github uses it for "Subscribe". I noticed that "CRMT" uses it for open/view but I think it makes sense in this context as well.

modules/templates/SSF/css.cfg
nursix: Is this a serious question?

Replace:

    ../themes/foundation/foundation-icons.css

with:

    bootstrap/font-awesome.css

...and then just change all "fi-..." classes into "icon-..." classes. Or even better follow my recommendation and use ICON with abstract icon names instead of any specific classes.

modules/templates/SSF/config.py
nursix: Does the "eye" explain the "Watch", or the "Watch" explain the "eye"?

As I see it, the "eye" does not add any information to the "Watch" label, nor does it give any hint what "Watching" a task actually means. Nor is it repeated in this context anywhere so that it would help to make a connection - so quite obviously it lacks a purpose. It is just decoration, not adding any value whatsoever. 

That does of course equally apply to the GitHub "Watch" button :P

modules/templates/SSF/css.cfg
hitesh96db: Ah, okay, I was unable to find font-awesome.css, I thought that /only/ needed ones were added to icons.css, and the default/icons.css is the only source.
Yes, ICON is much easier to configure and use.

modules/templates/SSF/config.py
nursix: The CRMT icon for "View" is even worse as it lacks a label.

The open folder icon, the magnifier icon, the arrow-into-document icon - all these are far more familiar for the "view" action than the somewhat exotic "eye" icon.

Anyway - the term "watching" a task is not self-explanatory. What does it mean?

"Subscribing" to task updates seems more self-explanatory (and indeed it gives a hint that this "Watching" is about getting notifications about updates), and so are probably subscribe/unsubscribe icons.

nursix: And the "eye" tells nothing about "receiving notifications on updates" whatsoever - not the slightest hint on email notifications, nor about task updates.

modules/templates/SSF/css.cfg
nursix: default/icons.css has rather historical reasons - as a method to have theme-specific icons before ICON existed.

It could be deprecated, in favour of icon-set specific CSS (easier to maintain) - and if you include font-awesome.css, you can of course omit default/icons.css.

For the default theme, however, I'd like to migrate to FA4 - but for that I first need to get all I-tags replaced with ICONs, which is still a bit of work "ToDo" ;) (meaning: feel free to join the effort)

modules/templates/SSF/config.py
nursix: But please - feel free to keep the "Watch" and the "eye". That won't be any problem for this PR (it's rather my opinion as an end-user).

My point here was actually about abstract ICONs vs. hardcoded icon classes.

hitesh96db: I guess adding a tooltip next to the 'Watch' button would help.

nursix: I don't know - as you can see, the GitHub "Watch" button is actually a drop-down menu that explains the different notification options, so this is certainly how I would do it. The icon is still superfluous, though ;)

hitesh96db:
Updated look: http://test.sunflower.sahanafoundation.org/eden/
Changes made -
* ICON instead of hardcoded classes
* Menu font slightly larger
* Header shadow removed
* Round buttons everywhere
* Usage of large, small and medium classes.
* Fixed Menu Navigation problem

nursix:
I'd say: good to merge.

The initial discussion had been open long enough, and subsequent enhancements can only be submitted once this is merged - so no point to hold off any further.

hitesh96db:
Added minified CSS file.

nursix:
No need to add a minified CSS file - will be minified upon merge by Fran ;)

PR #1057, opened 2015-01-28
Filter Widgets for event_site, event_asset, event_human_resource

gauravmittal1995:


gauravmittal1995:
@nursix Hey, So this code still needs re-factoring and also the widgets for budget, but logic wise, is this the implementation that u were thinking about?

modules/s3db/budget.py
```diff
@@ -1806,5 +1808,69 @@ def budget_rheader(r):
...
+        """
+            Represent a row for a particular budget_allocation
+
+            @param row: budget_allocation Row
+        """
+
+        try:
+            budget = "%s " % (row.budget_budget.name)
+            if row.budget_allocation.start_date or row.budget_allocation.end_date or row.budget_allocation.daily_cost:
+                budget += "(%s, %s, %s)" % (row.budget_allocation.start_date if row.budget_allocation.start_date else "",
```
nursix: Shudder...

Surely you can rewrite this in a more readable manner?

modules/s3db/budget.py
```diff
@@ -1806,5 +1808,69 @@ def budget_rheader(r):
...
+        db = current.db
+        s3db = current.s3db
+
+        table = self.table
+        ptable = s3db.budget_budget
+
+        count = len(values)
+        if count == 1:
+            query = (key == values[0])
+            limitby = (0, 1)
```
nursix: Superfluous variable

modules/s3db/budget.py
```diff
@@ -1320,6 +1320,8 @@ def model(self):
                           s3_comments(),
                           *s3_meta_fields())
 
+        current.s3db.budget_allocation.id.represent = budget_AllocationRepresent(lookup = "budget_allocation")
```
nursix: Not in the model please (lazy tables) - move into the controller where it is needed.

modules/s3db/event.py
```diff
@@ -1736,20 +1763,19 @@ def model(self):
                                                                        ],
                                                              ),
                                         )
+
+            # @ToDo: Add Filter Widgets for Budget
```
nursix: Add a question mark to these ToDos - since the requirements are unclear.

modules/s3db/budget.py
```diff
@@ -1806,5 +1808,69 @@ def budget_rheader(r):
...
+
+        try:
+            budget = "%s " % (row.budget_budget.name)
+            if row.budget_allocation.start_date or row.budget_allocation.end_date or row.budget_allocation.daily_cost:
+                budget += "(%s, %s, %s)" % (row.budget_allocation.start_date if row.budget_allocation.start_date else "",
+                                            row.budget_allocation.end_date if row.budget_allocation.end_date else "",
+                                            row.budget_allocation.daily_cost if row.budget_allocation.daily_cost else "0.0" if row.budget_allocation.daily_cost == 0.0 else ""
+                                            )
+            return budget
+        except:
```
nursix: Unspecific except - must have an exception class specified.

R.I.P. ;) =>
1) RECOVERY: Always only catch exceptions for which you have a correct, logical and effective recovery strategy. If there is no point to carry on after an exception (i.e. if you can not produce a useful result due to the exception) - then don't carry on! Log all exceptions where the recovery strategy (and thus the result) seriously deviates from the intended process.

2) INTEGRITY: Never catch exceptions you're not expecting - otherwise you may be hiding/obscuring errors which can have serious side-effects and potentially cause data loss or security leaks, but at least you'll make it seriously harder to find the underlying bugs.

3) PRECISION: Only ever catch exceptions where you are expecting them - do not have any statements inside the try/except that wouldn't throw that exception. Ideally, there should be only one statement inside the try/except, so that you know *precisely* where the exception occured.

nursix: Or short: don't catch things that should not happen, don't recover if you can't produce a plausible result, and always catch one (and only one) condition at once.

nursix: Remember that an exception will roll back invalid database transactions - unless you catch it. So you must be very very certain that you don't put the database at risk by catching an exception. Thus, you should put *every* try/except in question: is this safe? is this expected? would catching it hide a more serious issue?

In this particular case: what could possibly go wrong? Why could it go wrong? Would this really be an exception - or rather a permanent condition due to a programming mistake that should actually be fixed rather than recovered from?

As far as I can see the only reason for this to fail is if you access fields that you didn't extract before - and if that happens, then you've made a programming error, and hence it would be a permanent condition - the user would only see "Unknown" for each and every row without any indication that this is caused by a programming error.

So:
- this should not throw an exception => don't catch things that should not happen
- you can not produce a plausible result => don't recover if can't produce a plausible result
- you're hiding a programming error preventing it from being detected and hence fixed => don't catch programming errors

So, IMHO there shouldn't be a try/except here.

modules/s3db/budget.py
gauravmittal1995: @nursix Yes, sorry about this. This was just a basic implementation to check the logic .. I still have to refactor it and will make it better.

modules/s3db/budget.py
gauravmittal1995: :+1:

gauravmittal1995:
@nursix  Please Review.

gauravmittal1995:
@nursix Done. Please Review.
Also for the filter for budget, should there be a widget which filters it according to name??

gauravmittal1995:
@nursix Please Review

PR #1058, opened 2015-01-29
Adding 'Expires On' Field in CMS Posts

gauravmittal1995:


controllers/cms.py
```diff
@@ -212,6 +212,15 @@ def prep(r):
         return True
     s3.prep = prep
 
+    def postp(r,output):
+        if r.interactive:
+            if not r.component:
+                # Set the minimum expiry datetime to the same as the start datetime
+                s3.jquery_ready.append(
+'''S3.start_end_date('cms_post_date','cms_post_expiry_datetime')''')
```
nursix: Since this is s3_datetime, it should use the built-in variant start_end_date in S3DateTimeWidget rather than being added manually.

modules/s3db/cms.py
```diff
@@ -208,7 +208,12 @@ def model(self):
                            represent = s3_yes_no_represent,
                            ),
                      s3_datetime(default = "now"),
```
nursix: That is, this one should have an S3DateTimeWidget(set_min="cms_post_expiry_datetime") (this needs to be implemented in s3_datetime, though), and then omit the manual injection of the JavaScript that anyway doesn't take care whether the field is visible or not.

nursix: "needs implementation" means: s3_datetime needs to pass the "set_min"/"set_max" options to the widget ;) the widget (=S3DateTimeWidget) does already support it, though. The start_end_date script is not needed for S3DateTimeWidget.

gauravmittal1995:
@nursix Done. Please Review

gauravmittal1995:
@flavour Please Review

PR #1059, opened 2015-01-29
[Fix]: Form submission of create question popup

raj454raj:


controllers/dc.py
```diff
@@ -136,4 +136,15 @@ def prep(r):
 
     return s3_rest_controller(rheader = s3db.dc_rheader)
 
+# -----------------------------------------------------------------------------
+def template_question():
+    """
+        RESTful CRUD controller for options.s3json lookups
+        - needed for adding recipients
```
nursix: needed for adding what??

nursix:
Docstring doesn't tell the truth, else :+1:

raj454raj:
done ;) :+1:

PR #1060, opened 2015-01-30
Filter Widgets for event_event_type and event_event

gauravmittal1995:


gauravmittal1995:
@nursix Please review

nursix:
No need to ping for review - I get PR's notified anyway.

Scheduled for later - no time right now.

modules/s3db/event.py
```diff
@@ -322,7 +339,7 @@ def model(self):
                   filter_widgets = filter_widgets,
                   list_fields = ["id",
                                  "name",
-                                 "event_type_id$name",
+                                 (T("Type"), "event_type_id$name"),
```
nursix: I'm not sure why we would want to relabel "Event Type" as "Type"?

modules/s3db/event.py
```diff
@@ -265,37 +272,47 @@ def model(self):
...
+                                                  #  get_s3_filter_opts("event_event_type",
+                                                  #                     translate = True)
+                                                  )
+                                  )
+
 
         filter_widgets.extend((S3LocationFilter("event_location.location_id",
                                                 levels = levels,
                                                 label = T("Location"),
+                                                hidden = True,
```
nursix: Why hiding it initially?

modules/s3db/event.py
```diff
@@ -265,37 +272,47 @@ def model(self):
...
                                                 ),
                                # @ToDo: Filter for any event which starts or ends within a date range
                                S3DateFilter("start_date",
-                                            label = None,
+                                            label = T("Date"),
                                             hide_time = True,
-                                            input_labels = {"ge": "From", "le": "To"}
+                                            hidden = True,
```
nursix: Why hiding it initially?

modules/s3db/event.py
```diff
@@ -265,37 +272,47 @@ def model(self):
         # Which levels of Hierarchy are we using?
         levels = current.gis.get_relevant_hierarchy_levels()
 
+        filter_widgets = [S3TextFilter(["name",
```
nursix: Filtering events by name is probably a rare edge-case. I think event type, date and location are the key filters, name is really not so common.

nursix:
I'm not clear why you're making these changes - any requirements specification for that? Or a discussion that I wasn't part of? 

Apart from the name filter for event types, most changes look like a step backwards - reducing clarity and not really fitting with the model and intended workflows.

modules/s3db/event.py
```diff
@@ -322,7 +330,7 @@ def model(self):
                   filter_widgets = filter_widgets,
                   list_fields = ["id",
                                  "name",
-                                 "event_type_id$name",
+                                 (T("Event Type"), "event_type_id$name"),
```
gauravmittal1995: Gives "Name" otherwise. 
http://demo.eden.sahanafoundation.org/eden/event/event

gauravmittal1995:
@nursix Changed Accordingly. Please Review

PR #1061, opened 2015-01-31
[Scenario]: Minor Tweaks

raj454raj:
Add filter widgets for components of scenario.

### DIFF
--- a/controllers/scenario.py
+++ b/controllers/scenario.py
@@ -45,41 +45,8 @@ def scenario():
         return True
     s3.prep = prep
 
-    output = s3_rest_controller(rheader=scenario_rheader)
+    output = s3_rest_controller(rheader = s3db.scenario_rheader)
     return output
-
-# -----------------------------------------------------------------------------
-def scenario_rheader(r, tabs=[]):
-    """ Resource headers for component views """
-
-    rheader = None
-
-    if r.representation == "html":
-
-        if r.name == "scenario":
-            # Scenario Controller
-            tabs = [(T("Scenario Details"), None)]
-            if settings.has_module("hrm"):
-                tabs.append((T("Human Resources"), "human_resource"))
-            if settings.has_module("asset"):
-                tabs.append((T("Assets"), "asset"))
-            tabs.append((T("Organizations"), "organisation"))
-            tabs.append((T("Facilities"), "site"))
-            if settings.has_module("project"):
-                tabs.append((T("Tasks"), "task"))
-            tabs.append((T("Map Profile"), "config"))
-
-            rheader_tabs = s3_rheader_tabs(r, tabs)
-
-            record = r.record
-            if record:
-                rheader = DIV(TABLE(TR(TH("%s: " % T("Name")),
-                                       record.name),
-                                    TR(TH("%s: " % T("Comments")),
-                                       record.comments),
-                                    ), rheader_tabs)
-
-    return rheader
 
 # -----------------------------------------------------------------------------
 def person():
--- a/modules/s3db/scenario.py
+++ b/modules/s3db/scenario.py
@@ -34,6 +34,7 @@ __all__ = ("S3ScenarioModel",
            "S3ScenarioOrganisationModel",
            "S3ScenarioSiteModel",
            "S3ScenarioTaskModel",
+           "scenario_rheader",
            )
 
 from gluon import *
@@ -167,6 +168,18 @@ class S3ScenarioModel(S3Model):
                                       #                                current.messages.AUTOCOMPLETE_HELP))
                                     )
 
+        filter_widgets = [
+            S3TextFilter("name",
+                         label = T("Search")),
+            S3OptionsFilter("incident_type_id",
+                            label = T("Incident Type")),
+            ]
+
+        self.configure(tablename,
+                       filter_widgets = filter_widgets,
+                       list_fields = ["name", "incident_type_id"],
+                       )
+
         # ---------------------------------------------------------------------
         # Pass names back to global scope (s3.*)
         #
@@ -248,7 +261,7 @@ class S3ScenarioAssetModel(S3Model):
         # ---------------------------------------------------------------------
         # Pass names back to global scope (s3.*)
         #
-        return Storage()
+        return dict()
 
 # =============================================================================
 class S3ScenarioHRModel(S3Model):
@@ -266,6 +279,7 @@ class S3ScenarioHRModel(S3Model):
         # Staff/Volunteers
         # @ToDo: Use Positions, not individual HRs (Typed resources?)
         # @ToDo: Search Widget
+
         tablename = "scenario_human_resource"
         self.define_table(tablename,
                           self.scenario_scenario_id(),
@@ -287,7 +301,7 @@ class S3ScenarioHRModel(S3Model):
         # ---------------------------------------------------------------------
         # Pass names back to global scope (s3.*)
         #
-        return Storage()
+        return dict()
 
 # =============================================================================
 class S3ScenarioMapModel(S3Model):
@@ -368,7 +382,7 @@ class S3ScenarioOrganisationModel(S3Model):
         # ---------------------------------------------------------------------
         # Pass names back to global scope (s3.*)
         #
-        return Storage()
+        return dict()
 
 # =============================================================================
 class S3ScenarioSiteModel(S3Model):
@@ -407,7 +421,7 @@ class S3ScenarioSiteModel(S3Model):
         # ---------------------------------------------------------------------
         # Pass names back to global scope (s3.*)
         #
-        return Storage()
+        return dict()
 
 # =============================================================================
 class S3ScenarioTaskModel(S3Model):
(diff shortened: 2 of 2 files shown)


modules/s3db/scenario.py
```diff
@@ -404,10 +432,17 @@ def model(self):
             msg_record_deleted = T("Facility removed"),
             msg_list_empty = T("No facilities currently registered in this scenario"))
 
+        self.configure(tablename,
+                       filter_widgets = [S3TextFilter("site_id$name",
+                                                      label = T("Search")),
+                                         ],
+                       list_fields = ["site_id"],
```
raj454raj: Before: Without this  - IndexError is raised

nursix: Where?

modules/s3db/scenario.py
```diff
@@ -284,10 +302,16 @@ def model(self):
             msg_record_deleted = T("Human Resource removed"),
             msg_list_empty = T("No Human Resources currently registered in this scenario"))
 
+        self.configure(tablename,
+                       filter_widgets = [S3TextFilter("human_resource_id$person_id$first_name",
```
nursix: Ahm...just first_name? You're kidding, right?

controllers/scenario.py
```diff
@@ -40,48 +40,17 @@ def prep(r):
...
-                field = db.scenario_site
+                stable = s3db.scenario_site
+                field = stable.site_id
                 field.readable = field.writable = True
         return True
     s3.prep = prep
 
-    output = s3_rest_controller(rheader=scenario_rheader)
+    output = s3_rest_controller(rheader = s3db.scenario_rheader,
+                                hide_filter = False)
```
nursix: Do we really want filters on all tabs? I'm not so sure about this, especially since most of them are just text filters which actually works with the datatable filter already.

nursix:
Apart from the HR first_name-only filter, this looks ok. But I'm unclear about the requirements for filter widgets on the component tabs here - they don't seem to add much value (yet processing time)? Any specifications or discussion that triggered these changes where I can read up on?

nursix:
Don't get me wrong - I'm not saying this is wrong, just putting it in question. 

I'm generally reluctant to green-light additional processing in models simply because that increases the effort during model loading. And since I haven't yet heard of any requirements to have dedicated filter forms on the component tabs of scenarios, I need to ask where this is coming from.

modules/s3db/scenario.py
raj454raj: Ajax call -> redirects to a ticket

raj454raj:
> Any specifications or discussion that triggered these changes where I can read up on

Not really, just asked on IRC about what these todos were targeted to. Michael said that it should be for the filter widgets and actioning them woudn't take time. 
Fran seemed to be busy that time so he was not able to reply. So your call - required or not. @nursix  

Shall I remove hide_filter = False and keep the filters as it is so when its required this can be added?

modules/s3db/scenario.py
raj454raj: Sorry really forgot to add. Adding in a sec.

nursix:
A possible strategy to determine usefulness of filter widgets is to try to identify the questions they may answer:
=> If you can't find a meaningful question, then the filter widget isn't useful.
=> If the possible answer is of the yes/no type, then the filter widget is probably at the wrong end of an entity relationship

For example: the text filter on the "Asset" tab of a scenario would answer the following question: "Which assets with the number XYZ have been planned for this scenario?" (obviously nonsense, because the best answer to the question is an asset number - but you already have it, so no new information here).

So, rephrasing this as: "Has the asset with the number XYZ been planned for this scenario?"

Now, this gives a yes/no answer - so we may be looking at the wrong end of the link, i.e. we should probably better look at the "Scenario" tab of the asset (instead of the "Asset" tab of the scenario), and ask:

"Which scenarios has this asset been planned for?"

Now, this *is* a useful question - but you don't need a filter to answer it, the mere list is quite enough. However, the question can have qualifiers:

"Which scenarios of event type X has this asset been planned for?"

And to answer that, you will need an event-type filter on the "Scenario" tab of the asset! There you have it!

...so that is largely my strategy to identify useful filters where I don't have detailed requirements specifications. Doesn't necessarily mean that each filter identified this way is indeed needed by the users - but at least I can exclude those filters which really don't answer anything.

Wanna try this?

modules/s3db/scenario.py
nursix: Ok - my actual question was: Where in the code is the exception raised?

Reason: if an IndexError is raised, then obviously the function *there* is wrong and doesn't check whether the element it accesses exists in the list.

...which is wrong and should be fixed (instead of trying to avoid the condition by defining the element here)

nursix: Or in short: Do not work around bugs, fix them!

raj454raj:
@nursix  -
I'll look into this for sure. Do you want a scenario component for asset? Is it required really?
> And to answer that, you will need an event-type filter on the "Scenario" tab of the asset! There you have it!

So from this, what I understood is - 
Make scenario a component tab of asset. Add a filter to this tab which filters according to event_incident_type?

nursix:
Ahm - this was actually just an example :$

...not trying to suggest any particular requirements, but rather to illustrate a possible strategy to identify useful ones, just using the given context to clarify.

The general use-case of the "scenario" model is to develop response plans which can then be rolled out in case of a disaster - automatically activating resource deployments including alerting, availability checking, choosing alternatives, and reporting status/progress.

It should perhaps also be able to report on potential costs (for budget planning, fund raising), and allow checking of the status of response plans - and their feasibility (...resource availability, training status, maintenance status etc etc...).

Ability to pull in hazard and risk information, as well as demographics data, while planning responses is another useful feature - and at the highest end of the requirements are probably simulations of scenarios.

If you start with this high-level description, you can perhaps find additional features that would be required or useful, and then the answer to your question would come just naturally.

Ideally, you would also pull in (potential) end-users to the discussion - and listen to them.

In general, development shouldn't start with available functionality - but with the end-user needs. Not solutions looking for problems - but the other way around. So, start with specifying the problem, and then work up possible solutions.

If you come to me with a problem statement, then we can discuss possible solutions. If you come only with a solution, then it's unlikely I can help you much.

raj454raj:
> Ahm - this was actually just an example :$
...not trying to suggest any particular requirements, but rather to illustrate a possible strategy to identify useful ones, just using the given context to clarify.

Okay, what should be the status of the PR(sorry to be too straight) !!!

> If you come to me with a problem statement, then we can discuss possible solutions. If you come only with a solution, then it's unlikely I can help you much.

Will take care of this in future requests.

nursix:
Umh - I kinda said that already?

The IndexError must be fixed where it occurs, not be worked around.

Adding filters for the main scenario resource, changing the Storages into dicts, and moving the rheader into the model - these things are fine and can/should be merged.

The rest of the changes (=the filters for the component tabs plus the hide_filter=False) are questionable - not technically wrong, but unclear requirements-wise. I would take them out of this PR, and re-approach once there are clear requirements for that.

raj454raj:
Okay thanks :+1:

raj454raj:
@nursix : Done. :+1: 
Here's what the problem was - 
The site_id was made non-readable and non-writable-
https://github.com/flavour/eden/blob/master/modules/s3db/org.py#L2670
And it is again made readable and writable in the controller here -
https://github.com/flavour/eden/blob/master/controllers/scenario.py#L44
So now in this function - 
https://github.com/flavour/eden/blob/master/modules/s3/s3resource.py#L3872
The list_fields that we get is just ["id"] which is expected to be ["id", "site_id"], so if we explicitly pass site_id then it works. 
OR
The better one would be make this field readable before the list_fields() function is called ;).

nursix:
Doesn't make sense to me, sorry.
list_fields is not called before the prep, and even if we get just ["id"], it should still not raise an IndexError - and you have not yet told me where the IndexError is raised. Where is list fields expected to be ["id", "site_id"]? That can not just be expected!

nursix:
This is still working around the bug, not fixing it.

raj454raj:
@nursix : 
At this line https://github.com/flavour/eden/blob/master/modules/s3/s3resource.py#L3882 list_fields should have ["id", "site_id"]. 
The IndexError is raised here:
https://github.com/flavour/eden/blob/master/modules/s3/s3resource.py#L3712

nursix:
I can not see that  https://github.com/flavour/eden/blob/master/modules/s3/s3resource.py#L3882 anyhow expects ["id", "site_id"] - what makes you think that? There is nothing there that expects any particular list fields.

And https://github.com/flavour/eden/blob/master/modules/s3/s3resource.py#L3712 in fact catches the wrong Exception type - lists raise an IndexError not a KeyError. And that is basically all to it.

But what you keep overlooking:
https://github.com/flavour/eden/blob/master/controllers/scenario.py#L44

"s3db.scenario_site" is not a Field! Setting is readable/writable has no effect.
It should be "s3db.scenario_site.site_id" - and with that, it all works perfectly well.

I pushed a fix for both.

nursix:
Now, there is one more mistake here. I'll fix that too.

raj454raj:
Thats what I was thinking about - your changes didnt work for me

raj454raj:
Okay so I'll remove this fix from this commit

nursix:
Like this:
https://github.com/nursix/eden/commit/3d531bbaf86cd7b3fcd7527bebfce542b4579fed

The IndexError was caused by an interactive/aadata list_fields discrepancy: site_id was only set readable for interactive requests, but not for the aadata Ajax-sorting - so the sorting index would hit a nonexistent field.

However, that should not crash but get caught in datatable_filter (because it's a URL error, and thus "user data" in the wider sense, which should never crash) - but that didn't catch the right exception type.

nursix:
Well, it works for me with these changes - if they don't work for you, then maybe you have introduced another error somewhere else?

nursix:
I don't see anything wrong - it's good to merge now.

raj454raj:
> Well, it works for me with these changes

Yeah even for me now, I tried your previous commit.
> but that didn't catch the right exception type.

I tried this also but thought that this might break other cases. 
But thanks :) :+1:

PR #1062, opened 2015-02-01
Adding S3.start_end_date in S3DateWidget

gauravmittal1995:
For s3_date() , adding a start_field enables the script and for s3_datetime() we can use the S3DateTimeWidget(set_min ='....') to enable it.


modules/s3db/hrm.py
```diff
@@ -2539,6 +2539,7 @@ def model(self):
                                      ),
                      s3_datetime("start_date",
                                  label = T("Start Date"),
+                                 widget = S3DateTimeWidget(set_min = "hrm_training_event_end_date"),
```
nursix: I think that set_min can (should) actually be passed in to s3_datetime rather than the whole widget configuration.

modules/s3/s3widgets.py
```diff
@@ -1370,6 +1370,17 @@ def __call__(self, field, value, **attributes):
         if script not in jquery_ready: # Prevents loading twice when form has errors
             jquery_ready.append(script)
 
+        if start_field:
```
nursix: Hopefully this doesn't interfere with end_date_interval?

modules/s3db/event.py
```diff
@@ -208,17 +208,14 @@ def model(self):
...
-                                 label = T("Start Date"),
-                                 represent = "date",
-                                 widget = "date",
-                                 ),
-                     s3_datetime("end_date",
-                                 label = T("End Date"),
-                                 represent = "date",
-                                 widget = "date",
-                                 ),
+                     s3_date("start_date",
```
nursix: Is there a reason to change this into date?

gauravmittal1995: Actually it was to pass the start_field, but now that set_min should be passed to s3_datetime, will keep it as datetime.

modules/s3/s3widgets.py
gauravmittal1995: No, not that i could see. I tried it on a end_date_interval start_date, and was working as expected

modules/s3db/hrm.py
gauravmittal1995: Ok ... will do that

gauravmittal1995:
Done

gauravmittal1995:
@nursix Done

gauravmittal1995:
@nursix Done. Please Review

PR #1063, opened 2015-02-05
Revision to CERT Template

michaelhowden:
* default System Name Sahana Eden -> Sahana
* setting.hrm.multiple_orgs to indicate if there are staff/vols in 1 or >1 organisations
* Standardize Trainings/Certificate labels
* Add certificate filter
* Update  homepage
* Update Training import
* Add Certificate import
* Remove CERT Theme (use default theme)


modules/s3cfg.py
```diff
@@ -612,12 +612,12 @@ def get_system_name(self):
         """
             System Name - for the UI & Messaging
         """
-        return self.base.get("system_name", current.T("Sahana Eden Humanitarian Management Platform"))
+        return self.base.get("system_name", current.T("Sahana Disaster Management Platform"))
```
nursix: I would have thought that this requires a community discussion, at least. Dropping "Eden" goes a bit beyond a simple code change.

nursix: And surely I would drop "Sahana" instead of "Eden" ;)

Nah - just kidding... :P

michaelhowden: +1 - Will revert and take it to the community

flavour: Within the CERT template it can be whatever the users want...it's the default template that needs more consideration

modules/s3db/hrm.py
```diff
@@ -8440,23 +8440,24 @@ def hrm_human_resource_filters(resource_type=None,
...
+        else:
+            append_filter(S3OptionsFilter("organisation_id",
+                                          filter = True,
+                                          header = "",
+                                          #hidden = True,
+                                          ))
+    
+        # Location filter (always)
+        append_filter(S3LocationFilter("location_id",
+                                       label = T("Location"),
```
flavour: Location filter behind multiple_orgs? Slip or deliberate?
Although multiple locations /might/ mean multiple branches, I'm not sure that's necessarily the case...

modules/templates/CERT/views/index.html
```diff
@@ -1,12 +1,27 @@
 {{extend "layout.html"}}
+<link href="/{{=appname}}/static/themes/CERT/homepage.css" rel="stylesheet" type="text/css" media="screen" charset="utf-8" />
```
flavour: All CSS should be in HEAD for faster page loads, so should be injected in the controller into s3.stylesheets

modules/templates/CERT/config.py
```diff
@@ -24,7 +25,7 @@ def config(settings):
     settings.base.prepopulate = ("CERT", "default/users")
 
     # Theme
-    settings.base.theme = "CERT"
+    #settings.base.theme = "CERT"
```
flavour: I confess to feeling a pang at losing this prize-winning theme, but things are moving on & it certainly should at least be rebased on the current default theme.

modules/s3db/hrm.py
michaelhowden: Slip :)

modules/templates/CERT/config.py
michaelhowden: RIP CERT Theme
2011-2015
Winner of RHoK. Promoter of shiny colourful Icons. A standard to follow for better managing Community Emergency Response Team Volunteers.

michaelhowden: Was there anything that I culled unnecessarily?
The "Deploy" link was broken - this could be fixed, but would require a bit more work on the event module (some of which I've done - but needs to be merged from my DERA branch). I've added a @ToDo for this - but it's not needed in the current use case for this template.
Removing the theme itself is part of my push to standardize the design and remove duplication. I had a play around with leaving the theme in, but this forced having a dedicated (duplicate) layout.py file.
Sounds like you're OK with these changes however though - just... reminiscent? :)

michaelhowden:
I've pushed a new commit with changes addressing all the raised issues

modules/templates/CERT/config.py
nursix: Your push to standardize the design? Never heard of that.

Can you explain?

michaelhowden: Use and improve default theme where possible.

nursix: Our actual direction is to provide a larger degree of freedom for GUI design, including framework and tools to implement different designs.

The default theme is intended for RAD (with the D meaning both development and deployment), and therefore flexible, case-neutral and minimalistic. But full-scale development projects may not need to be that, and can thus very well use a different, case-specific design - e.g. to blend in, reduce training efforts, support specific workflows.

A push to standardize the design and use default wherever possible is thus news to me - and I wonder whether that shouldn't be discussed with the wider community first. Looks like an unilateral decision of yours.

nursix: And I would be rather reluctant with sudden design changes (regardless whether they are real UX improvements or just fashion), because such changes always mean a learning effort on the user side, and often break existing documentation.

Thus, the default theme is not the ideal place for cutting-edge GUI development - which though seems to be what you are actually looking for. The default theme is rather suitable as final destination for the stable and mature features, at a slow pace and with infrequent fluctuation, and after designs have been successfully field-tested, fully framework-integrated and documented.

I would therefore rather not push for a "standardization" of the default theme, but where possible, realize new design ideas in custom themes first, field-test them, and then cherry-pick, generalize, polish and integrate features back into the default design - which then applies them to the various RAD templates. That way, we don't put stability at risk, and yet can move GUI design forward.

That concept was actually successful - so why should we change it?

nursix: My push would be to rebase the CERT theme on the current default theme (as it was based on the default theme before), while retaining its specific features, maybe even improving them in view of responsiveness - rather than removing it.

nursix: ...and the homepage menu could be improved to use S3Navigation ;)

nursix: ...yet retaining the original icons...

michaelhowden: Why does CERT need it's own theme? Why can't it just use default theme? We don't need a new theme if there's just a couple of custom pages (do we?)
I'm hoping to utilize it more for more deployments - surely that's a good, RAD thing? I'm sure in the process minor incremental iterations (pixel pushing) can be made to the default them to improve it everywhere (there's none of those in this commit)
Homepage menus could certainly be improved to use S3Navigation.
Using font-awesome icons is much more extendable - there's more to choose from as we add more buttons.

nursix: I think I've answered most of that already.

Using the default theme for more deployments is not the same as RAD, no - these are totally orthogonal things.

RAD means to be able to quickly knock up an application from pre-built components, and the particular need are emergencies where there are neither time nor resources to develop new components from scratch, let alone to build a new theme - and yet very specific requirements.

However, that does not define any need or desire to maximize quantity - if it is necessary to use RAD components in order to save time and resources, then that is something we support.

But if there is enough time and resources for development of new features (and designs), then that is certainly the preferrable way - especially because it can eventually feed back into the RAD components and thus make new technology available for less well-resourced efforts.

That's what I suggested before: high end development projects with plenty of budget and/or resources develop cutting edge technology and designs, which then get field-tested and generalized through these projects - and eventually made ready and available for RAD.

RAD components need to be mature, stable, thoroughly tested and documented - and should not necessarily be the focus of the development. But RAD projects do neither have the resources nor the incentive to actually develop those components - and *that* is actually the point. 

So there should actually not be /as many as possible/ RAD deployments, but wherever possible, sufficiently resourced development projects providing new components for RAD deployments - and RAD where necessary.

michaelhowden:
@flavour: What was the hold up on merging this? I think that the static/formats/s3csv/hrm/person.xsl changes will be useful for SahanaCamp Turkey
Travis was failing, although the error was pretty cryptic.  minimal investigation revealed a pretty major bug (now fixed):
-        @ToDo: Add event/human_resource - but this requires extending event_human_resource to link to event.
+        # @ToDo: Add event/human_resource - but this requires extending event_human_resource to link to event.
I think that the RAD approach is still valuable when there is time and resources - if it delivers the right solution.  Regardless - this deployment does not have much in the way of time or resources to put into it, so RAD is preferable.

nursix:
"RAD is still valuable when there is time and resources - if it delivers the right solution"

...what a completely hollow statement! Delivering the right solution is always valuable, that's not a value added specifically through RAD.

If you have sufficient time and resources, then RAD components can be used to produce early prototypes - but this strategy is called "rapid prototyping", not RAD. 

This can help to facilitate continuous end-user participation in agile development, which tends to deliver more targeted (and hence more effective) solutions, and can dramatically reduce the defect rate (due to early continuous UAT). 

However - it does eventually cost *more* effort and not less (which is why you need time and resources), and such solutions are typically much less re-usable - especially they do rarely deliver or improve RAD components (they rather tend to tweak RAD-components into non-RAD).

The key thing for RAD are re-usable, generic, configurable, robust and stable components - as well as excellent conceptual integrity (all components following standard concepts and common principles) and consistency (especially in view of terminology), often referred to as the "no surprises" principle. The most important aspect of RAD is that developers do not need to in-depth study those documents, let alone to develop them.

Such components can not be developed through RAD itself - nor does rapid prototyping help (rather the contrary). They do need dedicated development projects with sufficient time and resources both for defining, implementing and testing as well as generalization and conceptual integration of these components.

And that is why RAD - regardless how valuable it is - is not possible without strategic non-RAD projects providing both requirements and resources to develop RAD components and framework.

The CERT template has definitely had enough *time* to be non-RAD even with just volunteer resources. It was rather lacking requirements development (mainly because it didn't get used much), which is why it remained at the prototype stage.

I still don't see the urgency - it seems that you're rather hijacking this template for a different use-case / target audience (aren't you?), which may be ok if declared that way. Or is there any CERT team suddenly interested in developing this further?

nursix:
The key aspect of RAD is that full-blown /production/ applications can be developed with almost no programming effort at all - with all forms, presentations and tools introspecting and automatically adapting to the data model and workflow descriptions.

However, it also involves a certain robustness and flexibility for subsequent remodelling, safeguarding data integrity and minimizing dependencies between code and database, basically allowing to re-use the same database with different applications/configurations and vice versa (which implies a 100% separation of database and application logic).

All of this is useful - sometimes, but not always and every time. Quantity is a result of RAD, but doesn't improve it. It creates an incentive, but doesn't produce a framework.

None of this is possible without dedicated framework/component development - which really requires that projects which have sufficient time and resources feed into it by developing new components, field-testing, generalizing and integrating them.

michaelhowden:
There is a CERT team that is interested in using Sahana - which is why I've made these changes.

PR #1065, opened 2015-02-09
Minor fixes & Tweaks

raj454raj:


static/styles/ui/fgtimepicker.css
```diff
@@ -23,7 +23,7 @@
 /* span for disabled cells */
 .ui-timepicker-table td span {
 	display: block;
-    padding: 0.2em 0.3em 0.2em 0.5em;
+    padding: 0.2em 1.5em 0.2em 0.5em;
```
raj454raj: screenshot from 2015-01-30 00 22 15

raj454raj: Now
screenshot from 2015-01-30 00 22 43

modules/s3db/inv.py
```diff
@@ -4099,7 +4099,7 @@ def model(self):
                            requires = IS_EMPTY_OR(IS_IN_SET(inv_item_status_opts)),
                            represent = lambda opt: \
                                        inv_item_status_opts.get(opt, UNKNOWN_OPT),
-                           default = 0,),
+                           default = 0),
```
nursix: Actually: trailing comma, and closing bracket in the next line ;)

controllers/survey.py
```diff
@@ -48,7 +48,7 @@ def create():
         - provides a simpler URL to access from mobile devices...
     """
 
-    redirect(URL(f="newAssessment.iframe",
```
nursix: I think that .iframe was intentional here.

nursix:
Good work - just two little things, then it should be good to merge.

controllers/survey.py
raj454raj: Okay I'll keep it, but I don't find any popup calling create() and this url-
http://127.0.0.1:8000/eden/survey/newAssessment.iframe?viewing=survey_series.1 
is empty

raj454raj:
Done. :+1:

nursix:
Yup - this can be merged, unless Travis objects ;)

controllers/survey.py
nursix: I know - but see what the docstring says: obviously there is a requirement for this, so rather than removing the iframe-extension, we should make it work ;)

nursix: Well...after investigating and understanding that requirement, of course.

raj454raj:
Oops travis. Wait a sec, what adding a newline fails the tests?
First it failed for mysql then now for the last one?
@nursix

nursix:
Don't worry, it's a temporary Travis-internal issue.

raj454raj:
@nursix: Added some minors. Please review.

controllers/survey.py
```diff
@@ -1081,7 +1083,7 @@ def import_xls(uploadFile):
         for row in xrange(1, sheetM.nrows):
             header += ',"%s"' % sheetM.cell_value(row, 0)
             code = sheetM.cell_value(row, 0)
-            qstn = s3.survey_getQuestionFromCode(code, series_id)
+            qstn = s3db.survey_getQuestionFromCode(code, series_id)
```
raj454raj: Without this importing completed assessments in xls format and populating to the database wont work.

nursix: Is this a question or just a note to yourself? I mean - that is well obvious ;)

raj454raj: Its a note - because I have referred this PR in the document I will be sending :)

nursix:
I'm surprised that this hasn't yet been merged (maybe due to the previous Travis error) - but I don't see much that speaks against merging.

For the form.insert(0, buttons) I need an explanation - obviously they are passed into the view as "subtitle", but maybe the view doesn't apply it correctly. Either of the variants has to go away, shouldn't do both - and honestly, inserting into FORM isn't the cleanest way to do this.

raj454raj:
@nursix : ^ Will find a better way to do this. Basically the button was not added to the form. Any help on how this "subtitle" option is handled?

nursix:
Well, we used to have a {{=subtitle}} in view/_create.html - but that was loooong ago. In the current version, we don't have such things anymore.

Injecting the buttons into the form is risky, because you don't actually know what the exact HTML structure of the form is - nor does this check whether form actually is an HTML structure (and not "").

The better (=safer) method is to wrap the form:

output["form"] = TAG[""](buttons, form)

...which equally prepends the buttons, but doesn't touch the form in any way.

nursix:
Note that this ^ does not add an HTML node - if you want that, make it a DIV instead of a TAG[""].

raj454raj:
@nursix : Again travis failed. Mergeable?

coder006:
@raj454raj Why close this PR?

raj454raj:
@coder006 : Sending it on Dominic's fork.

coder006:
> @coder006 : Sending it on Dominic's fork.

cool.

PR #1066, opened 2015-02-09
Migrate to "ICON"

hitesh96db:


modules/s3db/deploy.py
```diff
@@ -2095,7 +2095,7 @@ def deploy_response_select_mission(r, **attr):
             for a in atts:
                 url = URL(c="default", f="download",
                           args=a.file)
-                attachments.append(TR(TD(A(I(" ", _class="icon icon-paperclip"),
+                attachments.append(TR(TD(A(ICON("paper-clip"),
```
nursix: "attachment" not "paper-clip"

modules/s3db/deploy.py
```diff
@@ -2520,15 +2520,15 @@ def render_body(self, list_id, item_id, resource, rfields, record):
...
+                        doc_item = A(ICON("file"),
+                                     " ",
+                                     doc_name,
+                                     _href=doc_url,
+                                     )
                     docs.append(doc_item)
                     docs.append(", ")
                 if bootstrap:
-                    docs = DIV(A(I(_class="icon-paper-clip"),
+                    docs = DIV(A(ICON("paper-clip"),
```
nursix: "attachment" not "paper-clip"

modules/s3db/deploy.py
```diff
@@ -2677,7 +2677,7 @@ def render_body(self, list_id, item_id, resource, rfields, record):
                           vars={"refresh": list_id,
                                 "record": record_id
                                })
-                edit = A(I(" ", _class="icon icon-paperclip"),
+                edit = A(ICON("edit"),
```
nursix: Hmm, that may not be the idea here - it conflicts with the other "edit" icon on the same card. Should be a "paper-clip" in this case, indeed.

modules/s3db/deploy.py
```diff
@@ -2799,19 +2799,19 @@ def render_toolbox(self, list_id, resource, record):
...
            has_permission("update", table, record_id=record_id):
-            btn = A(I(" ", _class="icon icon-edit"),
+            btn = A(ICON("edit"),
                     _href=update_url,
                     _class="s3_modal",
                     _title=crud_string(tablename, "title_update"))
             toolbox.append(btn)
         elif open_url:
-            btn = A(I(" ", _class="icon icon-file-alt"),
+            btn = A(ICON("file-alt"),
```
nursix: Questionable. "read" would be better.

modules/s3db/deploy.py
```diff
@@ -2799,19 +2799,19 @@ def render_toolbox(self, list_id, resource, record):
...
                     _href=open_url,
                     _title=crud_string(tablename, "title_display"))
             toolbox.append(btn)
 
         if has_permission("delete", table, record_id=record_id):
-            btn = A(I(" ", _class="icon icon-trash"),
+            btn = A(ICON("trash"),
```
nursix: "delete" not "trash"

modules/s3db/doc.py
```diff
@@ -565,7 +565,7 @@ def doc_document_list_layout(list_id, item_id, resource, rfields, record):
             doc_name = current.messages["NONE"]
         doc_url = URL(c="default", f="download",
                       args=[file])
-        body = P(I(_class="icon-paperclip"),
+        body = P(ICON("paper-clip"),
```
nursix: "attachment" not "paper-clip" (I won't repeat this for the other cases anymore)

modules/s3db/doc.py
```diff
@@ -602,7 +602,7 @@ def doc_document_list_layout(list_id, item_id, resource, rfields, record):
     else:
         edit_btn = ""
     if permit("delete", table, record_id=record_id):
-        delete_btn = A(I(" ", _class="icon icon-trash"),
+        delete_btn = A(ICON("trash"),
```
nursix: "delete" not "trash"

modules/s3db/doc.py
```diff
@@ -613,7 +613,7 @@ def doc_document_list_layout(list_id, item_id, resource, rfields, record):
                    )
 
     # Render the item
-    item = DIV(DIV(I(_class="icon"),
+    item = DIV(DIV(ICON("_base"),
```
nursix: "icon" for placeholders.

modules/s3db/event.py
```diff
@@ -2451,7 +2451,7 @@ def event_incident_list_layout(list_id, item_id, resource, rfields, record,
     else:
         edit_btn = ""
     if permit("delete", table, record_id=record_id):
-        delete_btn = A(I(" ", _class="icon icon-trash"),
+        delete_btn = A(ICON("trash"),
```
nursix: "delete"

modules/s3db/event.py
```diff
@@ -2463,7 +2463,7 @@ def event_incident_list_layout(list_id, item_id, resource, rfields, record,
                    )
 
     # Render the item
-    item = DIV(DIV(I(_class="icon icon-%s" % icon),
+    item = DIV(DIV(ICON(icon),
```
nursix: For this you must check whether all cases are satisfied - i.e. where is it called, what are the possible values for icon, are all these icons defined?

modules/s3db/event.py
```diff
@@ -2561,7 +2561,7 @@ def event_resource_list_layout(list_id, item_id, resource, rfields, record):
     else:
         edit_btn = ""
     if permit("delete", table, record_id=record_id):
-        delete_btn = A(I(" ", _class="icon icon-trash"),
+        delete_btn = A(ICON("trash"),
```
nursix: "delete"

modules/s3db/hrm.py
```diff
@@ -7948,7 +7948,7 @@ def hrm_competency_list_layout(list_id, item_id, resource, rfields, record):
     else:
         edit_btn = ""
     if permit("delete", table, record_id=record_id):
-        delete_btn = A(I(" ", _class="icon icon-trash"),
+        delete_btn = A(ICON("trash"),
```
nursix: "delete"

modules/s3db/hrm.py
```diff
@@ -7959,7 +7959,7 @@ def hrm_competency_list_layout(list_id, item_id, resource, rfields, record):
                    )
 
     # Render the item
-    item = DIV(DIV(I(_class="icon"),
+    item = DIV(DIV(ICON("_base"),
```
nursix: just "icon" for placeholders

modules/s3db/hrm.py
```diff
@@ -8038,7 +8038,7 @@ def hrm_credential_list_layout(list_id, item_id, resource, rfields, record):
     else:
         edit_btn = ""
     if permit("delete", table, record_id=record_id):
-        delete_btn = A(I(" ", _class="icon icon-trash"),
+        delete_btn = A(ICON("trash"),
```
nursix: "delete" (I stop repeating this here)

modules/s3db/hrm.py
```diff
@@ -8049,7 +8049,7 @@ def hrm_credential_list_layout(list_id, item_id, resource, rfields, record):
                    )
 
     # Render the item
-    item = DIV(DIV(I(_class="icon"),
+    item = DIV(DIV(ICON("_base"),
```
nursix: just "icon" for placeholders (I'll not repeat this again)

modules/s3db/hrm.py
```diff
@@ -8085,7 +8085,7 @@ def hrm_experience_list_layout(list_id, item_id, resource, rfields, record):
 
     raw = record._row
 
-    card_line = lambda icon, item: P(I(" ", _class=icon),
+    card_line = lambda icon, item: P(ICON(icon),
```
nursix: Must check all cases

modules/s3db/org.py
```diff
@@ -6479,12 +6479,12 @@ def org_organisation_list_layout(list_id, item_id, resource, rfields, record):
...
                        _class="card-header-select",
                        ),
-                   DIV(P(I(_class="icon icon-phone"),
+                   DIV(P(ICON("phone"),
                          " ",
                          phone,
                          _class="card_1_line",
                          ),
-                       P(I(_class="icon icon-map"),
+                       P(ICON("map-marker"),
```
nursix: Careful - I con't think map-marker is correct here. It should probably be "link" instead.

modules/s3/s3widgets.py
```diff
@@ -7626,7 +7626,8 @@ class ICON(I):
             "list": "icon-list",
             "mail": "icon-envelope-alt",
             "map-marker": "icon-map-marker",
-            "offer": "icon-truck",
+            "offer": "icon-certificate",
```
nursix: Hmm, this used to be icon-truck, and sudden icon-changes aren't really good. But it's ambiguous so we may indeed need to look for a better icon. However, "certificate" is already used in a different context.

nursix:
Good start - but needs work:

"trash" should generally be "delete"
"paper-clip" should be "attachment" for the attachment (doc) context - whilst for appraisals in "deploy" it should indeed be paper-clip for now.

Placeholder I-tags should be rendered as ICON("icon") - that is because "_base" will change, but the CSS-selector for the placeholder will always be .icon.

Where you replace variable I-tags, you must check all cases for whether all possible values are actually defined. This can be a tedious job, and is indeed the most difficult part of it (must find and analyse all callers)

Changing a default-icon requires careful consideration - since users may be used to certain icons, and re-training users isn't exactly something you can do on-the-fly. In the case of "offer", however, we do indeed need a different icon because the truck is already used for another item class. The certificate icon is too, though :/ so we need something else.

Else - many thanks for the help :) :)

nursix:
P.S. Note that not all ICONs are currently used with font-awesome, nor do all themes use font-awesome. Some use a different icon set, and some even use handcrafted background-images.

For that reason, unfortunately, you must test across templates and verify before/after for every case.

nursix:
...on the bright side, though, almost every other icon set currently in use in Eden also uses icon-* classes, so these are sometimes just overrides (which though means you can't just change the class).

The only exception is Foundation icons which looks for fi-*

nursix:
Ah - and another P.S.: Please try to use abstract icon names rather than descriptive, e.g.:

"paper-clip" => "attachment"

That way, when you configure a new icon set that uses a completely different symbology you can pick alternatives by semantic meaning rather than by visual appearance - which obviously makes more sense.

E.g. if there is no "paper-clip" icon, you shouldn't look for something that looks similar to a paper-clip, but rather for something that can symbolize an attachment.

hitesh96db:
:+1: 
Should I continue with these(which are in s3db) ?

nursix:
Yes, please complete this PR - but make sure you're testing everything (especially /deploy/mission and deploy/member with the IFRC theme)

modules/s3/s3widgets.py
hitesh96db: Maybe a "tag" ?

modules/s3db/event.py
hitesh96db: What could be a possible icon for "incident". I can't think of an apt one.
...and none of the places calling this configure the icon so far, so it'll always be the default one for now.

hitesh96db: @nursix ^

PR #1067, opened 2015-02-09
s3_richtext_widget for About and Help

raj454raj:
s3_richtext_widget instead of plain textbox for update posts of About and Help

### DIFF
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -955,13 +955,18 @@ def about():
         item = db(query).select(table.id,
                                 table.body,
                                 limitby=(0, 1)).first()
+
+        get_vars = {"module": module,
+                    "resource": resource}
+
         if item:
             if ADMIN:
                 item = DIV(XML(item.body),
                            BR(),
                            A(T("Edit"),
                              _href=URL(c="cms", f="post",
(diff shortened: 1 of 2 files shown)


controllers/default.py
```diff
@@ -955,13 +955,18 @@ def about():
         item = db(query).select(table.id,
                                 table.body,
                                 limitby=(0, 1)).first()
+
+        get_vars = {"module": module,
+                    "resource": resource}
```
nursix: Uhm - "resource" correct in this context?

In this case you must extend: https://github.com/flavour/eden/blob/master/controllers/cms.py#L157

...so you get a proper default name, and the title readable/writable etc...

raj454raj: :+1:

raj454raj: @nursix : How is this decided for About and help?
https://github.com/flavour/eden/blob/master/controllers/cms.py#L166

nursix: Hmm, find me a website with a reply-option on the About or Help pages...

My gut feeling says these pages are in class with Contact - so False should be just right.

raj454raj:
@nursix : Done. :+1:

PR #1070, opened 2015-02-16
Add setup url config

gnarula:


nursix:
Config setting without use-case? I would re-submit this together with the code that uses it.

gnarula:
This was discussed in the Configuration Editor thread in the mailing list.

Added a link to the coapp in the admin menu.

nursix:
I'm aware of what has been discussed - but adding a setting without also adding the use-case for it is obviously a bit pointless or at least premature.

Better now - but you should of course use the getter to retrieve the setting, not access the dict directly.

gnarula:
Sorry for that. Done now :)

PR #1071, opened 2015-02-24
Threaded attachments on post

yashpalsaini:
add attachment table for cms-post

### DIFF
--- a/modules/s3db/cms.py
+++ b/modules/s3db/cms.py
@@ -68,6 +68,7 @@ class S3ContentModel(S3Model):
              "cms_tag",
              "cms_tag_post",
              "cms_comment",
+             "cms_attachment"
              )
 
     def model(self):
@@ -343,6 +344,7 @@ class S3ContentModel(S3Model):
         # Components
         add_components(tablename,
                        cms_comment = "post_id",
+                       cms_attachment = "post_id",
                        cms_post_layer = "post_id",
                        cms_post_module = "post_id",
                        cms_post_user = {"name": "bookmark",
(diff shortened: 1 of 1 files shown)


flavour:
There shouldn't be any need for this as I don't see that it does anything different to the existing doc_document resource?
This is already a component via the doc_id super_key

yashpalsaini:
Than what should to do in this
https://github.com/flavour/eden/blob/master/modules/s3db/cms.py#L515

flavour:
Comments don't yet have attachments, which might be useful

yashpalsaini:
In  this task
https://github.com/flavour/eden/blob/master/modules/s3db/cms.py#L211
can't we just add a new field like this
s3_datetime ("expireson",
         default = "now",
          label = T("Expireson"),
),

nursix:
Field name should probably be "expires_on", and the default ideally *not* "now" ;)

yashpalsaini:
move gis_location to link table and add expires_on field on cms_post
please have a look
https://github.com/yashpalsaini/eden/commit/5bb99e98cb9937d9950e2135883989c6585fb8e9

nursix:
Your new commit looks good - but changing the model is only half of the work required for this ToDo.

There is a bunch of dependent functionality in templates (e.g. MCOP, DRMP) that would break unless you adapt it. Without adapting them, I'm afraid this change cannot easily be merged.

Also note that both templates are in active use (i.e. with existing databases!) - and without migration script they would loose all the existing location links of any cms_post entries, so such a migration script would be a nice addition to your change (although not a hard requirement, as usually a task for the server admin).

nursix:
NB there may be more templates using the newsfeed, and there is also a general newsfeed list layout involving the location_id - so there is still a fair bit of work to do.

yashpalsaini:
so is it ok if i make out change out in MCOP, DRMP according to new location table

yashpalsaini:
move on doing some work on newsfeed in cms controllers for location_id

yashpalsaini:
try to write  lazy table for pr_image table
please have a look
https://github.com/yashpalsaini/eden/commit/6f6f7c828d269ffaa5a781f4d66a1a76c72044db
but i am getting error while executing it
please let me know if i am doing it wrong

yashpalsaini:
make changes in controllers/cms.py and modules/s3db/cms.py for adapting
changes made in location_id(basically create Posts <> Locations Link Table).
please let me know me if i am moving in right way.
Their still some work to do in some templates

yashpalsaini:
oops link
https://github.com/yashpalsaini/eden/commit/eec51a8e4a12ae054ffc994fdd868cd27d5e1084

yashpalsaini:
please have a look over changes so that i move on to some other templates
https://github.com/yashpalsaini/eden/commit/eec51a8e4a12ae054ffc994fdd868cd27d5e1084

PR #1073, opened 2015-03-08
Geonames search replaced with internal search of gis locations table.

alch4:
The geonames search feature has been replaced with a search of the internal DB via calling search_gis_locations() in controllers/gis.py. search_gis_locations returns a JSONP object similar in format to geonames.

Fallback on geonames if 0 results from search_gis_locations has not been implemented yet. Thus far search_gis_locations replaces geonames. One possible strategy for fallback on geonames would require some way of sending a HTTP GET request to the geonames inside the search_gis_locations method and decomposing the return JSONP into a JSON then into a dictionary. Unsure of how HTTP GET requests can be made in python - some guidance here would be most helpful.

A note on gis_location.level (gis.py lines 3804-3816): In order to make the GeoNamesSearchCombo.js zoom properly when selecting a result from the drop down list, we had to translate the level column from the gis_location table into ADM codes used by geonames. The translation is the best to our ability giving a logical ranking of levels 

Ex: Country should have a lower zoom value than a state because a country is larger than a state. Thus country is assigned the string PCL which causes the GeoNamesSearchCombo.js to set the zoom to 5. Likewise a county or a district is smaller than a state thus the zoom level for a country/district is higher.

### DIFF
--- a/controllers/gis.py
+++ b/controllers/gis.py
@@ -3782,6 +3782,7 @@ def screenshot():
 
 # =============================================================================
 def search_gis_locations():
+    import urllib2
     
     user_str = request.vars['name_startsWith']
     callback_func = request.vars['callback']
@@ -3800,7 +3801,7 @@ def search_gis_locations():
     for row in rows:
         count+=1
         result = {}
-        
+         
         #Convert the level colum into the ADM codes geonames returns
         #fcode = row["gis_location.level"]
         level = row["gis_location.level"]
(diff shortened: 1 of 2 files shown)


alch4:
For tracking this is for http://eden.sahanafoundation.org/ticket/655

anurag-ks:
I don't know much about geonames but for HTTP GET requests, can't you use
the request module ?

anurag-ks-29

alch4:
Closing pull request. New pull requested with new updates.

PR #1074, opened 2015-03-12
Updates from SahanaCamp Turkey branch

michaelhowden:
* Default "index" resources to summary method
* Adding Report Options & Defaults
* Removing extra embedded table if there is no Org Logo (return "")
* Menus check if skills and certificates are enabled
* Remove rheader borders & backgrounds
* Better introspect project menus, list_fields, filters and report_options


modules/s3db/org.py
```diff
@@ -511,6 +511,22 @@ def model(self):
...
 
+        report_fields = ["sector_organisation.sector_id",
+                         "organisation_organisation_type.organisation_type_id"]
+        report_options = Storage(rows = report_fields,
+                                 cols = report_fields,
+                                 fact = ["count(name)",
+                                         "list(name)",
+                                         ],
+                                 defaults=Storage(rows = "sector_organisation.sector_id",
+                                                  cols = "organisation_organisation_type.organisation_type_id",
```
nursix: Again - sector is not used by default, so making this the default report axis looks wrong.

modules/s3db/org.py
```diff
@@ -4132,6 +4150,21 @@ def model(self):
...
+            rows = report_fields,
+            cols = report_fields,
+            fact = ["count(id)",
+                    "list(name)",
+                    ],
+            defaults = Storage(rows = lfield, # Lowest-level of hierarchy
+                               cols = "office_type_id",
+                               fact = "count(id)",
+                               totals = True,
+                               chart = "barchart:rows",
```
nursix: My plea is to make spectrum:rows the default rather than barchar:rows

modules/s3db/project.py
```diff
@@ -60,7 +60,6 @@
            "project_jnap_opts",
            "project_pifacc_opts",
            "project_rfa_opts",
-           "project_project_filters",
```
nursix: I don't quite see why the function gets removed. Filter configuration of such complexity should be in a function (like hrm_human_resource_filters) is.

modules/s3db/org.py
michaelhowden: +1 to that and sectors in Org reports. Sorry I missed them in the review - I blame lambda

modules/s3db/project.py
michaelhowden: a) Only got called in one place
b) Filter configuration uses the same settings checks as report & list_fields configure - so best to do them all together
I could see an argument for a project_project_field_filter_report_form_config func - or *maybe* framework support to to this automatically...

nursix: Ok, fine. But this needs to be cleaned up (PEP8, logical order, append+insert rather than list concatenation with +).

The critical aspect here is that it should be easy to:
- easily see which filters/fields eventually appear in which order (which is much harder with this mix)
- be able to easily add a new filter/field in a template in the right place (which is unsafe if you can't easily see which filters/fields are there in which order)

I think the mix here makes this a lot more difficult and error-prone - unless it goes indeed field by field (i.e. filters corresponding to fields = same rules, same order). This is the most frequently customized CRUD page, so this must become more developer-friendly - while the tiny trade-off through repeated option checks seems irrelevant to me.

nursix: Look at project_project_filter - very clean function, base-widgets first, then adding extensions per setting. Makes it easy to see which filter widgets will be active for your configuration and in which order.

Your mix obscures this, and makes it much harder. I am pretty -1 to that, and I can easily accept repeated option checks (which are tiny) to retain this clarity.

nursix: So, if we move this inside the model-function, then it should be separated by filters, list fields and report options. Like:

    config1 = settings.get_bla_bla1()
    config2 = settings.get_bla_bla2()

    # Filters
    filter_widgets = [...base widgets...]
    if config1:
        filter_widgets.append(...)
    if config2:
        filter_widgets.append(...)

    # List fields
    list_fields = [...base list fields...]
    if config1:
        list_fields.append(...)
    if config2:
        list_fields.append(...)

    # Report options
    report_fields = [...base report fields...]
    if config1:
        report_fields.append(...)

That way, you can easily see what's going on, and where to insert your custom filter/list field/report field in the template. The repeated "if"s here cause no relevant overheads, as you call the getter only once on top.

modules/s3menus.py
```diff
@@ -1013,7 +1014,8 @@ def hrm():
                         M("Create", m="create"),
                     ),
                     M("Skill Catalog", f="skill",
-                      check=manager_mode)(
+                      check=[manager_mode,
```
nursix: Is there /any/ need for the line break here?

modules/s3menus.py
```diff
@@ -1029,7 +1031,8 @@ def hrm():
                         #M("Course Certificates", f="course_certificate"),
                     ),
                     M("Certificate Catalog", f="certificate",
-                      check=manager_mode)(
+                      check=[manager_mode,
```
nursix: or here?

modules/s3menus.py
```diff
@@ -1099,7 +1105,8 @@ def vol():
                         M("Create", m="create"),
                     ),
                     M("Skill Catalog", f="skill",
-                      check=manager_mode)(
+                      check=[manager_mode,
```
nursix: or here?

modules/s3menus.py
```diff
@@ -1115,7 +1122,8 @@ def vol():
                         #M("Course Certificates", f="course_certificate"),
                     ),
                     M("Certificate Catalog", f="certificate",
-                      check=manager_mode)(
+                      check=[manager_mode,
```
nursix: or here?

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+        multi_org = lambda i: settings.get_project_multiple_organisations()
+        community = lambda i: settings.get_project_community()
+        mode_3w = lambda i: settings.get_project_mode_3w()
+        mode_task = lambda i: settings.get_project_mode_task()
         if community:
-            IMPORT = "Import Project Communities"
+            local_label = "Project Communities"
         else:
-            IMPORT = "Import Project Locations"
+            local_label = "Project Locations"
```
nursix: All-uppercase for constants?

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+               check = mode_task)(
+                M("Create", m="create"),
+                M("Import", m="import", p="create"),
+             ),
+             M("Daily Work", f="time", check = mode_task,)(
+                M("My Logged Hours", vars={"mine":1}),
+                M("My Open Tasks", f="task", vars={"mine":1}),
+             ),
+             M("Reports", f="location", m="report")(
+                M("3W", f="location", m="report"),
```
nursix: mode_3w?

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+                M("Beneficiaries", f="beneficiary", m="report",
```
nursix: Not needed for mode_task.

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+                 M("Import", m="import", p="create"),
+                 M("Map", m="map"),
+                 M("List Contacts", f="location_contact"),
+              ),
+             M("Tasks", f="task", m="summary",
+               check = mode_task)(
+                M("Create", m="create"),
+                M("Import", m="import", p="create"),
+             ),
+             M("Daily Work", f="time", check = mode_task,)(
```
nursix: Trailing comma and closing bracket on the same line - either/or. PEP8=no blanks around keyword assignments.

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+                M("My Logged Hours", vars={"mine":1}),
```
nursix: PEP8=space after colon.

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
-                    M("Create", m="create"),
-                    M("Import", m="import", p="create"),
-                 ),
-                )
+        menu(
+             M("Projects", f="project", m="summary")(
+                 M("Create", m="create"),
+                 M("Import", m="import", p="create"),
+                 M("Map", f="location", m="map"),
+                 M("Open Tasks for Project", vars={"tasks":1}, check = mode_task,),
```
nursix: Same here: trailing comma and closing bracket on the same line => either/or

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
-                 M("Projects", f="project")(
-                    M("Create", m="create"),
-                    M("Import", m="import", p="create"),
-                 ),
-                )
+        menu(
+             M("Projects", f="project", m="summary")(
+                 M("Create", m="create"),
+                 M("Import", m="import", p="create"),
+                 M("Map", f="location", m="map"),
```
nursix: mode_3w?

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+                 M("List Contacts", f="location_contact"),
+              ),
+             M("Tasks", f="task", m="summary",
+               check = mode_task)(
+                M("Create", m="create"),
+                M("Import", m="import", p="create"),
+             ),
+             M("Daily Work", f="time", check = mode_task,)(
+                M("My Logged Hours", vars={"mine":1}),
+                M("My Open Tasks", f="task", vars={"mine":1}),
```
nursix: No check for STAFF role? Doesn't seem right.

nursix:
The project menu change seems to break a lot of things - not sure that is how we actually want to have it. What happened to the simple project-menu (i.e. neither of 3W/DRR/Task - but just projects tracking, i.e. simple list)? That should still be the default!

Very inconsistent style (PEP8, inconsistent indents, unnecessary line breaks).

controllers/cr.py
```diff
@@ -23,7 +23,7 @@ def index_alt():
     """
 
     # Just redirect to the list of Shelters
-    redirect(URL(f="shelter"))
+    redirect(URL(f="shelter", args = "summary"))
```
nursix: PEP8 = no blanks around keyword assignments.

But at least, do it consistently, not one so and another one differently.

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
         settings = current.deployment_settings
         #activities = lambda i: settings.get_project_activities()
         activity_types = lambda i: settings.get_project_activity_types()
-        community = settings.get_project_community()
+        multi_org = lambda i: settings.get_project_multiple_organisations()
+        community = lambda i: settings.get_project_community()
```
nursix: Dropped this? (not using it anymore?) Why?

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
-                 M("Sectors", f="sector",
-                   check=sectors)(
-                    M("Create", m="create"),
-                 ),
-                 M("Themes", f="theme",
-                   check=themes)(
-                    M("Create", m="create"),
-                 ),
-                )
+        menu = M(c="project")
```
nursix: What's the point now with this variable?

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+        menu(
+             M("Projects", f="project", m="summary")(
+                 M("Create", m="create"),
+                 M("Import", m="import", p="create"),
+                 M("Map", f="location", m="map"),
+                 M("Open Tasks for Project", vars={"tasks":1}, check = mode_task,),
+              ),
+             M("Project Organizations", f="organization", m="summary",
+                check = multi_org)(
+              M("Create", m="create"),
```
nursix: inconsistent indent

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+                               fact="sum(hours)",
+                               week=1)),
+                M("Last Month's Work", f="time", m="report",
+                  check = mode_task,
+                  vars=Storage(rows="person_id",
+                               cols="week",
+                               fact="sum(hours)",
+                               month=1)),
+                M("Project Time Report", f="time", m="report",
+                  check = mode_task,),
```
nursix: Trailing comma and closing bracket on the same line => either/or

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+             ),
+             # Is "Admin" the right term? "Master Data"? "Look Ups"? "Lists"? Copy this menu heading to other modules
```
nursix: "Administration" or "Master Data" - but definitely not "Admin" (that's not even a word)

nursix: Preferrably "Administration" ("Master Data" is difficult to translate)

nursix: "Lookups" is not translatable at all (no corresponding word in any language I know), "Lists" is too generic.

modules/s3menus.py
```diff
@@ -1528,126 +1540,87 @@ def project():
...
+             M("Daily Work", f="time", check = mode_task,)(
+                M("My Logged Hours", vars={"mine":1}),
+                M("My Open Tasks", f="task", vars={"mine":1}),
+             ),
+             M("Reports", f="location", m="report")(
+                M("3W", f="location", m="report"),
+                M("Beneficiaries", f="beneficiary", m="report",
+                  check = stats,
+                  ),
+                M("Funding", f="organisation", m="report"),
```
nursix: Not used in mode_task/mode_community?

modules/s3db/org.py
michaelhowden: OK - checked out spectrum:rows - couple of issues:
1) They are not represented in the chart options at the top of the chart (pie, bar, per Axis Breakdown)
2) I could only ever get the bar chart at the bottom to show one variable - which seems the wrong visualization tool (especially as it doesn't change size)
3) The interaction doesn't make sense to me... I can see how this might make sense with the breakdown - not sure how to enable that.
4) Not documented on http://eden.sahanafoundation.org/wiki/S3/S3Report
Only 2 & 3 are show stoppers for me adopting this
Do you have another example for me to check out?

modules/s3db/project.py
michaelhowden: Downside - not very DRY
but I've done it *just* to make you happy :)

modules/s3menus.py
michaelhowden: No

modules/s3menus.py
michaelhowden: -        community = settings.get_project_community()
was a mistake - this setting is only tested once - so no need for variable

modules/s3menus.py
michaelhowden: OK (although this is a variable)

modules/s3menus.py
michaelhowden: None

modules/s3menus.py
michaelhowden: Yes

modules/s3menus.py
michaelhowden: removed comma

modules/s3menus.py
michaelhowden: Fixed

modules/s3menus.py
michaelhowden: Yes - and I stand by this:
1) STAFF isn't a system role (you could have a time tracking template without this role)
2) STAFF isn't the only role to have tasks assigned to them (Vol?)

modules/s3menus.py
michaelhowden: 3W report only in  mode_3w... I *guess* that makes sense... :)

modules/s3menus.py
michaelhowden: Is there are better check to make?

modules/s3menus.py
michaelhowden: Could be - but only if also mode_3w - added that test

modules/s3menus.py
michaelhowden: This is just the Org report - not ness funding - why have it?

modules/s3menus.py
michaelhowden: Done

modules/s3menus.py
michaelhowden: Hmmm... I don't like Administration because it has a different meaning - it's not just the administrator who can look / edit these. But It's the best option, so have adopted it

michaelhowden:
Revised and (hopefully) ready to merge

modules/s3db/org.py
nursix: I have no idea what you're talking about - 1-3 are not true.

modules/s3db/project.py
nursix: Then you've not understood what I mean.

nursix: ...because it is totally DRY

modules/s3menus.py
nursix: Actually, the /best/ option would be "reference data" - because that is exactly what it is.

But I suppose you won't like it because it partially exposes the model (i.e. technical term).

Dominic

modules/templates/default/config.py
```diff
@@ -501,6 +501,8 @@ def config(settings):
     # Organisations
     # Uncomment to use an Autocomplete for Organisation lookup fields
     #settings.org.autocomplete = True
+    # Enable the Organisation Sector field
+    #settings.org.branches = True
```
nursix: Should be "settings.org.sectors" no?

modules/s3menus.py
```diff
@@ -1526,130 +1534,99 @@ def project():
...
-                    M("Import", m="import", p="create"),
-                 ),
-                )
+            PROJECT_LOCATION_LABEL = "Project Locations"
+
+        ADMIN = current.session.s3.system_roles.ADMIN
 
-        return menu
+        return M(c="project")(
+             M("Programmes", f="programme", m="summary",
```
nursix: This isn't ready yet! Please don't expose yet.

modules/s3db/project.py
```diff
@@ -267,36 +264,151 @@ def model(self):
...
+            append("partner.organisation_id")
+            append("donor.organisation_id")
             append((T("Total Funding Amount"), "total_organisation_amount"))
         if multi_budgets:
             append((T("Total Annual Budget"), "total_annual_budget"))
-        list_fields += ["start_date",
-                        "end_date",
-                        "location.location_id",
-                        ]
+        extend(["start_date",
```
nursix: Is there any need for the line break here? Can't this be a tuple? Trailing comma and closing bracket on the same line => either/or

modules/s3db/project.py
```diff
@@ -267,36 +264,151 @@ def model(self):
...
-        report_fact_fields = [(field, "count") for field in report_fields]
-        report_fact_default = "project.organisation_id"
-        #report_fact_default = "theme.name"
+        # Report Options
+        report_fields = ["status_id",
+                         "organisation_id"]
+        append = report_fields.append
+        extend = report_fields.extend
+        if programmes:
+            append("programme_project.name")
```
nursix: Hmm, programme_project has no "name" => must be programme_id

controllers/cr.py
nursix: Not fixed.

modules/s3db/org.py
```diff
@@ -397,6 +397,14 @@ def model(self):
                        "comments",
                        ]
 
+        use_sector = settings.get_org_sector()
+        if use_sector:
+            form_fields.insert(3, S3SQLInlineLink("sector",
+                                                  columns = 4,
+                                                  label = T("Sectors"),
+                                                  field = "sector_id",),
```
nursix: Trailing comma and closing bracket on the same line => either/or

modules/s3db/org.py
```diff
@@ -437,6 +445,8 @@ def model(self):
                        "acronym",
                        "comments",
                        ]
+        if use_sector:
+            text_fields.append("sector_organisation.sector_id")
```
nursix: Not a text field!

modules/s3db/org.py
```diff
@@ -490,6 +500,16 @@ def model(self):
                                           widgets = org_widgets,
                                           )
 
+        list_fields = ["id",
+                       "name",
+                       "acronym",
+                       "organisation_organisation_type.organisation_type_id",
+                       "website"
```
nursix: No trailing comma despite closing bracket on the next line :D

modules/s3db/org.py
```diff
@@ -498,18 +518,38 @@ def model(self):
...
+                                                     options = lambda: \
+                                                     get_s3_filter_opts("org_sector",
+                                                                        location_filter=True,
+                                                                        none=True,
+                                                                        translate=True),
+                                                     ),
+                                  )
+
+        report_fields = ["organisation_organisation_type.organisation_type_id",
+                         "country"
```
nursix: No trailing comma and yet closing bracket on the next line ;)

modules/s3db/org.py
```diff
@@ -498,18 +518,38 @@ def model(self):
...
-                                                   translate=True),
-                            ),
             S3OptionsFilter("country",
                             #label = T("Home Country"),
                             ),
             ]
+        if use_sector:
+            filter_widgets.insert(2, S3OptionsFilter("sector_organisation.sector_id",
+                                                     options = lambda: \
+                                                     get_s3_filter_opts("org_sector",
```
nursix: NB That's actually deprecated - supported by S3OptionsFilter directly.

modules/s3db/org.py
```diff
@@ -4100,14 +4133,18 @@ def model(self):
...
-                       "name",
-                       "organisation_id", # Filtered in Component views
-                       "office_type_id",
-                       ]
+        report_fields = ["name",
+                         "organisation_id", # Filtered in Component views
+                         "office_type_id",
+                         ]
 
+        list_fields = ["id"] + report_fields
```
nursix: "id" is included automatically, no?

modules/s3db/org.py
nursix: To clarify: 

1) Spectrum is one of the breakdown-options (because it's a two-axis chart), and very well shown in the chart options.

2) Spectrum consists of a half-donut chart showing the distribution along the first axis (=the "spectrum"), and a barchart showing the distribution along the second axis.

Using the drop-down, you can additionally filter the second axis by the individual data points of the first axis. Alternatively, you can just click on the data points in the half-donut.

3) The interaction does absolutely make sense to me. Example:

A "spectrum" of offices by office type (second axis = Organisations):

ss42

...shows me the distribution of offices per office type (the half-donut), ...and per organisation (the bars).

Now I can use the drop-down to select one specific office type - which filters the per-organisation bars to that type:

ss43

ss44

Each bar presents me an onhover-tooltip, and clicking on that bar takes me to a list of offices filtered by this office type for this organisation.

ss45

The advantage over a standard barchart is that you have this filter in the chart, and can thus easily walk through the data points of the first axis.

nursix: Note that "spectrum" means the half-donut, not the bars ;)

Hence when you do spectrum:rows, then the rows will appear as the half-donut, and the cols will be the bars.

nursix: For obvious reasons, though, the chart title uses the second axis (because the user would naturally see this as a bar chart that is filtered by the first axis, and not as a half donut that is broken down by the second axis).

However, technically it's the latter - the first axis gives the half-donut, the second axis gives the bars.

nursix: Or in other words: that the bars only show one axis (=offices per organisation) is actually the point of it.

Barchart:rows does nothing else (it also shows exactly one axis = offices per organisation). Lets compare that:

ss46

The advantage of spectrum over a standard barchart is the option to "focus" a single data point on the other axis, (which is btw something that you suggested as being a useful interaction):

ss42

Apart from that, it's "just" a barchart like barchart:rows.

nursix: Another nice example - hours worked per person.

A normal barchart shows me this:
ss49

Spectrum shows me the same:
ss47

...but additionally allows me to focus on a particular project:
ss48

nursix: ...of course I could also use the filter form to "focus" on a particular project.

But that wouldn't show the project-context in the chart, and it would require an additional server-side request hence be much slower, and not exactly provide for a walk-through through the individual projects.

And of course you wouldn't see the particular project's total hours put in relationship to the overall hours worked (=the half donut) - but that's just an additional goody.

In the case of offices: if you make the Lx the barchart-axis, then you can see offices per location in the barchart:rows. 

With spectrum:rows you achieve the same, but you can /additionally/ focus on a particular office type - directly inside the chart, without server-side request - which makes totally sense to me.

Of course, you can also produce a chart that shows the offices per location broken down by type, as grouped or stacked bars. That too makes sense.

But I'm not arguing that "spectrum" would be an alternative to "breakdown" - I'm saying it is a better *barchart*.

michaelhowden: OK - that clarifies some things for me (this would be much better on the list - do you mind if I move the conversation there?) - couple of things remain:
* I was running into issues because my secondary axis didn't have any meaningful data - I really wanted to chart the primary axis. This is just a simple case of switching the axises (either I will in config or implementation if you think it belongs there)
* I think that it is a bit ambiguous what the spectrum represents - it could benefit from a title
* I think that the chart could work better with 3 separate titles Fact, Primary Axis & Secondary Axis. The "X per Y" is a bit confusing. Also - if fact =count(id) I don't think that a label is needed - it's pretty implicit
* I'm not sure that the Org Type drop down adds value. It repeats the filter functionality of the spectrum - perhaps if it were under the spectrum it would act more like the title I'm seeking?
* The chart labels have multiple 1's , 2's etc - but that's the same with all graphs
I will spend a bit more time investigating this. Perhaps the biggest point right now is that I think this introduces more complexity into the basic chart. With that comes value - but I think it is better to introduce something very simple to users - and allow them to dig deeper into the complexity (which they can)
Perhaps hiding the spectrum by default and having a simple ability to show it would be a better approach?

michaelhowden: NOTE: This isn't a refusal to use it - I just think that this needs to be shared with the community - then made a global default.

modules/s3menus.py
michaelhowden: OK - but this was controlled by settings....

modules/templates/default/config.py
michaelhowden: Yes - thanks!

michaelhowden:
Take 3... :)

michaelhowden:
Take 4 (Thank you Travis - and great error reporting too!)

modules/s3db/org.py
nursix: I would disagree on all points here, and I think that it is NIH that motivates them rather than usability.

nursix: Hmm, it has been shared with the community, and it doesn't make sense as global default.

modules/s3db/project.py
```diff
@@ -1546,7 +1645,7 @@ class S3ProjectAnnualBudgetModel(S3Model):
         This model holds the annual budget entries for projects
     """
 
-    names = ("project_annual_budget",)
+    names = ("project_annual_budget")
```
nursix: This needs a trailing comma (otherwise not recognized as tuple)

modules/s3db/project.py
```diff
@@ -1690,7 +1789,7 @@ def model(self):
                      # - can't override field name, ondelete or requires
                      super_link("parameter_id", "stats_parameter",
                                 empty = False,
-                                instance_types = ("project_beneficiary_type",),
+                                instance_types = ("project_beneficiary_type"),
```
nursix: Same here - must have a trailing comma to be recognized as tuple.

modules/s3db/project.py
```diff
@@ -3016,7 +3115,7 @@ def model(self):
                                       "fkey": "pe_id",
                                       "pkey": "person_id",
                                       "filterby": "contact_method",
-                                      "filterfor": ("EMAIL",),
+                                      "filterfor": ("EMAIL"),
```
nursix: Without trailing comma, this is just a string, so no brackets needed either.

nursix:
Hmm, now you have even removed trailing commas of single-item tuples, which makes it a syntax error. 

    x = ("a",) # tuple
    y = ("b") # simple string

The biggest problem with that is that both tuple and string are iterable - so this may not raise an Exception immediately.

With multiple items, it's a style issue:

    x = ("a", "b",) # trailing comma and closing bracket on the same line

This does basically defeat the purpose of the trailing comma, which is to be able to easily add another item. Therefore, if you can expect other items to be added, it should be:

    x = ("a",
         "b",
         )

That does also apply to "names" of models, even with a single item:

    names = ("my_table",
             )

...because this allows easy adding of names by just inserting a line (=no editing of other lines required):

    names = ("my_table",
             "my_table_id",
             )

I thought I had marked exactly those which needed to be corrected?

nursix:
Note that this does not apply to constructed tuples, i.e. where the tuple is left of a =

    (x, y) = example()

In this case, there should be neither a trailing comma nor a line break.

michaelhowden:
You did make the ones needing changing - I got a bit over zealous.... fixed now.
Take 5?

nursix:
Hmm, still a bunch of coding convention issues, including PEP8. What shall I do now? 

It seems that every round you're fixing 2/3 of them and introducing a couple of fresh ones - so this is going to be a never-ending review I'm afraid. Let me try one more time to indicate all of them - and pls fix only them, but all of them.

I'll give you notice when ready...hang on.

modules/s3db/org.py
```diff
@@ -397,6 +397,14 @@ def model(self):
                        "comments",
                        ]
 
+        use_sector = settings.get_org_sector()
+        if use_sector:
+            form_fields.insert(3, S3SQLInlineLink("sector",
+                                                  columns = 4,
+                                                  label = T("Sectors"),
+                                                  field = "sector_id"),
```
nursix: Mixed style, should be like:

    field = "sector_id",
    ),

modules/s3db/org.py
nursix: This is not fixed. If you have the closing bracket on a new line, then you are inviting line swaps or insertions, hence there should be a trailing comma after "website".

modules/s3db/org.py
```diff
@@ -498,18 +519,32 @@ def model(self):
...
+                                 fact = ["count(id)",
+                                         "list(name)",
+                                         ],
+                                 defaults=Storage(rows = "organisation_organisation_type.organisation_type_id",
+                                                  cols = "country",
+                                                  fact = "count(id)",
+                                                  totals = True,
+                                                  chart = "spectrum:rows",
+                                                  #table = "collapse",
+                                                  )
```
nursix: This lacks a trailing comma /after/ the closing bracket

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
-        list_fields = ["id"]
+        # Which levels of Hierarchy are we using?
+        levels = current.gis.get_relevant_hierarchy_levels()
+        location_fields = ["location.location_id$%s" % level for level in levels]
+
+        # List Fields 
+        list_fields = ["id",
+                       "name",
+                       "status_id",
+                       "organisation_id"]
```
nursix: This should rather be trailing comma plus closing bracket on a new line (since extensible).

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
-                        "end_date",
-                        "location.location_id",
-                        ]
+        extend(("start_date", "end_date"))
+
+        # Filter widgets
+        text_filter_fields = ["name",
+                              "status_id",
+                              "description",
+                              "organisation_id"]
```
nursix: Similar here - better to have a trailing comma and the closing bracket on a new line

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+        append_text_filter_fields = text_filter_fields.append
+        extend_text_filter_fields = text_filter_fields.extend
+        if use_codes:
+            append_text_filter_fields("code")
+        if mode_3w:
+            extend_text_filter_fields(location_fields)
+            append(S3LocationFilter("location.location_id",
+                                    # Default should introspect
+                                    levels = levels,
+                                    hidden = True)
```
nursix: Similar here - better to have:

    hidden = True,
    )

...to allow easy adding of options or line swaps. Ideally, options should be in alphabetical order.

nursix: But be sure to NOT have a trailing comma /after/ the closing bracket in append (that'd be a syntax error)

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                   )
+        if programmes:
+            append(S3OptionsFilter("programme_project.programme_id",
+                                   label = T("Programme"),
+                                   hidden = True)
```
nursix: Same here.

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                   )
+        if use_sectors:
+            append_text_filter_fields("sector.name")
+            append(S3OptionsFilter("sector_project.sector_id",
+                                                  location_filter = True,
```
nursix: Wrong indent, options should be in alphabetical order, ideally a trailing comma after the last option + closing bracket on a new line.

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                                                  none = True,
+                                                  hidden = True)
+                                  )
+        if mode_drr:
+            append_text_filter_fields("hazard.name")
+            append(S3OptionsFilter("hazard_project.hazard_id",
+                                    label = T("Hazard"),
+                                    help_field = project_hazard_help_fields,
+                                    cols = 4,
+                                    hidden = True)
```
nursix: Same here.

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                                    )
+        if use_themes:
+            append_text_filter_fields("theme.name")
+            append(S3OptionsFilter("theme_project.theme_id",
+                                    label = T("Theme"),
+                                    help_field = project_theme_help_fields,
+                                    cols = 4,
+                                    hidden = True)
```
nursix: ...and here

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                   )
+        if mode_drr:
+            hfa_opts = project_hfa_opts()
+            options = dict((key, "HFA %s" % key) for key in hfa_opts)
+            append(S3OptionsFilter("drr.hfa",
+                                    label = T("HFA"),
+                                    options = options,
+                                    help_field = hfa_opts,
+                                    cols = 5,
+                                    hidden = True)
```
nursix: ...and here.

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                    )
+        if multi_orgs:
+            append_text_filter_fields("partner.organisation_id")
+            append_text_filter_fields("donor.organisation_id")
+            append(S3OptionsFilter("partner.organisation_id",
+                                   label = T("Partners"),
+                                   hidden = True)
```
nursix: and here.

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                   )
+            append(S3OptionsFilter("donor.organisation_id",
+                                   label = T("Donors"),
+                                    hidden = True)
```
nursix: and here (+wrong indent)

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                   )
+        filter_widgets.insert(0,S3TextFilter(text_filter_fields,
```
nursix: There should be a space after the comma:

    filter_widgets.insert(0, S3TextFilter(...

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                              )
 
-        report_fields = list_fields
-        report_col_default = "location.location_id"
-        report_fact_fields = [(field, "count") for field in report_fields]
-        report_fact_default = "project.organisation_id"
-        #report_fact_default = "theme.name"
+        # Report Options
+        report_fields = ["status_id",
+                         "organisation_id"]
```
nursix: Ideally have a trailing comma here plus closing bracket on a new line.

Or - if you don't expect more fields to be added, then the line break is questionable.

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+            append("drr.hfa")
+        if multi_orgs:
+            append("partner.organisation_id")
+            append("donor.organisation_id")
+        report_options = Storage(rows = report_fields,
+                                 cols = report_fields,
+                                 fact = ["count(id)","list(name)"] + report_fields,
+                                 defaults = Storage(rows = "organisation_id",
+                                                    cols = "status_id",
+                                                    fact = "count(id)"),
```
nursix: Like this:

    fact="count(id)",
    ),

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+        if mode_3w:
+            append((T("Themes"), "theme.name"))
+        if mode_drr:
+            append("drr.hfa")
+        if multi_orgs:
+            append("partner.organisation_id")
+            append("donor.organisation_id")
+        report_options = Storage(rows = report_fields,
+                                 cols = report_fields,
+                                 fact = ["count(id)","list(name)"] + report_fields,
```
nursix: Missing space after the comma in the list. Should be:

    ["count(id)", "list(name)"]

Mixing method specification and simple field selectors in "fact" is not ideal.

nursix: Additionally - specifying a label for each fact is recommended, for two reasons:
a) auto-generated fact labels do not appear in language files (hence not visible for translators)
b) auto-generated fact labels just indicate the aggregation formula, but do not put the fact into the report context, hence not as user-friendly

modules/s3menus.py
```diff
@@ -1029,7 +1030,7 @@ def hrm():
                         #M("Course Certificates", f="course_certificate"),
                     ),
                     M("Certificate Catalog", f="certificate",
-                      check=manager_mode)(
+                      check=[manager_mode, certificates])(
```
nursix: This can be a tuple

modules/s3menus.py
```diff
@@ -1013,7 +1014,7 @@ def hrm():
                         M("Create", m="create"),
                     ),
                     M("Skill Catalog", f="skill",
-                      check=manager_mode)(
+                      check=[manager_mode, skills])(
```
nursix: This can be a tuple

modules/s3menus.py
```diff
@@ -1099,7 +1103,7 @@ def vol():
...
```
nursix: This can be a tuple

modules/s3menus.py
```diff
@@ -1115,7 +1119,7 @@ def vol():
                         #M("Course Certificates", f="course_certificate"),
                     ),
                     M("Certificate Catalog", f="certificate",
-                      check=manager_mode)(
+                      check=[manager_mode, certificates])(
```
nursix: This can be a tuple

modules/s3menus.py
```diff
@@ -1449,16 +1453,20 @@ def org():
...
-                    M("Offices", f="office")(
+                    M("Offices", f="office", m="summary")(
                         M("Create", m="create"),
                         M("Map", m="map"),
                         M("Import", m="import")
                     ),
-                    M("Facilities", f="facility", m="summary")(
+                    M("Faciliites", f="facility", m="summary")(
+                        M("Create", m="create"),
+                        M("Import", m="import")
```
nursix: Should have a trailing comma here

modules/s3menus.py
```diff
@@ -1526,130 +1534,94 @@ def project():
...
+             M("Reports", f="location", m="report")(
+                M("3W", f="location", m="report",
+                  check=mode_3w),
+                M("Beneficiaries", f="beneficiary", m="report",
+                  check=[mode_3w, stats],
+                  ),
+                M("Activity Report", f="activity", m="report"),
+                M("Last Week's Work", f="time", m="report",
+                  check=mode_task,
+                  vars=Storage(rows="person_id",
```
nursix: vars does not need to be a Storage (Storage is not wrong, but a wee bit slower in construction than a dict)

modules/s3menus.py
```diff
@@ -1526,130 +1534,94 @@ def project():
...
+                               fact="sum(hours)",
+                               month=1)),
+                M("Project Time Report", f="time", m="report",
+                  check=mode_task),
+             ),
+             M("Administration", restrict=[ADMIN])(
+                M("Activity Types", f="activity_type"),
+                M("Beneficiary Types", f="beneficiary_type",
+                  check=stats),
+                M("Demographics", f="demographic",check=stats),
```
nursix: space after comma

modules/s3menus.py
```diff
@@ -1526,130 +1534,94 @@ def project():
...
+                M("Hazards", f="hazard", check=hazards),
+                M("Partner Organizations",  f="partners"),
```
nursix: ...but only one space after comma, not two.

modules/templates/default/config.py
nursix: You say yes - but you don't fix it?

nursix:
Ok, I think I'm through.

Where I say "can", these are just tips how things could be improved - but not required to action them, since your variant would be correct.

modules/s3db/project.py
```diff
@@ -267,36 +264,149 @@ def model(self):
...
+                                   label = T("Partners"),
+                                   hidden = True)
+                   )
+            append(S3OptionsFilter("donor.organisation_id",
+                                   label = T("Donors"),
+                                    hidden = True)
+                   )
+        filter_widgets.insert(0,S3TextFilter(text_filter_fields,
+                                             label = T("Search"),
+                                             comment = T("Search for a Project."))
```
nursix: Better have a trailing comma and closing bracket on new line here too.

michaelhowden:
@flavour Could you merge? It passed Dominic's review. I've just rebased on latest code to save you the hassle. Happy Easter!

nursix:
I have reviewed it but you haven't yet fixed all the things I've pointed to.

On a side note: The redirects in index() controllers tend to swallow errors (especially when it's multiple consecutive redirects like index=>index_alt=>summary), so users are left without information that something went wrong, and what went wrong.

If you redirect from index, you must make sure that you preserve response.error, response.warning and response.confirmation (via session) - and obviously, that needs to be fixed in cms too.

PR #1076, opened 2015-03-16
OCR Module Included

opravipati:


flavour:
This is disabled for a reason, so no need to enable by default :)

PR #1079, opened 2015-04-06
Font download task

anurag-ks:
This downloads the font files during pre-pop.

### DIFF
--- a/models/zzz_1st_run.py
+++ b/models/zzz_1st_run.py
@@ -233,6 +233,7 @@ if len(pop_list) > 0:
     s3.import_role = bi.import_role
     s3.import_script = bi.import_script
     s3.import_user = bi.import_user
+    s3.import_font = bi.import_font
 
     # Relax strict email-matching rule for import updates of person records
     email_required = settings.get_pr_import_update_requires_email()
--- a/modules/s3/s3import.py
+++ b/modules/s3/s3import.py
@@ -37,6 +37,7 @@ __all__ = ("S3Importer",
 import cPickle
 import os
 import sys
+import requests
 import urllib2          # Needed for error handling on fetch
 import uuid
 from copy import deepcopy
@@ -3957,6 +3958,39 @@ class S3BulkImporter(object):
                 create_role(rulelist[0],
                             rulelist[1],
                             *acls[rulelist[0]])
+
+    # -------------------------------------------------------------------------
+    def import_font(self, url):
+        """
+            Import a Font File
+        """
+
+        filename = url.split("/")[-1]
+        extension = url.split(".")[-1]
+        response = requests.get(url)
+        if response.status_code == 200:
+            data = response.content
+
+            if extension == "zip":
+                import zipfile
+                zf = zipfile.ZipFile(StringIO(data))
+                for fn in zf.namelist():
+                    data = zf.read(fn)
+
+            if extension == "gz":
+                import tarfile
+                tf = tarfile.open(fileobj=StringIO(data))
+                for fn in tf.getnames():
+                    data = tf.extractFile(fn).read()
+
+            f = open(os.path.join(current.request.folder,
+                                  "static",
+                                  "fonts",
+                                  filename), "wb")
+            f.write(data)
+            f.close()
+        else:
+            current.log.error("Network Error, Unable to import font")
 
     # -------------------------------------------------------------------------
     def import_user(self, filename):
--- a/modules/s3cfg.py
+++ b/modules/s3cfg.py
@@ -92,6 +92,40 @@ class S3Config(Storage):
                     "zh-cn": "%Y-%m-%d",
                     "zh-tw": "%Y/%m/%d",
                     }
+    # Custom fonts for each language
+    # fontset format -> [normal-version, bold-version]
+    font = {"ar": ["Helvetica", "Helvetica-Bold"],
+            "bs": ["Helvetica", "Helvetica-Bold"],
+            "de": ["Helvetica", "Helvetica-Bold"],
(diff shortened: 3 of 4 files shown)


modules/s3cfg.py
```diff
@@ -92,6 +92,40 @@ class S3Config(Storage):
                     "zh-cn": "%Y-%m-%d",
                     "zh-tw": "%Y/%m/%d",
                     }
+    # Custom fonts for each language
+    # fontset format -> [normal-version, bold-version]
+    font = {"ar": ["Helvetica", "Helvetica-Bold"],
```
nursix: Why do we need to download a font for every language? If one of them is the default, why repeat it?

modules/s3cfg.py
```diff
@@ -1435,6 +1469,10 @@ def get_pdf_excluded_fields(self, resourcename):
 
         return excluded_fields
 
+    def get_pdf_export_font(self):
```
nursix: Docstring?

modules/s3/s3import.py
```diff
@@ -37,6 +37,7 @@
 import cPickle
 import os
 import sys
+import requests
```
nursix: We should not use two libraries for the same purpose (we use urllib2 instead of requests)

modules/s3/s3import.py
```diff
@@ -3959,6 +3960,39 @@ def parseACL(_acl):
...
+                    data = tf.extractFile(fn).read()
+
+            f = open(os.path.join(current.request.folder,
+                                  "static",
+                                  "fonts",
+                                  filename), "wb")
+            f.write(data)
+            f.close()
+        else:
+            current.log.error("Network Error, Unable to import font")
```
nursix: Too little information - which font?

modules/s3/s3import.py
```diff
@@ -3959,6 +3960,39 @@ def parseACL(_acl):
...
+
+        filename = url.split("/")[-1]
+        extension = url.split(".")[-1]
+        response = requests.get(url)
+        if response.status_code == 200:
+            data = response.content
+
+            if extension == "zip":
+                import zipfile
+                zf = zipfile.ZipFile(StringIO(data))
```
nursix: Not checking any errors here?

modules/s3/s3import.py
```diff
@@ -3959,6 +3960,39 @@ def parseACL(_acl):
...
+                for fn in zf.namelist():
+                    data = zf.read(fn)
+
+            if extension == "gz":
+                import tarfile
+                tf = tarfile.open(fileobj=StringIO(data))
```
nursix: Not checking any errors here?

modules/s3/s3import.py
```diff
@@ -3959,6 +3960,39 @@ def parseACL(_acl):
...
+                for fn in tf.getnames():
+                    data = tf.extractFile(fn).read()
+
+            f = open(os.path.join(current.request.folder,
```
nursix: Not checking any errors here?

We do not normally give write access to static/*

modules/s3/s3import.py
```diff
@@ -3959,6 +3960,39 @@ def parseACL(_acl):
                             *acls[rulelist[0]])
 
     # -------------------------------------------------------------------------
+    def import_font(self, url):
+        """
+            Import a Font File
+        """
+
+        filename = url.split("/")[-1]
+        extension = url.split(".")[-1]
```
nursix: URLs can be a lot more complex than that. And what about local files?

models/zzz_1st_run.py
```diff
@@ -233,6 +233,7 @@
     s3.import_role = bi.import_role
     s3.import_script = bi.import_script
     s3.import_user = bi.import_user
+    s3.import_font = bi.import_font
```
nursix: Where is this used? Why does it need to be in s3?

modules/s3cfg.py
```diff
@@ -1435,6 +1469,10 @@ def get_pdf_excluded_fields(self, resourcename):
...
+        language = current.session.s3.language
+        return self.font.get(language)
```
nursix: This is not what I would call a config setting. How would we configure it?

modules/s3cfg.py
nursix: It does btw. seem to duplicate this function?

nursix:
Both the  import_font function and the PDF config setting do already exist - you just redefine them, but for what purpose? If your actual intention was to improve them, you should modify the existing functions rather than redefining them.

However, I don't quite see what you wanted to improve - what effectively does it do differently, and why?

anurag-ks:
Hi nursix,
I was working on this during GCI, here is all the previous talk we did - https://github.com/edengci/eden/pull/80

nursix:
That doesn't answer my question: the functionality already exists in trunk, and your PR just duplicates the existing functions, which is clearly wrong. 

If you intended to improve it, then what exactly did you try to improve? If you can explain it, then I may be able to guide you to it, otherwise I'm afraid this PR is not acceptable.

nursix:
Existing functionality is here:
https://github.com/flavour/eden/blob/master/modules/s3/s3import.py#L4084

So, I don't see the point with this pull request - especially since you're not modifying the existing functionality, but adding duplicates in the same places.

I've given a number of pointers for possible improvements, so if you wanted to action any of those and then re-submit your PR maybe it would make start to make sense.

But right now it doesn't really, sorry.

anurag-ks:
Hmm, I didn't knew someone implemented it already, must have been between the time period of GCI and now, since it's already their I am closing the PR.

PR #1080, opened 2015-04-26
Rename README to README.md

PeterDaveHello:
So that GitHub will parse it as a markdown document and we can make it more friendly and rich.


PeterDaveHello:
anybody here?

flavour:
Sorry, I was away when this was submitted & then it got drowned in my InBox.

I just had a check & it seems that the Web2Py Admin interface uses ABOUT & LICENSE but not README, so that doesn't appear to be a blocker, so I guess this is a useful step, thanks :)

PeterDaveHello:
thanks

PR #1081, opened 2015-04-26
improve zh-tw translation

PeterDaveHello:


PeterDaveHello:
anybody here?

flavour:
Very hard for me to review this since every line has been changed...I guess this is due to line-endings.
The key things I look for are correct use of variables as this can crash if incorrect.
We have 2 optiosn for proceeding: (A) You resubmit without changing line-endings so I can see what you've changed or (B) I delegate responsibility for this translation to you.
Can I assume that you've done at least partial testing with this new file?
Do you have a live deployment of the software?

PeterDaveHello:
I'll fix it soon and then ping you, thanks!

PeterDaveHello:
@flavour sorry for the very late response, this got forgotten totally.

I checked the diff and looks like the line-endings weren't changed, all the changed lines should contain translation update, would you please take a look?

PR #1082, opened 2015-04-29
Misc Changes

michaelhowden:
* Default "index" resources to summary method
* Adding Report Options & Defaults
* Removing extra embedded table if there is no Org Logo (return "")
* Menus check if skills and certificates are enabled
(introspect project menus, list_fields, filters and report_options have been removed as these will now have to be re-implemented on updated code)

### DIFF
--- a/controllers/asset.py
+++ b/controllers/asset.py
@@ -142,13 +142,6 @@ def supplier():
         msg_list_empty = T("No Suppliers currently registered")
         )
 
-    # Modify filter_widgets
-    filter_widgets = s3db.get_config("org_organisation", "filter_widgets")
-    # Remove type (always 'Supplier')
-    filter_widgets.pop(1)
-    # Remove sector (not relevant)
-    filter_widgets.pop(1)
-
     return s3db.org_organisation_controller()
 
 # -----------------------------------------------------------------------------
--- a/controllers/inv.py
+++ b/controllers/inv.py
@@ -453,12 +453,6 @@ def supplier():
         msg_list_empty = T("No Suppliers currently registered")
         )
 
-    # Modify filter_widgets
-    filter_widgets = s3db.get_config("org_organisation", "filter_widgets")
-    # Remove type (always 'Supplier')
-    filter_widgets.pop(1)
-    # Remove sector (not relevant)
-    filter_widgets.pop(1)
 
     # Open record in this controller after creation
     s3db.configure("org_organisation",
--- a/modules/s3db/org.py
+++ b/modules/s3db/org.py
@@ -397,6 +397,15 @@ class S3OrganisationModel(S3Model):
                        "comments",
                        ]
 
+        use_sector = settings.get_org_sector()
+        if use_sector:
+            form_fields.insert(3, S3SQLInlineLink("sector",
+                                                  columns = 4,
+                                                  label = T("Sectors"),
+                                                  field = "sector_id",
+                                                  ),
+                               )
+
         if settings.get_org_summary():
             # Include Summary fields in form
             position = form_fields.index("year")
@@ -437,6 +446,8 @@ class S3OrganisationModel(S3Model):
                        "acronym",
                        "comments",
                        ]
+        if use_sector:
+            text_fields += ["sector.name", "sector.abrv"]
 
         if settings.get_L10n_translate_org_organisation():
             text_fields.extend(("name.name_l10n",
@@ -490,26 +501,60 @@ class S3OrganisationModel(S3Model):
                                           widgets = org_widgets,
                                           )
 
-        filter_widgets = [
-            S3TextFilter(text_fields,
-                         label = T("Search"),
-                         comment = text_comment,
-                         #_class = "filter-search",
-                         ),
-            # NB Order is important here - gets popped in asset & inv controllers & IFRC template
-            type_filter,
-            # NB Order is important here - gets popped in asset & inv controllers & IFRC template
-            S3OptionsFilter("sector_organisation.sector_id",
-                            options = lambda: \
-                                get_s3_filter_opts("org_sector",
-                                                   location_filter=True,
-                                                   none=True,
-                                                   translate=True),
-                            ),
-            S3OptionsFilter("country",
-                            #label = T("Home Country"),
-                            ),
-            ]
+        list_fields = ["id",
+                       "name",
+                       "acronym",
+                       "organisation_organisation_type.organisation_type_id",
+                       "website"
+                       ]
+
+        if use_sector:
+            list_fields.insert(4,"sector_organisation.sector_id")
+
+        filter_widgets = [S3TextFilter(text_fields,
+                                       label = T("Search"),
+                                       comment = text_comment,
+                                       #_class = "filter-search",
+                                       ),
+                          ]
+        append = filter_widgets.append
+
+        # Don't add Type or Sector Filters for Supplier organizations in the asset and inv controllers
+        if current.request.function != "supplier":
+            append(type_filter) 
+            if use_sector:
+                append(S3OptionsFilter("sector_organisation.sector_id",
+                                       options = lambda: \
+                                           get_s3_filter_opts("org_sector",
+                                                              location_filter=True,
+                                                              none=True,
+                                                              translate=True),
+                                       )
+                       )
+
+        append(S3OptionsFilter("country",
+                               #label = T("Home Country"),
+                                ),
+               ) 
+
+        report_fields = ["organisation_organisation_type.organisation_type_id",
+                         "country",
+                         ]
+        if use_sector:
+            report_fields.insert(1, "sector_organisation.sector_id")
+        report_options = Storage(rows = report_fields,
+                                 cols = report_fields,
+                                 fact = ["count(id)",
+                                         "list(name)",
+                                         ],
+                                 defaults=Storage(rows = "country",
+                                                  cols = "organisation_organisation_type.organisation_type_id",
+                                                  fact = "count(id)",
+                                                  totals = True,
+                                                  chart = "spectrum:cols",
+                                                  #table = "collapse",
+                                                  ),
+                                 )
 
         location_context = settings.get_org_organisation_location_context()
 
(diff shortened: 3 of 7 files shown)


modules/templates/default/config.py
```diff
@@ -503,6 +503,8 @@ def config(settings):
     # Organisations
     # Uncomment to use an Autocomplete for Organisation lookup fields
     #settings.org.autocomplete = True
+    # Enable the Organisation Sector field
+    #settings.org.branches = True
```
nursix: Still saying org.branches not org.sector

controllers/asset.py
```diff
@@ -146,8 +146,6 @@ def supplier():
     filter_widgets = s3db.get_config("org_organisation", "filter_widgets")
     # Remove type (always 'Supplier')
     filter_widgets.pop(1)
-    # Remove sector (not relevant)
-    filter_widgets.pop(1)
```
nursix: I don't think the setting alone makes the sector suddenly relevant in this view

controllers/inv.py
```diff
@@ -457,8 +457,6 @@ def supplier():
     filter_widgets = s3db.get_config("org_organisation", "filter_widgets")
     # Remove type (always 'Supplier')
     filter_widgets.pop(1)
-    # Remove sector (not relevant)
-    filter_widgets.pop(1)
```
nursix: Same here - the use_sector setting alone doesn't quite make sector relevant in this view

modules/s3db/org.py
```diff
@@ -498,18 +520,32 @@ def model(self):
...
+                         "country",
+                         ]
+        if use_sector:
+            report_fields.insert(1, "sector_organisation.sector_id")
+        report_options = Storage(rows = report_fields,
+                                 cols = report_fields,
+                                 fact = ["count(id)",
+                                         "list(name)",
+                                         ],
+                                 defaults=Storage(rows = "organisation_organisation_type.organisation_type_id",
```
nursix: Maybe better the other way around (because more countries than types)?

controllers/asset.py
michaelhowden: +1

controllers/inv.py
michaelhowden: +1

nursix:
I do have my concerns about making summary pages default - it has performance implications which are not generally acceptable, and it is not easy to reduce it (always easier to inflate than to deflate the default).

modules/s3db/org.py
michaelhowden: Hmmm - but the Type is probably more interesting for the chart?
Switch them then do spectrum:cols?

modules/templates/default/config.py
michaelhowden: D'oh - thanks

michaelhowden:
Is that a block toward making the summary pages default? Or just a concern? 
Where are the performance constraints / difference vs. a filter list page?

nursix:
NB Making summaries the module index page by default also leads to inconsistencies where the template does not use summaries elsewhere - and with the List-button in the create-view.

I don't think that this is the right approach - it would be better to make the no-method default view in CRUD configurable to ensure consistency.

nursix:
A list-filter page has about a third of the volume of a default summary page - both in terms of back-end processing, DOM nodes to load/render and JS to run. This does clearly have impact on performance, and if there is no need for a summary page, then why render one?

michaelhowden:
OK - is it possible to configure the no method -> summary pre-template?

michaelhowden:
However I'm curious that the summary page is 3x the load of the list-filter, when the the only difference is the tabs(?). Would it be possible to load more of the magic behind the tabs (reports / maps) only on demand?

nursix:
There is really no point in pre-template configurability ;) But in-template global configuration can be made possible.

Currently, the no-method default is not configurable - but if I'm given enough time I can make that work. That was planned anyway, just a question of bandwidth ;)

michaelhowden:
pER-template - we're talking about the same thing - just from opposite ends of the day! ;)
That would be ideal... currently we're not exposing the summary page for some resources... :S

Everything should be in order now...

nursix:
I wasn't actually talking about server load - but about page performance ;)

Summary pages are primarily slow client-side - the server-side part is acceptable. Client-side performance depends massively on the user device and browser - and that has a very broad range far beyond our control.

The hidden content widgets are all loading only on demand - but that too requires additional JS, and the placeholders to hook them in. And it does inflate the client-side processing massively, as does the cross-tab filtering.

NB the tabs themselves are actually /not/ loaded from the server (in fact, they are not even mandatory for summary pages) - they are one of the many things on a summary page that are JS-generated client-side. That too adds to the page complexity and makes it slow, of course.

Simple datatables is what most clients can manage easily. Summaries and Profiles require fast end-devices, and that isn't exactly the standard everywhere yet.

michaelhowden:
@nursix Are you happy with everything in this PR now?

controllers/asset.py
```diff
@@ -146,8 +146,9 @@ def supplier():
...
+    if settings.get_org_sector():
+        # Remove sector (not relevant)
+        filter_widgets.pop(1)
```
nursix: Hmm...insert(2,...) but pop(1)? Is that correct?

nursix: Yes, in that order it is correct. Very fragile, though ;)

michaelhowden: Hey - not my code! (well at least not in this PR)

controllers/inv.py
nursix: In other places, we use something like:

   filter_widgets = [fw for fw in filter_widgets 
                     if fw.field not in ("organisation_organisation_type.organisation_type_id",
                                         "sector_organisation.sector_id",
                                         )]
    s3db.configure(tablename, filter_widgets=filter_widgets)

That is certainly more complicated, but much more robust - and more obvious for debugging. Alternatively, the model could skip the insertion of the filters when current.request.controller is "inv" or "asset".

nursix: Not inserting the filters when not needed is btw. the preferred variant ;) as always. Most efficient.

michaelhowden: Sure - feel free to make that change - I'm not touching that pop in this PR - just adding the (required) conditional

controllers/asset.py
nursix: So?

You mean since you didn't write it, it's not your responsibility to check whether it works correctly with the changes you are proposing? Not quite the right attitude, I'm afraid - if you make changes, then you are of course to verify /all/ related functionality as well ;)

controllers/inv.py
nursix: I think you should make this more robust as suggested /for/ and through this PR. It's too fragile as it is now.

michaelhowden:
The code was functional (admittedly fragile) - hardly a priority when we're tryng to respond to a disaster... If you have an issue with existing code then I think that's on you to fix. 
However I've made some changes as I really want to get these changes merged so I can move on with some other work. 
How does that look?
Do consider blocking PRs like this de-incentivized people from contributing - not many active contributors right now - which I think is a bigger issue than code quality.
I do appreciate the reviews, but I would have been much happier to fix this in a subsequent PR

michaelhowden:
I'm not wanting to make those changes to the IFRC config as I know that AidIQ is managing that

nursix:
I don'ẗ think your changes here are in response to a disaster, and what I'm asking for is just to make it more robust now that you've introduced a deployment-setting to control the sector filter and hence a third variable which makes it more fragile.

No changes are required for the IFRC template, merely the asset and inv controllers shouldn't pop the widgets, but rather they not be inserted in the first place when current.request.controller == "inv" or "asset".

That's a minor enhancement to your PR and should be less than 10mins for you to do.

Please don't try to put pressure on me - I am not blocking the PR, but trying to prevent potential sources of regression problems - which are much more harmful especially in the middle of an emergency deployment than requesting you to invest 10 more minutes in your PR.

Please try to act responsibly and not use your position to apply thumb screws on gate keepers.

michaelhowden:
I've done the enhancement already! 
current.request.function != "supplier however... OK?
Can you review?

flavour:
Personally I think that code quality is *much* more important than number of code contributors...I think our users would agree too.

nursix:
I don't see any further issues.

Currently I don't have visibility of any corresponding requirements, so can't validate whether these changes are appropriate. But they seem technically ok.

modules/s3db/org.py
```diff
@@ -437,6 +446,9 @@ def model(self):
                        "acronym",
                        "comments",
                        ]
+        if use_sector:
+            text_fields.append("sector.name")
+            text_fields.append("sector.abrv")
```
flavour: Minor enhancement (not blocking, just for next time)
text_fields += ["sector.name", "sector.abrv"]

modules/s3db/org.py
```diff
@@ -490,26 +502,53 @@ def model(self):
...
-                         ),
-            # NB Order is important here - gets popped in asset & inv controllers & IFRC template
-            type_filter,
-            # NB Order is important here - gets popped in asset & inv controllers & IFRC template
-            S3OptionsFilter("sector_organisation.sector_id",
-                            options = lambda: \
-                                get_s3_filter_opts("org_sector",
-                                                   location_filter=True,
-                                                   none=True,
-                                                   translate=True),
```
flavour: You're deliberately removing this filter?
In general that's good as the core function does some, but does it handle the none/location_filter options?
I certainly doubt the latter...

modules/s3db/org.py
```diff
@@ -4430,7 +4481,7 @@ def org_organisation_logo(id,
                    _height=60,
                    )
         return logo
-    return DIV() # no logo so return an empty div
+    return ""
```
flavour: Why prefer "" to an empty DIV?

modules/s3db/org.py
michaelhowden: Nope - not sure how that got missed - added back in

modules/s3db/org.py
michaelhowden: I actually thought that was bad practice... 
thanks for not blocking - have added it in with the other change

modules/s3db/org.py
michaelhowden: Cleaner HTML, less nodes, and the styling was causing double borders (or similar weirdness - it was a while ago)

michaelhowden:
Changes updated...

michaelhowden:
OK to merge?

michaelhowden:
@flavour Would make my git management much easier to get this merged

PR #1084, opened 2015-05-02
Minor Homepage enhancements for Nepal

michaelhowden:


controllers/inv.py
```diff
@@ -1009,6 +1009,11 @@ def set_track_attr(status):
             tracktable.recv_bin.writable = True
 
     def prep(r):
+        if r.vars.get("recv.status") == '2':
+            s3.crud_strings.inv_recv.title_list = T("Existing Shipments to Received")
```
flavour: "Existing Shipments to Receive"?

I am wary of making this change as it means new strings to translate suddenly.
Especially wary just after Honduras Red Cross spent a long time going through this module & tweaking...perhaps this could go into template?

controllers/inv.py
```diff
@@ -1009,6 +1009,11 @@ def set_track_attr(status):
...
+        if r.vars.get("recv.status__ne") == '2':
```
flavour: " not ' ;)
And one above too

modules/s3db/gis.py
```diff
@@ -320,6 +320,7 @@ def model(self):
                                       widget = S3LocationSelector(show_address=True,
                                                                   show_map=settings.get_gis_map_selector(),
                                                                   show_postcode=settings.get_gis_postcode_selector(),
+                                                                  levels=["L1", "L2", "L3", "L4"], #@ToDo: Fix
```
flavour: This can't be merged, sorry.
The default is to let it auto-detect, which works generally as far as I know...is there a case where it fails?

modules/s3db/inv.py
```diff
@@ -2142,7 +2150,8 @@ def inv_send_controller(cls):
         s3 = response.s3
 
         # Limit site_id to sites the user has permissions for
-        error_msg = T("You do not have permission for any facility to send a shipment.")
+        error_msg = T("You do not have permission for any %(site)s to send a shipment.") % \
+            dict(site=current.deployment_settings.get_inv_facility_label)
```
flavour: Note I encountered a few crashes with this kind of syntax...something about pickling.
e.g.:
https://github.com/flavour/eden/blob/master/modules/s3db/req.py#L1290

This is perhaps something that can be fixed upstream in Web2Py since I guess this is a Web2Py change which has meant this stopped working...but am wary of merging more of these until we can track down the issue.

modules/s3db/inv.py
```diff
@@ -2534,7 +2561,7 @@ def inv_send_process():
                                 site_id = send_record.to_site_id,
                                 comments = send_record.comments,
                                 status = SHIP_STATUS_SENT,
-                                type = 1, # 1:"Another Inventory"
+                                type = 11, # 11:"Another Inventory"
```
flavour: hmm, this seems like we're hardcoding somethign which is configurable? Always nasty, but can be mitigated with lots of warnings in code about other locations this is used.

modules/s3db/inv.py
```diff
@@ -1095,6 +1095,8 @@ def model(self):
                                                   not_filter_opts = (True,),
                                                   )),
                            ),
+                     self.gis_location_id(label = T("To Location"),
+                                          ),
```
flavour: This is certainly something that we'd like to be able to do...before this need to create a generic facility for the destination if we're doing distributions, which is ugly...being able to pick a village is a good idea.
However I do worry about rushing this through - it needs careful validation to make any changes in this area.

modules/s3db/inv.py
```diff
@@ -2406,6 +2415,24 @@ def prep(r):
...
+    $("#inv_send_to_site_id__row").show();
+    $("[id^='inv_send_location_id']").hide();
+  } else {
+    $("#inv_send_to_site_id__row").hide();
+    $("[id^='inv_send_location_id']").show();
+  }
+  
+});
+"""
+                    s3.jquery_ready.append(script)
```
flavour: Better in static, of course

modules/s3db/supply.py
```diff
@@ -434,7 +434,7 @@ def model(self):
                                  supply_item_represent,
                                  sort=True),
             sortby = "name",
-            widget = S3AutocompleteWidget("supply", "item"),
+            #widget = S3AutocompleteWidget("supply", "item"),
             comment=S3AddResourceLink(c="supply",
```
flavour: This should be a deployment_setting.
We have similar already, such as:
https://github.com/flavour/eden/blob/master/modules/s3cfg.py#L2803

modules/templates/Nepal/config.py
```diff
@@ -23,7 +23,7 @@ def config(settings):
     T = current.T
 
     settings.base.system_name = T("Nepal Sahana Disaster Management Platform")
-    #settings.base.system_name_short = T("Sahana")
+    settings.base.system_name_short = T("Sahana")
```
flavour: I set the default to this now :)

modules/s3db/inv.py
nursix: Generally not so good as it needs to create a new location for the street address every time you sent a shipment there - much smarter to indeed create a location-only destination facility, which has a site_id and can be re-used.

This would also make it easier to track all shipments to the same location (e.g. for reporting).

modules/templates/Nepal/tasks.cfg
```diff
@@ -43,3 +43,14 @@ org,office_type,default/office_type.csv,office_type.xsl
...
+# Inventory
+inv,inv_item,inv_item.csv,inv_item.xsl
+# Requests
+req,req,test/inv-mngt/req_req.csv,req.xsl
+req,req_item,test/inv-mngt/req_item.csv,req_item.xsl
+# The order in which the items are shipped is important
+# so do not try to change the order or merge the following track_item files
+inv,track_item,test/inv-mngt/track_item1.csv,track_item.xsl
+inv,track_item,test/inv-mngt/track_item2.csv,track_item.xsl
+inv,track_item,test/inv-mngt/track_item3.csv,track_item.xsl
```
flavour: These should be in a Demo/ subfolder as we don't want this in a Prod instance do we?

modules/s3db/inv.py
nursix: Shouldn't that be:

    dict(site=current.deployment_settings.get_inv_facility_label())

i.e. call it instead of passing in the function?

static/fonts/font-awesome/css/font-awesome.css
```diff
@@ -0,0 +1,1801 @@
+/*!
+ *  Font Awesome 4.3.0 by @davegandy - http://fontawesome.io - @fontawesome
```
flavour: What's the reason to start using the newer Font Awesome?
I know it's a good idea generally, as the older release is unsupported, but what is the driver?

modules/s3db/inv.py
nursix: This does create a strong layout-/theme-dependency here - shouldn't that be custom?

Anyway - I'm not convinced of location_id in inv_send yet.

modules/templates/Nepal/config.py
```diff
@@ -128,6 +128,108 @@ def config(settings):
...
+                  "buttons": [{"args": "create",
+                               "icon": "square",
+                               "label": T("Your Site"), #@ToDo: look up name of facility
+                               },
+                              {"args": "list",
+                               "icon": "th",
+                               "label": T("All"),
+                               }]
+                  },
+                 {"title": T("Warehouses"),
```
nursix: It will be useful to use S3Navigation here (not at the moment, perhaps - but in general). Also, ICON() is better than hardcoding i-tags.

modules/templates/Nepal/config.py
```diff
@@ -272,10 +374,10 @@ def config(settings):
             name_nice = T("Hospitals"),
             #description = "Helps to monitor status of hospitals",
             restricted = True,
-            module_type = 10
+            module_type = 3
         )),
         ("patient", Storage(
-            name_nice = T("Patient Tracking"),
+            name_nice = T("Patient"),
```
nursix: Patients?

modules/templates/Nepal/views/index.html
```diff
@@ -8,18 +8,27 @@
...
+  {{i+=1}}
+  {{if i>4:}}
+   {{i=0}}
+    </div>
+    <div id='home_menu' class='row'>
+  {{pass}}
   <div class='large-3 columns'>
     <a href='{{=URL(c=menu["module"], f=menu["function"], args=menu.get("args",None))}}'>
-     <h2><i class='icon-{{=menu["icon"]}} menu'></i></h2>
+     <h2><i class='fa fa-{{=menu["icon"]}} menu'></i></h2>
```
nursix: {{=ICON(menu["icon"])}} maybe?

modules/templates/default/css.cfg
```diff
@@ -4,7 +4,7 @@
 ../themes/default/widgets.css
 ../themes/default/shortcut.css
 ../themes/default/homepage.css
-bootstrap/font-awesome.css
+../fonts/font-awesome/css/font-awesome.css
```
nursix: ...is breaking default

static/themes/CERT/homepage.css
```diff
@@ -2,14 +2,34 @@
...
+    background: None;
+    padding: 0;
+}
+
+.home-page .row,
+.home-page .panel {
     margin: 0 auto;
+}
+
+#menu-page .row,
```
nursix: Fixed width for foundation? There's a better way to do that which doesn't interfere with the grid

nursix:
I don't see why we need the SCSS of font-awesome in trunk? We're not going to customize it, are we?
I think the CSS is absolutely good enough

static/fonts/font-awesome/less/animated.less
```diff
@@ -0,0 +1,34 @@
+// Animated Icons
```
flavour: We shouldn't really have all this scss in the fonts folder I don't think

static/fonts/font-awesome/css/font-awesome.css
flavour: & css should be in the CSS folder not in fonts folder

modules/templates/default/css.cfg
nursix: If you want to switch to FA4, then do switch - not just change the CSS.

Just changing the CSS breaks all ICON tags - you also need to fix that.

static/fonts/font-awesome/css/font-awesome.css
nursix: @flavour: many more icons available (which in itself is both good and bad), but in any case it has more semantic icons which is why I was wanting to switch to FA4 too.

However, I don't see a need to include all the SCSS sources, we're not going to re-compile FA anyway.

But /if/ we switch, we need to activate FA4 in ICON, and change the default icon set in S3Config - otherwise this change breaks all ICON tags all over the place.

modules/s3db/inv.py
flavour: I think the point is for Distributions - where there isn't a Street Address or anything and certainly no Organisation.
This is for Lx, typically Villages...where the Location Selector does work (although should have the widget customised here to NOT include a Street Address/Postcode or a Map...it should just have the Lx dropdowns...or there should be an investment in the LocationSelector to be able to pick existing locations if we do need Street Addresses (NB This was a feature of the original LocationSelector but was de-scoped from the new one)

modules/s3db/inv.py
flavour: Yes, this is very hardcoded to a specific theme indeed :/

static/fonts/font-awesome/css/font-awesome.css
flavour: My understanding was that we could have the 2 co-exist: FA3 for legacy themes & FA4 for newer themes...although if this is a clear superset of FA3 then we can migrate the core ICON & hope there's not too much hard-coded FA3 not wrapped by ICON...which I think is probably too hopeful at this stage?

modules/s3db/inv.py
nursix: You can actually inline a facility here - that way the user would perceive this as sending to a location, and yet the same address can be referenced again without creating a new location every time (which you do when you enter a street address).

I would do this via a link table inv_send <=> org_facility, and only expose the location selector of the facility, using the address as name and populating site_id automatically. Next time you need to send to the same address, you can pick it from the facility list (site_id).

Then you can track/report shipments by site_id - which isn't possible when you enter the street address and hence create a new location every time.

static/fonts/font-awesome/css/font-awesome.css
nursix: Yes, we can have both co-exist, but the default template uses FA3, so changing the CSS without switching the theme to FA4 breaks all ICON instances.

S3Config returns FA3 by default, which must be changed - and ICON does have FA4 icons commented, which needs to be un-commented if we switch /any/ theme to FA4.

Else both sets can easily co-exist. My complaint was merely about changing the css.cfg but not enabling FA4 for the theme and in ICON.

nursix: Note that when we make FA4 the default, then all themes which still include FA3 CSS must be explicitly set to FA3 (or have their css.cfg adapted, which is possible if they have no hardcoded i-tags).

nursix: There are no hardcoded I-tags in core anymore - only in templates. So, switching is relatively easy - you just need to do the whole job not just half of it.

1) Change default/css.cfg to include FA4 instead of FA3
2) Un-comment the FA4 icons in s3widgets/ICON
3) Change the default for get_ui_icons to FA4 (currently FA3)
4) Make sure that all /templates/ which still have hardcoded FA3 I-tags do define FA3 as their icon set (settings.ui.icons="font-awesome")

nursix: Recommend to *not* introduce any new hardcoded I-tags, but rather use ICON - much easier to switch the ICON theme then.

modules/s3db/inv.py
nursix: I get the idea - but then sending to an Lx without street address may work in Nepal, and maybe in Sweden :D but it may be very hard for logistics in places like LA, Berlin or Paris?

Indeed - if the location selector could select existing addresses at the Lx, then that would be a solution, although  you still wouldn't have anything in gis_location to manage the reception - and you are operating outside of Auth with locations.

So, maybe as a deployment option, but not by default?

flavour: Locations are not meant to replace Sites, just to supplement them...since not all destinations are Sites

nursix: And I wonder whether selecting a location shouldn't enforce "Distribution" - for integrity reasons?

nursix: Yeah - I know that this isn't replacement, but an alternative end-point.

And yet, in major cities the same Lx may have multiple distribution end points (=drop-off points), and they may receive more than one shipment each.

Now, I think that a drop-off point is a type of site, not just a naked location - even if it can, in simple cases, just be an Lx without address.

Modelling drop-off points as sites makes things more consistent - and there is then no need to implement site_id/location_id case discriminations for reports and filters (but instead can filter by site type, much easier).

Doesn't mean that this would need to be exposed to the user - but in the interest of consistency and simplicity/robustness of the code I think that all distribution end-points should be modelled as sites as well.

nursix: And that could then be one selector (S3SiteSelector), analogous to S3LocationSelector - where you can either choose an existing site, or create a new one inline - which can indeed be both, a named facility, or a location-only drop-off point.

Functionality would be quite similar to S3LocationSelector, except that it has the ability to choose existing drop-off points besides named facilities.

Receipts may then also cite the drop-off site (without though the items being tracked further after that point), which is probably a good idea if we want to track not only whether something has gone out, but also that it has arrived at the intended location.

Tracking drop-off points may btw also be interesting for the public - even if they are not treated as facilities.

nursix: You could create distribution end-points as inv_distribution and make that a site-instance (and an OU of the sending organisation). This has obviously advantages for multi-tenancy, where one org doesn't want (or isn't supposed to see) the distribution end points of another (if you now were to re-use existing end points, or report shipments by end point which may be more interesting in this context).

flavour: The fundamental point of org_site is a link table between gis_location & org_organisation.
These distribution points do NOT have an org_organisation.
That doesn't completely rule out using Sites but does make them a little odd.
I understand the point of consistency & I don't like the 'Location or Site' "wizard" of a radio button: poor for the user & often very fragile for the devs.
What I'd prefer to see here is the use of Location & have the LocationSelector able to select existing sites...but this is far from trivial & GPF want a solution now.
NB Honduras RC didn't see it as /too/ clumsy to simply create an org_facility for distribution end-points...so this /could/ work here too until we can resource getting a proper solution in-place...I'm certainly very wary about such quick hacks...

nursix: Umh, yes - distribution end-points kinda do have an organisation, which is the sending organisation (or the org conducting the distribution).

That may be irrelevant for this deployment, but multi-tenancy deployments may wish to not share their drop-off points, especially when that poses a security risk.

However, that wasn't quite why I proposed the site super-entity, but rather to not complicate the code that is to process the data. If we have a dual location_id in this model (one in inv_send and one in org_site), then it will become much harder to aggregate the data per location (or Lx), e.g. to analyse how much has been delivered to a certain L2. Or to filter shipments by destination Lx.

Such complications, apart from GUI, make the whole thing more fragile, and require a lot more coding around this inconsistency. I see how using org_site is bending the ontology a little bit, but then we also need to think practically. If any end-point is always a site - then there is only one possible link to the destination location_id, and thus accessing it for analysis, aggregation and reporting becomes a whole lot easier.

The destination org is btw separate anyway - so for this particular model the org_site link to organisation isn't relevant anyway. For inv_send, the site_id is primarily a link to a location with a *type* (encoded as instance-type), and it is the type that distinguishes named facilities and naked drop-off locations as end-points, not the organisation.

flavour: The alternative for unified analysis would be to write location_id back to inv_send onaccept from the site...sure this isn't ideal from an integrity perspective.

What does the user want to see?
* If selecting a village they just want to see the Lx dropdowns and don't want any knowledge of the fact that a site is being created in the background (I prefer supply_distribution_site for this than org_facility with a type since filtering out types is always painful)
* If creating a new facility, they basically want to use the Locationselector & have the site created invisibly in the background (the original LocationSelectorWidget had an extension to do this). The one part that might be visible to them would be a radio or dropdown to select the type: Office, Warehouse, Hospital, etc (would need to be auto-adaptive &/or configurable based on list of enabled site types merged with facility types)
* If selecting an existing site, I'm thinking we'd do this like the AddPersonWidget:
They have a Name field which does an AC lookup to select existing or 'None of these' to create a new one.
They /might/ wish to have the list filtered by Lx or Site Type if it's a large list, but this seems to be a future extension to worry about that.

Whilst this feels right longer-term, I think the best short-term option is to simply do what Honduras RC accepted: create an org_facility for each distribution site....slightly clumsy but 100% reliable & working right now.

nursix: Sounds right to me :)

controllers/inv.py
michaelhowden: +1

modules/s3db/gis.py
michaelhowden: Of course not! It's a quick hack. It does appear that the auto-detect is broken however... Will investigate.

modules/s3db/inv.py
michaelhowden: Thanks for the review.
I remember considering site_id vs. location_id when I first developed this code - and I must have agreed with your approach then :)
I'm presuming that a S3SiteSelector is no small task?
I'll remove the location_id, but make sure that there is an "Create Facility" link (missing now - is this not a pain for HRC?)

modules/s3db/inv.py
michaelhowden: Of Course better in Static - but when playing with ideas fast and rough is better. Given above discussion, this will be removed!

modules/s3db/supply.py
michaelhowden: +1

modules/templates/Nepal/config.py
michaelhowden: :)

modules/templates/Nepal/tasks.cfg
michaelhowden: +1

static/fonts/font-awesome/less/animated.less
michaelhowden: +1

controllers/inv.py
michaelhowden: Although I'm wary about any one template dictating default... Moved to the template for now - but I think that these changes could have generally improved UX with inv

modules/s3db/gis.py
michaelhowden: Hmm... appears to be working now...

modules/s3db/inv.py
michaelhowden: Removed dict and fixed call

modules/s3db/inv.py
michaelhowden: Yes - this is nothing new...
However we were hard coding it wrong - so this has been fixed. 
No time to go through and change all of these now.

static/fonts/font-awesome/css/font-awesome.css
michaelhowden: Both - thanks for the review :) (And correct - FA4 was needed for new icons)
Dominic - thanks for the guidelines - I will do my best to action this (although not this week).
Some questions:
* Can you confirm - delete FA3?
* Can you give me some examples of Icons in Core?
* I would rather keep all the font awesome files together in the same place (under Fonts), rather than scatter them...
* Is there a way to specific a min.css to use rather than compiling the css into the eden.min.css (the relative file reference breaks when doing this - forcing me to edit the Font Awesome CSS

modules/templates/Nepal/config.py
michaelhowden: I know - It's on the list.

modules/templates/Nepal/config.py
michaelhowden: +1

static/themes/CERT/homepage.css
michaelhowden: What did you have in mind? My aim was to keep the grid - but give a max size... is there a problem with this, or just the method...?

michaelhowden:
@nursix + @flavour Just to clarify, I was under no assumptions that this code was trunk worthy - hence no PR. However your review was extremely timely. I've tried to address all of these - with the last outstanding this being the deprecation of FA3

michaelhowden:
PR Updated

static/fonts/font-awesome/css/font-awesome.css
nursix: No - completely wrong :/

1) Keep FA3 - FA3 and FA4 can co-exist, and FA3 is still required by many templates!
2) No change in core required, all cases have been migrated to ICON
3) Change default/css.cfg to include FA4 instead of FA3
4) Uncomment FA4 icons in s3widgets.py/ICON.
5) Make S3Config.get_ui_icons default to FA4 instead of FA3
6) Make sure all templates which still use FA3 specify a settings.ui.icons (they may rely on the default of the setting which would break when you change the default)

And as Fran said: CSS should be under css, and fonts under fonts

nursix:
Still a lot of .less files without need. LESS is even /less/ needed than SCSS ;)

CSS and font files should be good enough.

modules/s3/s3query.py
```diff
@@ -676,7 +676,8 @@ def __init__(self, resource, selector, label=None):
         # Fall back to the field label
         if label is None:
             fname = self.fname
-            if fname in ["L1", "L2", "L3", "L3", "L4", "L5"]:
+            # Does this really belong here? Better on gis_location fields
```
nursix: This does belong here! The labels vary by config location, so the model must be loaded before you can determine which labels to use. Please remove the comment.

static/themes/CERT/homepage.css
nursix: No problem if you do that in your theme's style.css only, not across themes.

static/fonts/font-awesome/css/font-awesome.css
nursix: Sorry about the complications - but as long as you keep introducing new hardcoded I-tags this is not going to become any easier.

ICON works this way:
- instead of I(_class="fa fa-xyz") you write just ICON("xyz")
- ICON is an HTML helper which will be rendered as <i class=...> - you can specify additional attributes with underscore as usual, like ICON("organisation", _class="hide"), it can be used in place of I(). Note that no additional whitespace is required - the horizontal spacing between icon and label is CSS now (at least in the default theme and derivates, as well as IFRC)
- ICON looks up the setting.ui.icons setting, and chooses the corresponding CSS classes for the respective icon set
- ICON has some semantic icon names pre-configured internally (rather than using the descriptive names), e.g. ICON("organisation") rather than I(_class="fa fa-sitemap"). This is recommended, because it allows you to configure icons by /what/ they symbolize rather than by /how/ they look. E.g. there are ICON("edit") and ICON("delete") - rather than "pencil" and "trash".
- In addition to the pre-configured icons, the template can specify custom icons as settings.ui.custom_icons - if you e.g. want to include some icons that are not the default set, like in MCOP which uses both FA3 but also DRMP icons, or when you want to override the default in ICON (e.g. use a different class for "organisation" than what is configured in ICON).
- icon names which are not defined anywhere are used as classes, e.g. ICON("undefined") will give <i class="fa fa-undefined"> in FA4. That way, you are not forced to configure each and every icon class - only where you want to use semantic icon names you will have to define them somewhere
- it is possible to override the HTML layout of ICON, although FA3, FA4 or FI currently all have the same layout
- provided you are only using ICON and no hardcoded I-tags, you can switch a template to a different icon set by just changing settings.ui.icons and including a different CSS. All core code (s3 and s3db) is completely migrated to ICON and will thus follow the setting - only some older templates are still using hardcoded I-tags. However, just changing the included CSS without changing the setting will break it ;)
- the theme must of course include all corresponding CSS for the icon classes. It is generally possible to include multiple (e.g. MCOP), but remember that where they use the same selectors, the order of inclusion matters!
- of course, if you are using JS or CSS in your template that relies on specific icon CSS classes, then you can not easily switch the icon set. That is why this is discouraged at least for all core JS and CSS - only the icon CSS itself should refer to icon classes

ICON is used throughout the core - in data list layouts, profile pages, and some form widgets. Since this code is used across many templates, no hardcoded I-tags should be introduced in core anymore.

Changing a theme's icon set is easy - but changing the *default* requires more care. The majority of templates currently uses FA3 - but implicitly, relying on the default. Where there are hardcoded I-tags in the template, they can not easily be "switched" to FA4. However, if you don't have the time to migrate all hardcoded I-tags, you can just configure these templates to continue using FA3 for now - via setting.ui.icons. Obviously, you can then not "remove" FA3 just yet.

nursix:
Interesting that you say there was no PR - obviously this here is a PR?

static/themes/CERT/homepage.css
nursix: Bit odd however that you change CERT/style.css while it is not used for the Nepal template - and odd also to set a fixed width again while we're struggling to much with fixed widths. Generally, I'm very sceptical - and don't quite see what this would improve.

nursix: Also note that the PR retains the default/eden.min.css even though you removed the changes in its source now. Should remove the minified file as well.

michaelhowden: 1) Nepal & CERT Homepages use the same css: https://github.com/flavour/eden/blob/master/modules/templates/Nepal/controllers.py#L19
2) +1. Although planning to change Default to use FA4 - any complaints?

michaelhowden: "Although planning to change Default to use FA4 - any complaints?"
Just read our comment:
"Changing a theme's icon set is easy - but changing the default requires more care. The majority of templates currently uses FA3 - but implicitly, relying on the default. Where there are hardcoded I-tags in the template, they can not easily be "switched" to FA4. However, if you don't have the time to migrate all hardcoded I-tags, you can just configure these templates to continue using FA3 for now - via setting.ui.icons."
There should be no need to switch default to FA4 (yet), just switch setting.ui.icons = to FA4 in Nepal... I'll dig into that...

michaelhowden:
@nursix Changes made - how's this look?

controllers/default.py
```diff
@@ -63,8 +63,13 @@ def register_validation(form):
 
 # =============================================================================
 def index():
-    """ Main Home Page """
-
+    """ Customized Main Home Page (in modules/<TEMPLATE>/config.py) """
```
nursix: I'm -1 to introducing this - we already have a (much better) method to customize the home page controller, and this introduces double standards, which is confusing.

But I can see why you did this - you wanted to have access to the IM layout definition.

However, that's not necessary - templates are now just normal modules, so you can move everything into the right place:
- move the IM layout into layouts.py
- move the inv index controller into controllers.py

...and import:
- import IM into controllers.py like: from layouts import IM
- import the inv index controller into config.py and import it into config.py like: from controllers import customise_inv_homepage

Alternatively, you can keep the customise_inv_homepage in config.py, and import the IM layout there.

You can even import across templates, e.g. a layout like: from templates.Nepal.layouts import IM

...so there is less need to duplicate code across templates.

modules/s3/s3widgets.py
```diff
@@ -7911,75 +7911,89 @@ class ICON(I):
...
-            #"twitter": "fa-twitter",
-            #"unsent": "fa-times",
-            #"up": "fa-caret-up",
-            #"upload": "fa-upload",
-            #"user": "fa-user",
-            #"zoomin": "fa-zoomin",
-            #"zoomout": "fa-zoomout",
-        #},
+        "font-awesome4": {
+            "_base": "fa",
```
nursix: I see that you changed/added some keys here. If you do that, please make sure the same keys are available in every icon set - or, if that is not possible, move the additional keys into settings.ui.custom_icons (same dict structure).

All icon sets should have a consistent set of default icons.

modules/s3db/inv.py
```diff
@@ -1050,6 +1050,9 @@ def model(self):
         #send_type_opts.update(self.inv_item_status_opts)
         send_type_opts.update(settings.get_inv_send_types())
 
+        self.org_facility # Load model + crud_strings
```
nursix: Really? A cross-model dependency just for a CRUD string? That looks ugly to me.

nursix: I think it's better to replicate this CRUD string here - the implications of loading another model in a model are too heavy, especially if that happens unconditionally.

nursix: I mean - it's not technically wrong, but it is totally disproportionate for just one CRUD string. It's not worth it.

modules/s3db/supply.py
```diff
@@ -425,6 +425,10 @@ def model(self):
         supply_item_represent = supply_ItemRepresent(show_link=True)
 
         # Reusable Field
+        if settings.get_supply_autocomplete():
+            default_widget =  S3AutocompleteWidget("supply", "item"),
```
nursix: Wrong syntax (trailing comma)

static/themes/Nepal/index.css
```diff
@@ -0,0 +1,48 @@
...
+
+div#index-page h1,
+div#index-page h2 {
+    font-weight: inherit;
+    background: None;
+    padding: 0;
+}
+
+#index-menu h2 i {
+    font-size: 64px;
```
nursix: Absolute font-size? Maybe you missed that - the others you used rem for (which is appropriate)

nursix:
Sorry for the delay with the review, couldn't find the time earlier. Please note that there are merge conflicts, so rebasing may be necessary.

controllers/default.py
michaelhowden: Hmmm... (pondering)
I was expecting some debate around this - and generally I'm OK with this approach!
Some comments:
* One main reason to move it into config.py is to avoid restarting web2py for every change!
* We already have double standards with the "customise_home" setting... should that have been introduced?
* I actually see a strong reason for putting more in controllers.py than config.py so that it can be reused between templates - eg, let's say that the customise_org_organization_controller needed to be reused in multiple templates... is this right? (Not something I'm planning to do right away)

modules/s3db/inv.py
michaelhowden: OK

controllers/default.py
michaelhowden: Thanks for the import pointers - took <5min to change :)

modules/s3/s3widgets.py
michaelhowden: Moved to settings.ui.custom_icons

static/themes/Nepal/index.css
michaelhowden: Changes

michaelhowden:
@nursix Thanks for the review. Everything addressed and merge conflicts resolved (nothing major) - hopefully good to merge!

modules/s3/s3codecs/xls.py
```diff
@@ -171,6 +171,9 @@ def encode(self, data_source, **attr):
         list_fields = attr.get("list_fields")
         if not list_fields:
             list_fields = data_source.list_fields()
+        # Always export comments field
+        if "comments" in data_source.fields and "comments" not in list_fields:
```
nursix: Clear -1 to this - this should definitely not be enforced in the codec, and especially not without any option to turn it off. Exporting comments is not a general requirement for every deployment, and not for every resource either.

modules/templates/IFRC/config.py
```diff
@@ -483,78 +483,8 @@ def currency_default(default):
...
-    settings.req.req_type = ["Stock"]
-    settings.req.use_commit = False
-    # Should Requests ask whether Transportation is required?
-    settings.req.ask_transport = True
-    settings.req.pack_values = False
-    # Disable Request Matching as we don't want users making requests to see what stock is available
-    #settings.req.prompt_match = False # HNRC
-    # Uncomment to disable Recurring Request
-    #settings.req.recurring = False # HNRC
-
```
nursix: Why would you remove these settings?

modules/templates/Nepal/controllers.py
```diff
@@ -2,6 +2,7 @@
 
 from gluon import *
 from s3 import S3CustomController
+from layouts import IndexMenuLayOut
```
nursix: NB PEP8 recommends CamelCase for word separation, but Lay and Out are not two words ;) such fanciness is a source for bugs and should be avoided.

nursix:
There's a lot of non-standard code formatting here still, like blank lines missing where they should be, inconsistent identation (closing brackets), non-standard line breaks, extra whitespace ... If Fran wants to clean it up after merge, then fine - but I actually think you should do that yourself.

Enforcing the comments field in the XLS codec is plain wrong - especially without any option to turn it off. And neither can I see why you would remove some of the settings from IFRC/config.py.

Regarding the homepage customization patterns - there is a significant difference between the default/index customization and the module index pages: the default/index pattern allows for a multi-page extension, which is very useful for e.g. embedding online documentation. This pattern is by far more powerful and cleaner in view of error handling and forwarding than the customise_xxx_homepage pattern.

I am not sure whether the customise_xxx_homepage pattern should have been introduced - to me it always looked like a hack, and not very thought-through. The idea seems ok, but the design is questionable. Especially it does not take into account that errors redirect to the module index (and sometimes even swallows the errors), nor does it consider the implications on permissions. It is also loading functions without need - which is certainly the ugliest part of it (whereas the default/index pattern only ever loads the functions when you actually access default/index). I see this as a rather weak pattern that needs a lot more thought before being applied everywhere.

Regarding the re-usability of code across templates: you are /not/ limited to controllers.py in any way. As I said: templates are now regular Python packages, you can have as many modules inside it as you like. You can import from templates.XYZ.config just as much as from templates.XYZ.controllers or templates.XYZ.layouts.

However, there are of course "moral" limits - some things don't make sense. For example, importing from a template into core would be stupid, so I would be strict about not allowing that. Nor would I actually want things to get imported from one config.py into another (because that'd pave the road for regression bugs).

Generally, we should *not* strive after maximising cross-template code sharing - if a certain function or class is relevant for many templates, then it should be core. However, there are sometimes two templates within the same general use-case (e.g. RMS and RMS Americas) - and even if they did share certain elements, these elements would still be specific to the case and not necessarily generic. 

For such cases (derivates and alternatives), it make maintenance a lot easier to import from one template to the other, and helps to maintain consistency.

However, I would suggest to move this debate to a more visible forum - it's a bit misplaced on your PR.

Note that your PR may take forever if you keep introducing new things into every review iteration - better you just fix/cleanup what you already have in it, and leave the next changes to the next PR. Also note that the more different/independent changes you put into the same PR, the harder it will become to get any of them merged (one-change-blocking-the-other effect).

controllers/default.py
nursix: måndagen den 1 juni 2015 14.50.34 skrev  Michael Howden:
> > @@ -63,8 +63,13 @@ def register_validation(form):
> >  #
> >  ========================================================================
> >  =====> 
> >  def index():
> > -    """ Main Home Page """
> > -
> > +    """ Customized Main Home Page (in modules/<TEMPLATE>/config.py) """
> 
> Hmmm... (pondering)
> I was expecting some debate around this - and generally I'm OK with this
> approach! Some comments:
> * One main reason to move it into config.py is to avoid restarting web2py
> for every change! 
This is out-of-date information.

All of the template is now imported, so it doesn't make any difference whether 
it's config.py or controllers.py. As long as you have module tracking enabled 
(debug mode), there's no need to restart web2py.

However, note that git repository dates may be older than your .pyc's so that 
applying changes with git (e.g. switching branches) may however require a 
web2py restart.

Also, Eclipse does sometimes not follow these re-imports - even if web2py does 
(that's hearsay, though - I'm not using Eclipse).

Generally, one should be more relaxed about having to restart web2py - if run 
from the command line, it's just two key strokes away, and not really 
disruptive for the flow. We can not seriously make code design decisions 
dependend on such.

michaelhowden:
I've cleaned things up and addressed all comments. If you let me know where there are specific formatting issues I'll address these too.

michaelhowden:
@flavour Any delay in merging this?

PR #1086, opened 2015-05-15
Only import people with a First, Middle or Last Name 

michaelhowden:
(because only horse can have no name!)


nursix:
-1 What's this for? The name rules are enforced by the validator upon import - if you skip faulty rows already in the stylesheet, then you will get no error message and thus they go undetected. I'm relatively strongly against this practise - it is necessary to indicate errors like this, otherwise users will struggle to understand why certain data don't get imported.

PR #1088, opened 2015-05-25
Bug fix 349: handled issue of missing test directory at location of s…

mattmarquard:
…tatic_path

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -1092,14 +1092,17 @@ def result():
 
     file_list = UL()
     static_path = os.path.join(request.folder, "static", "test")
-    for filename in os.listdir(static_path):
-        link = A(filename,
-                 _href = URL(c = "static",
-                             f = "test",
-                             args = [filename]
-                             )
-                 )
-        file_list.append(link)
+    try:
+        for filename in os.listdir(static_path):
+            link = A(filename,
+                    _href = URL(c = "static",
+                                f = "test",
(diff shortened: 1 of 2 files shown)


nursix:
Good idea.

However, better would be to disable the menu item if the directory doesn't exist, or at least inform the user that we're not showing anything because there isn't anything to show. As you have it here, it would be just empty, and not obvious for the user why (<=which you can be sure will be reported as a bug).

At least, try something like

    response.warning = T("No test results available")

in the except-branch.

Note that the test suite that produces the test results is deprecated and not working at the moment ;) so this "bug" is only the very tip of the iceberg.

What is really needed is something that actually produces test results - and while we don't have that, I would recommend to comment this menu item and put in a ToDo (in s3menus.py). No point to show a menu item that doesn't ever work.

nursix:
We do have a new smoke test functionality, however - it's just not wired up to show any test results inside the application. And while it isn't, there's no point to have that menu item visible for the users.

mattmarquard:
Thanks for the review @nursix . That response warning is definitely a better way to communicate to the user there are no test results. I'll do what you said regarding removing the menu item and adding a ToDo.

nursix:
Superseded by:
https://github.com/nursix/eden/commit/656045f74ef5ccfacf86cf722a58ea9683c4759b
...hence closing this one.

Thanks Matt :)

PR #1090, opened 2015-06-03
Template Title Required; Crud Strings for Template

biplovbhandari:
The title of the CAP template made as a required field
The Crud Strings for CAP Template was showing CAP Alert Crud Strings, changed that

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -271,7 +271,8 @@ def template():
             redirect(URL(c="cap", f="template", args=[_id]))
 
     def prep(r):
-        for f in ["identifier", "msg_type"]:
+        # because template incidents value is not copied to the alert
+        for f in ["identifier", "msg_type", "incidents"]:
             field = atable[f]
             field.writable = False
             field.readable = False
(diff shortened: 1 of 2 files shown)


controllers/cap.py
```diff
@@ -271,14 +271,15 @@ def template():
             redirect(URL(c="cap", f="template", args=[_id]))
 
     def prep(r):
-        for f in ["identifier", "msg_type"]:
+        # because template incidents value is not copied to the alert
```
flavour: This is a change comment not a code comment...it doesn't make sense to have it in the code

modules/s3db/cap.py
```diff
@@ -780,13 +793,13 @@ def model(self):
...
-                                    #                     name = "location",
-                                    #                     label = "",
-                                    #                     multiple = False,
-                                    #                     fields = [("", "location_id")],
-                                    #                     ),
+                                    S3SQLInlineComponent("location",
+                                                         name = "location",
+                                                         label = "",
+                                                         multiple = True,
+                                                         fields = ["location_id",
```
flavour: I think you want to keep the tuple format to avoid the extra label no?
(The comment as to why commented is indeed resolved now, so this shoudl work :) )

PR #1091, opened 2015-06-04
cap_area_location inline form and some minor tweaks

biplovbhandari:
rebase the changes into one
- cap_area_location is inline form
- crud strings for template
- incident field disabled in the template form as this is not copied to the alert derived from template
- template title is required

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -271,14 +271,14 @@ def template():
             redirect(URL(c="cap", f="template", args=[_id]))
 
     def prep(r):
-        for f in ["identifier", "msg_type"]:
+        for f in ["identifier", "msg_type", "incidents"]:
             field = atable[f]
             field.writable = False
             field.readable = False
             field.requires = None
         for f in ["status", "scope"]:
             atable[f].requires = None
-        atable.template_title.required = True
+        atable.template_title.requires = IS_NOT_EMPTY()
         atable.status.readable = atable.status.writable = False
         itable = db.cap_info
         for f in ["urgency", "certainty",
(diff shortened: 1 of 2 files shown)


modules/s3db/cap.py
```diff
@@ -414,8 +414,21 @@ def model(self):
                         method = "import_feed",
                         action = CAPImportFeed())
 
-        if crud_strings["cap_template"]:
+        if current.request.function == "template":
```
flavour: Why did you move these from the controller to the model?
What is the benefit?
In general, better to push everything unnecessary out of the model since the model is loaded for all CAP requests whereas that controller just for that function...

controllers/cap.py
```diff
@@ -271,14 +271,14 @@ def template():
             redirect(URL(c="cap", f="template", args=[_id]))
 
     def prep(r):
-        for f in ["identifier", "msg_type"]:
+        for f in ["identifier", "msg_type", "incidents"]:
```
flavour: ok, clear :)

modules/s3db/cap.py
```diff
@@ -1276,57 +1248,31 @@ def func(form):
...
-                    return
-
-            atable = db.cap_area
-            area = db(atable.id == area_id).select(atable.alert_id,
-                                                   limitby=(0, 1)).first()
+        info_id = form_vars.get("info_id", None)
+        if not info_id:
+            # Get the full record
+            item = db(table.id == _id).select(table.alert_id,
+                                                table.info_id,
```
flavour: alignment

modules/s3db/cap.py
```diff
@@ -1276,57 +1248,31 @@ def func(form):
...
-                alert_id = info.alert_id
-            except:
-                # Nothing we can do
+            if alert_id:
+                # Nothing to do
                 return
 
+        itable = db.cap_info
+        info = db(itable.id == info_id).select(itable.alert_id,
+                                                limitby=(0, 1)).first()
```
flavour: alignment

PR #1093, opened 2015-06-10
configuration of message identifier task 1.1

biplovbhandari:
true for info template fixed


flavour:
All fine other than the instance-specific settings being in the general config...comment these out & we're good to merge :)

biplovbhandari:
then we allow user to make changes on templates/<>/config.py ?

nursix:
Maybe a more plausible commit message? "Task 1.1" reference is not very conclusive outside of the PM context

flavour:
Right, good to refer to the spreadsheet:
https://docs.google.com/spreadsheets/d/1N9xUqZQhignq7QM_7CtrV4zv86sq4lMdJj81IQuch3Y/edit#gid=0

PR #1095, opened 2015-06-13
default CAP identifier, source and expire offset configuration

biplovbhandari:
default source configuration

offset expiry day configuration

is_template is true for the one derived from alert template


biplovbhandari:
@flavour Please merge this

biplovbhandari:
@flavour Please merge this

nursix:
I recommend that you explain the urgency rather than repeating your request for merge - although it seems unclear how you would be dependent on a merge?

modules/s3db/cap.py
```diff
@@ -956,6 +960,16 @@ def generate_sender():
...
         return "%s/%d" % (current.xml.domain, user_id)
+    
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def generate_source():
+        """
+            Generate a source for CAP alert
+        """
+        settings = current.deployment_settings
+        public_url = settings.get_base_public_url()
```
flavour: No need to bring settings into scope if just using ionce: this just adds memory usage

modules/s3db/cap.py
```diff
@@ -956,6 +960,16 @@ def generate_sender():
...
+        return "%s@%s" % (current.xml.domain, public_url)
```
flavour: No need for the intermediate variable public_url since only used once...better to simply use direct

modules/templates/000_config.py
```diff
@@ -90,6 +90,17 @@
...
+# CAP Settings
+# Change for different authority and organisations
+# See http://alerting.worldweather.org/ for prefix and oid
+#settings.cap.identifier_prefix = "PAGASA-DOST"
+#settings.cap.identifier_oid = "2.49.0.0.608.0"
+#settings.cap.identifier_suffix = "alert"
+# Change this for the offset period in days that the alert will be effective for
+# Expire Date = Effective Date + expire_offset
+# Default is 2 days
+#settings.cap.expire_offset = 2
```
flavour: I would move all these CAP settigns *much* lower in the file.
Pretty much all deployments will want to change email settings yet perhaps less than 1% will use CAP, so the Email ones should be much higher in the file...

biplovbhandari:
@flavour thanks for the review...I have applied the suggestions

modules/s3db/cap.py
```diff
@@ -956,6 +960,14 @@ def generate_sender():
...
 
         return "%s/%d" % (current.xml.domain, user_id)
+    
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def generate_source():
+        """
+            Generate a source for CAP alert
+        """
+        return "%s@%s" % (current.xml.domain, current.deployment_settings.get_base_public_url())
```
flavour: Try to keep within 80 char page width - the ----- is a visual guide for this.
Can have a line break after the comma

modules/templates/000_config.py
flavour: I'd still move lower: what is more likely?
Restrict to an individual countrey? (most deployments) or CAP (very few)

flavour:
Also, please always rebase/compress when fixing PRs - thanks!

PR #1098, opened 2015-06-30
SAMBRO Layout Update

jfmnet:
Updated layout.


nursix:
That looks wrong to me. 
- Why include Foundation, but then override it with non-responsive layouts and styles?
- You can not deactivate mandatory modules (gis, pr, org), and surely you should not deactivate "cap", out of all modules?

modules/templates/SAMBRO/views/layout.html
```diff
@@ -0,0 +1,204 @@
...
+    {{include "layout_iframe.html"}}
+    {{include}}
+    {{else:}}
+    <body>
+        <div id="sambro-menu">
+            {{=current.menu.main}}
+        </div>
+
+        <div id="sambro-left">
+            <div class="sambro-panel-title"><i class="fa icon-bullhorn"></i>ALERTS</div>
```
nursix: This seems wrong: "fa" class is for font-awesome 4, but "icon-bullhorn" is font-awesome 3. We also try to avoid hardcoding icon classes, and instead use the ICON helper with abstract names that indicate the meaning rather than the look.

modules/templates/SAMBRO/views/layout.html
```diff
@@ -0,0 +1,204 @@
...
+            </div>
+        </div>
+
+        <div id="sambro-body">
+            <div id="sambro-map">
+                <div class="sambro-panel-title"><h1>Sahana Alert System</h1><p>Welcome to the Sahana alert system</p></div>
+                {{=XML(map)}}
+            </div>
+
+            <div id="sambro-alert-table">
```
nursix: This should probably be on the index page only, not in the layout frame - because that way it would appear on /every/ page, while no other dynamic contents (like CRUD forms, Admin dialogs, tables etc) would ever be shown.

static/themes/SAMBRO/style.css
```diff
@@ -0,0 +1,278 @@
...
+    height: inherit;
+    -webkit-box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.08), 0px 2px 0px 0px rgba(0, 0, 0, 0.02);
+    box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.08), 0px 2px 0px 0px rgba(0, 0, 0, 0.02);
+}
+
+.top-bar-section ul li {
+    background: #888;
+}
+
+.top-bar-section li:not(.has-form) a:not(.button) {
```
nursix: This is not responsive design in any way - you're overriding foundation's dynamic positioning and grid sizes, but what's the point with a responsive CSS framework then?

static/themes/SAMBRO/style.css
```diff
@@ -0,0 +1,278 @@
...
+
+        #sambro-alert-table table .fa {
+            margin-right: 5px;
+            color: #F83;
+        }
+
+#sambro-left {
+    background-color: #FFF;
+    margin: 20px 0px 0px 20px;
+    width: 280px;
```
nursix: Hardcoded widths, floats, positions can not adapt to screen size. Suggest that you instead utilize the grid provided by Foundation (fluid grid), which is controlled by HTML classes rather than styles.

modules/templates/SAMBRO/css.cfg
```diff
@@ -0,0 +1,42 @@
+../themes/default/layout.css
+../themes/default/footer.css
+../themes/default/widgets.css
+../themes/default/shortcut.css
+../themes/default/homepage.css
```
nursix: You're including a lot of CSS for no obvious reason - should only include what you really need for your layout.

nursix:
Experimental code?

I don't think this can be merged at this stage (not functional) - and generally the design should be (seriously!) re-considered. Responsive design is de-facto standard in Sahana, and utilizing the CSS framework when available seems more than obvious.

NB the broken unit-test is in trunk, not caused by your code.

PR #1100, opened 2015-07-02
SAMBRO Home Page Update (index.html)

jfmnet:
home page update


modules/templates/SAMBRO/controllers.py
```diff
@@ -7,14 +7,63 @@
         import simplejson as json # try external module
     except:
         import gluon.contrib.simplejson as json # fallback to pure-Python module
-
```
flavour: I would avoid such whitespace changes - the existing style was the one used in all the other code

modules/templates/SAMBRO/controllers.py
```diff
@@ -7,14 +7,63 @@
...
+
+    def __call__(self):
+        response = current.response
+        s3 = response.s3
+        T = current.T
+
+        map = current.gis.show_map()
+
+        output = {}
+        output["map"] = map
```
flavour: If these are right after one another then slightly more efficient to do the 2 together:
output = dict(map=map)

modules/templates/SAMBRO/controllers.py
```diff
@@ -7,14 +7,63 @@
...
+
+        # Image Carousel
+        # Latest 4 Events and Alerts
+        from s3.s3query import FS
```
flavour: We normally prefer to just import from the main s3 module (this may have been wrong in the code you copy/pasted)

modules/templates/SAMBRO/controllers.py
```diff
@@ -7,14 +7,63 @@
...
+        s3 = response.s3
+        T = current.T
+
+        map = current.gis.show_map()
+
+        output = {}
+        output["map"] = map
+
+        # Image Carousel
+        # Latest 4 Events and Alerts
```
flavour: Note that in the code you're copying from, these are both just types of cms_post, yet in SAMBRO an alert should be a cap_alert, no?
Does SAMBRO need Events at all?

modules/templates/SAMBRO/controllers.py
```diff
@@ -320,8 +360,7 @@ def _update_subscription(self, subscription):
                                     method=subscription["method"])
             subscription_id = success
         else:
-            success = db(stable.id == subscription_id).update(
-                            pe_id=pe_id,
+            success = db(stable.id == subscription_id).update(pe_id=pe_id,
```
flavour: If you're going to tweak all this whitespace then please sort indents

modules/templates/SAMBRO/css.cfg
```diff
@@ -0,0 +1,42 @@
+../themes/default/layout.css
```
flavour: If all you need is a little extra CSS for the homepage then moving to a full theme probably isn't required (a lot more maintenance effort).
I would suggest just inserting the extra CSS into the homepage.
(via s3.stylesheets)

modules/templates/SAMBRO/views/index.html
```diff
@@ -0,0 +1,53 @@
+{{extend "layout.html"}}{{theme=response.s3.theme}}
+<link href="/eden/static/styles/../themes/SAMBRO/style.css" rel="stylesheet" />
```
flavour: Better to do this via s3.stylesheets, as I explained

modules/templates/SAMBRO/views/index.html
```diff
@@ -0,0 +1,53 @@
...
+
+<div id="sambro-body" class="row">
+    <div class="large-12 columns">
+        <div class="row">
+
+            <div id="sambro-left" class="large-2 small-3 columns">
+                <div class="sambro-panel">
+                    <h4><i class="icon icon-bullhorn"></i>{{=T("Alerts")}}</h4>
+                    <div class="sambro-list">
```
flavour: This should be a for item in list_of_items: etc no?

static/themes/SAMBRO/style.css
```diff
@@ -0,0 +1,36 @@
...
+#sambro-center {
+}
+
+.sambro-list {
+    padding: 5px 5px;
+}
+
+.sambro-panel {
+    background-color: #FFF;
+    margin: 20px 0px 0px 0px;
```
flavour: I don't think this is responsive is it?

jfmnet: This is responsive. I am using "row" and "column" in the html. The css above is for formatting only, adding some paddings and margin, doesn't have to do with layout.

nursix: margins and paddings are fine, although Foundation recommends to use rem instead of px as that would make them adapt to font scaling ;)

static/themes/SAMBRO/style.css
```diff
@@ -0,0 +1,36 @@
...
+    border: 1px solid #e7eaec;
+    border-top: 4px solid #e7eaec;
+}
+
+    .sambro-panel div {
+    }
+
+    .sambro-panel h4 {
+        border-bottom: 1px solid #DDD;
+        font-size: 16px;
```
nursix: rem is recommended as it allows scaling (mobile devices often do auto-scaling of fonts)

modules/templates/SAMBRO/controllers.py
```diff
@@ -397,4 +436,42 @@ def _update_subscription(self, subscription):
...
+    #orderby = resource.table[orderby]
+    datalist, numrows, ids = resource.datalist(fields=list_fields,
+                                               start=None,
+                                               limit=limit,
+                                               list_id=list_id,
+                                               orderby=orderby,
+                                               layout=layout)
+    if numrows == 0:
+        # Empty table or just no match?
+        from s3.s3crud import S3CRUD
```
nursix: from s3 import S3CRUD (probably copy/pasted, I know)

static/themes/SAMBRO/style.css
```diff
@@ -0,0 +1,36 @@
...
+
+.sambro-panel {
+    background-color: #FFF;
+    margin: 20px 0px 0px 0px;
+    border: 1px solid #e7eaec;
+    border-top: 4px solid #e7eaec;
+}
+
+    .sambro-panel div {
+    }
```
nursix: Empty? So better to remove?

nursix:
Much better than before ;)

static/themes/SAMBRO/style.css
jfmnet: ok... thanks.

PR #1101, opened 2015-07-03
Priority dropdown based on event

biplovbhandari:
@flavour Please merge this


modules/s3db/cap.py
```diff
@@ -487,14 +488,59 @@ def model(self):
...
         ])
 
+        # ---------------------------------------------------------------------
+        # Warning Priorities for CAP
+        
+        tablename = "cap_warning_priority"
+        define_table(tablename,
+		             Field("priority_rank", "integer",
+		                    length = 2,
+				            ),
```
flavour: indent

modules/s3db/cap.py
```diff
@@ -487,14 +488,59 @@ def model(self):
...
+                     Field("event_code",
+		                    label = T("Event Code"),
+		                    ),
```
flavour: indent: you still have tabs here which need to all be removed & replaced with spaces!

modules/s3db/cap.py
```diff
@@ -534,9 +580,24 @@ def model(self):
                                                 ),
                            widget = S3MultiSelectWidget(),
                            ), # 1 or more allowed
-                     Field("event",
-                           required = True,
-                           ),
+                     self.event_type_id(requires = 
+                                       IS_ONE_OF(
```
flavour: requires here is the same as the default isn't it? So better not to copy here.
Seems the only thing you are doing is making this mandatory which can be achieved by adding empty = False (since this is an S3ResuableField which will strip off the IS_EMPTY_OR in the default validator)

modules/s3db/cap.py
```diff
@@ -546,8 +607,11 @@ def model(self):
                            widget = S3MultiSelectWidget(),
                            ), # 0 or more allowed
                      Field("priority",
+                           represent = priority_represent,
                            requires = IS_EMPTY_OR(
-                                        IS_IN_SET(cap_info_priority_opts)
+                                        IS_ONE_OF(
+                                                db, "cap_warning_priority.id"
```
flavour: If this is to be accessed via UI, then you will need a represent in the validator too

static/formats/s3csv/cap/warning_priority.csv
```diff
@@ -0,0 +1 @@
+priority_rank,event_code,name,event_type,urgency,severity,certainty,color_code
```
flavour: For CSV Importers we generally replace things like priority_rank with the more human "Priority Rank"

biplovbhandari:
@flavour Please merge this

biplovbhandari:
@flavour Please merge this

controllers/cap.py
```diff
@@ -289,7 +289,7 @@ def prep(r):
             field.readable = False
             field.required = False
 
-        for f in ["category", "event"]:
+        for f in ["category"]:
```
flavour: Is this likely to become a list again?
If not then no need for a loop of 1...

modules/s3db/cap.py
```diff
@@ -487,14 +488,59 @@ def model(self):
...
+                           label = T("Event Code"),
+                           ),
+                     Field("name", notnull = True, length = 64,
+                           label = T("Name"),
+                           ),
+                     Field("event_type",
+                           label = T("Event Type"),
+                           ),
+                     Field("urgency",
+                           required = True,
```
flavour: I don't think required = True is needed here?
requires = IS_IN_SET already achieves this

modules/s3db/cap.py
```diff
@@ -487,14 +488,59 @@ def model(self):
...
+                           requires = IS_IN_SET(cap_info_certainty_opts),
+                           ),
+                     Field("color_code",
+                           label = T("Color Code"),
+                           ),
+                     *s3_meta_fields())
+        
+        priority_represent = S3Represent(lookup = tablename,
+                                         fields = ["name"])
```
flavour: No need to include the fields argument if it's the default

modules/s3db/cap.py
```diff
@@ -534,9 +580,24 @@ def model(self):
                                                 ),
                            widget = S3MultiSelectWidget(),
                            ), # 1 or more allowed
-                     Field("event",
-                           required = True,
-                           ),
+                     self.event_type_id(default = IS_ONE_OF(
```
flavour: default = ?
Shouldn't this be requires = ?

flavour: & what is different about this requires to the default one?

static/formats/s3csv/cap/warning_priority.xsl
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0"?>
+<xsl:stylesheet
+    xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
+
+    <!-- **********************************************************************
+         CAP Warning Priority - CSV Import Stylesheet
+
+         CSV column...........Format..........Content
+
+		 Priority Rank........integer.........Warning Priority for Event
```
flavour: indent

static/formats/cap/export.xsl
```diff
@@ -144,7 +144,7 @@
                 </xsl:call-template>
             </xsl:if>
 
-            <event><xsl:value-of select="data[@field='event']"/></event>
+            <event><xsl:value-of select="reference[@field='event_type_id']"/></event>
```
flavour: This also needs changing in the Import.xsl (I am doinjg this atm as I need this for my work)

PR #1103, opened 2015-07-10
Fix for bug #1630 - XLS EXPORTS LIMITED TO 65536 ROWS

ahaym:
http://eden.sahanafoundation.org/ticket/1630
Easy workaround for the xlwt export limit by creating a multi-sheet workbook. Done by creating an array of sheets with size determined by the amount of rows being exported.

TODO: redo the code that allows the user to add a title row (currently doesn't work anymore, though the fix should be simple).

### DIFF
--- a/modules/s3/s3codecs/xls.py
+++ b/modules/s3/s3codecs/xls.py
@@ -206,13 +206,23 @@ List Fields %s""" % (request.url, len(headers), len(items[0]), headers, list_fie
         # Create the workbook
         book = xlwt.Workbook(encoding="utf-8")
 
+
+
         # Add a sheet
         # Can't have a / in the sheet_name, so replace any with a space
         sheet_name = str(title.replace("/", " "))
         # sheet_name cannot be over 31 chars
         if len(sheet_name) > 31:
             sheet_name = sheet_name[:31]
-        sheet1 = book.add_sheet(sheet_name)
+        sheets = []
+        rowLimit = 65536 #.xls exports are limited to 65536 rows per sheet, we bypass this by creating multiple sheets
+        sheetnum = len(rows) / rowLimit
+        count = 1
+        while len(sheets) <= sheetnum:
+            sheets.append(book.add_sheet('%s-%s' % (sheet_name, count)))
+            count += 1
+
+
 
         # Styles
         styleLargeHeader = xlwt.XFStyle()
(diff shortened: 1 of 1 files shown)


ahaym:
1. Is it further necessary to notify the user that a multi-sheet workbook is being created? If so, how does Sahana handle error notifications (ie: integrity errors)? Would they be passed through the codec file in python, or through javascript? (I assume the former)

2. The ticket suggests also using xlsxwriter in order to make exporting multiple sheets unnecessary. Is this a worthwhile endeavor? I say this because it may be too much to make the user install two near-identical libraries for a single function in order to bypass a single bug.

nursix:
This looks good. Small cleanups may be necessary, but nothing serious.

Unit tests would be nice here.

For your questions:

1) Issuing a warning that a multi-sheet workbook has been created is not critical (and not easy either since the exporter window closes immediately after successful export - you would thus have to do some JavaScript acrobatics to make this work, and I don't like that idea).

2) This should not be a replacement, but an alternative per requested format extension. If the requested format says ".xls", then we export a multi-sheet workbook in XLS - and if the requested format is ".xlsx", then we export an XLSX.

flavour:
Thanks - could you rebase/compress please?
This also has my most recent commits mixed in...many thanks!

ahaym:
@flavour : I believe I rebased just now (maybe not, I'm new to this resolving conflicts thing.) Thanks!

flavour:
Nope, sorry...I see 16 commits here: https://github.com/flavour/eden/pull/1103
I should be seeing 1
See this doc on compressing commits during the rebase:
http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/Git#Summaryofcodingworkflowusinggit

ahaym:
done, unless i somehow resolved the conflicts incorrectly.

modules/s3/s3codecs/xls.py
```diff
@@ -206,13 +206,23 @@ def encode(self, data_source, **attr):
...
         # sheet_name cannot be over 31 chars
         if len(sheet_name) > 31:
             sheet_name = sheet_name[:31]
-        sheet1 = book.add_sheet(sheet_name)
+        sheets = []
+        rowLimit = 65536 #.xls exports are limited to 65536 rows per sheet, we bypass this by creating multiple sheets
+        sheetnum = len(rows) / rowLimit
+        count = 1
+        while len(sheets) <= sheetnum:
+            sheets.append(book.add_sheet(sheet_name + '-' + str(count)))
```
flavour: In Python, string substition is faster than appending so better to do:
"%s-%s" % (sheet_name, count)

modules/s3/s3codecs/xls.py
```diff
@@ -250,53 +260,54 @@ def encode(self, data_source, **attr):
...
+                # currentRow = sheet.row(1)
+                # currentRow.write(0, str(current.T("Date Exported:")), styleNotes)
+                # currentRow.write(1, request.now, styleNotes)
+                # Fix the size of the last column to display the date
+            #if 16 * COL_WIDTH_MULTIPLIER > width:
+            #    sheet.col(colCnt).width = 16 * COL_WIDTH_MULTIPLIER
 
         # Initialize counters
-        totalCols = colCnt
+            totalCols = colCnt
```
flavour: I'm not clear where the indent should be here but if thiss is correct then the comment should also be indented?

modules/s3/s3codecs/xls.py
```diff
@@ -314,10 +330,10 @@ def encode(self, data_source, **attr):
                 represent = s3_strip_markup(s3_unicode(row[report_groupby]))
                 if subheading != represent:
                     subheading = represent
-                    sheet1.write_merge(rowCnt, rowCnt, 0, totalCols,
+                    sheets(sheetCnt).write_merge(rowCnt, rowCnt, 0, totalCols,
                                        subheading, styleSubHeader)
```
flavour: would prefer to align indents so that the s of subheading was under the r of the 1st rowCnt

flavour:
This looks great, thanks - some very minor comments which I've not just merged/fixed since I wasn't 100% about the totalCols indent.

Bad luck that I happened to be modifying this file at the same time to give conflicts!...ah well, resolving is all good practise ;)

ahaym:
Done, and as far as I know, it works on my end.
http://a.uguu.se/hwbkam_localhost_ItemsinStock%283%29.xls

In this example, I set the row limit to 4. As we can see, the first page will actually have one less row than the limit (in the real world, 65535).

flavour:
Great, thanks, merged. Have you filled out a CLA?
http://bit.ly/SSF-eCLA

ahaym:
@flavour Yes, I have.

PR #1108, opened 2015-07-26
submit for approval and approve/reject alert

biplovbhandari:
@flavour Please merge this


modules/s3db/cap.py
```diff
@@ -1213,6 +1239,41 @@ def cap_rheader(r):
...
+                    auth = current.auth                    
+                    has_permission = auth.s3_has_permission
+                    alert_id = r.id
+                    approved_by = r.record.approved_by
+                    # Display 'Submit for Approval' based on permission  
+                    # and deployment settings
+                    if not approved_by and \
+                       has_permission("create", "cap_alert") and \
+                       has_permission("update", "cap_alert", record_id=alert_id) and \
+                       current.deployment_settings.get_cap_authorisation():
```
flavour: The order of queries matters here since they will stop after the 1st failure.
The first is simplest & hence correct to have first, but then the 2nd simplest is currently last, so I would move that 2nd.

modules/s3db/cap.py
```diff
@@ -1213,6 +1239,41 @@ def cap_rheader(r):
...
+                        # Get the user id for the role alert_approver
+                        agtable = s3db.auth_group
```
flavour: auth_group can always be loaded straight from db.auth_group

modules/s3db/cap.py
```diff
@@ -1213,6 +1239,41 @@ def cap_rheader(r):
...
+                    has_permission = auth.s3_has_permission
+                    alert_id = r.id
+                    approved_by = r.record.approved_by
+                    # Display 'Submit for Approval' based on permission  
+                    # and deployment settings
+                    if not approved_by and \
+                       has_permission("create", "cap_alert") and \
+                       has_permission("update", "cap_alert", record_id=alert_id) and \
+                       current.deployment_settings.get_cap_authorisation():
+                        # Get the user id for the role alert_approver
```
flavour: Can there not be multiple?

flavour: Yes there can, so misleading comment

modules/s3db/cap.py
```diff
@@ -1213,6 +1239,41 @@ def cap_rheader(r):
...
+                        group_row = current.db(agtable.role == "Alert Approver").\
+                                                        select(agtable.id,
+                                                               limitby=(0, 1)).first()
+                        if group_row:
+                            group_id = group_row.id
+                            
+                            user_ids = auth.s3_group_members(group_id) # List of user_ids
+                            pe_ids = [] # List of pe_ids
+                            for user_id in user_ids:
+                                pe_ids.append(auth.s3_user_pe_id(int(user_id)))
```
flavour: I would rewrite this to avoid the loop of DB queries - more efficient to have a single DB query with a .belongs() in (which we could further optimise by using == if we only have a single approver)

biplovbhandari: @flavour I am creating the list of person ids from user ids here

flavour: Yes. That doesn't answer my point in any way. I am saying how you can do that *better*

biplovbhandari:
@flavour Please review this

controllers/cap.py
```diff
@@ -470,6 +470,31 @@ def priority_get():
...
+        Send message to the people with role of Alert Approval
+    """
+    
+    pe_ids = get_vars.get("pe_ids")
+    alert_id = get_vars.get("cap_alert.id")
+    has_permission = auth.s3_has_permission
+    
+    if settings.has_module("msg") and \
+       has_permission("create", "cap_alert") and \
+       has_permission("update", "cap_alert", record_id=alert_id):
```
flavour: Why check for permission to create alerts here?

modules/s3db/cap.py
```diff
@@ -371,6 +371,12 @@ def model(self):
                                                   )),
                            widget = S3MultiSelectWidget(),
                            ),
+                     # approved_on field for recording when the alert was approved
+                     s3_datetime("approved_on",
+                                 readable = False,
+                                 writable = False,
+                                 default = "now",
```
flavour: I don't think we want a default here do we? I think it does nothing since the field is writable=False, but may as well remove for cleaner code (& probably faster if ou look at what this does in the reusable field)

modules/s3db/cap.py
```diff
@@ -1213,6 +1239,42 @@ def cap_rheader(r):
...
+                    auth = current.auth                    
+                    has_permission = auth.s3_has_permission
+                    alert_id = r.id
+                    approved_by = r.record.approved_by
+                    # Display 'Submit for Approval' based on permission  
+                    # and deployment settings
+                    if not approved_by and \
+                       current.deployment_settings.get_cap_authorisation() and \
+                       has_permission("create", "cap_alert") and \
+                       has_permission("update", "cap_alert", record_id=alert_id):
```
flavour: Again, why this multiple permissionc heck?
Why check that you can create alerts & then update this one?
Doesn't make sense to me...

modules/s3db/cap.py
```diff
@@ -1213,6 +1239,42 @@ def cap_rheader(r):
...
+                        # Get the user ids for the role alert_approver
+                        db = current.db
+                        agtable = db.auth_group
+                        query = db(agtable.role == "Alert Approver")._select(agtable.id)
```
flavour: This isn't a query.
The query is (agtable.role == "Alert Approver")

This is a rows....please use variable names which aren't misleading

modules/s3db/cap.py
```diff
@@ -1213,6 +1239,42 @@ def cap_rheader(r):
...
+                        group_rows = db(agtable.id.belongs(query)).select()
```
flavour: If you only need tho use the .id field then better to limit to just this in the select() as less data to read from the DB

biplovbhandari:
@flavour Please merge

biplovbhandari:
@flavour Please merge

PR #1109, opened 2015-07-28
fix for really small syntax error for xls codec

ahaym:
As the title says, it's a <b>really</b> small syntax error, but it appears to break exporting for all report_groupby resources.


ahaym:
Furthermore, that commented out code dealing with title rows right above that section kind of bothers me. How would I access options from either 000_config.py or some other config file so a user can toggle the title row through there? (I apologize for my green-ness, I'm still learning how the Eden codebase is constructed).

flavour:
Thanks :)
deployment_settings are set in either 000_config.py (for a specific local instance) or a template's config.py
They should have a function in modules/s3cfg.py which accesses this & you then put your caller in the code.
In general we prefer backwards-compatibility, so a settings.get_xls_title_row() should default to False
(I didn't work too hard on the name of the fn so feel free to suggest something better!)

PR #1111, opened 2015-08-01
Add option to include a title row in xls exports

ahaym:
Previously, a deployment would have to uncomment/comment multiple non-adjacent lines in the codec itself if they wanted to add a title row to xls_exports. This change allows deployments to do so by specifying "settings.base.xls_title_row = True" in 000_config.py.

Summary:
- The codec in modules/s3/s3codecs/xls.py adds a title row if it is specified in the deployment settings.
- Added function in modules/s3cfg.py to access this setting.

Pull request is currently two commits because I don't know whether you want the commented option in modules/templates/000_config.py or not. The first commit adds the title row code to the codec and creates the corresponding function in s3cfg.py. The second commit adds commented lines to 000_config.py; the user would simply uncomment the line if they want to add a title row. I'll simply revert the second commit if you don't want those in the config file.

Example of a file with the title row added: https://d.maxfile.ro/forjbttqgt.xls

### DIFF
--- a/modules/s3/s3codecs/xls.py
+++ b/modules/s3/s3codecs/xls.py
@@ -263,7 +263,11 @@ List Fields %s""" % (request.url, len(headers), len(items[0]), headers, list_fie
         for sheet in sheets:
             # Header row
             colCnt = 0
-            headerRow = sheet.row(0)
+            # Move this down if a title row will be added
+            if settings.get_xls_title_row():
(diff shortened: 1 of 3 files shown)


flavour:
Thanks - I think this is the kind of setting which would nromally be in a template so we'd prefer the commented example in modules/templates/default.py

ahaym:
@flavour fixed and squashed

PR #1112, opened 2015-08-04
Minor MCOP and SSF Changes

michaelhowden:
* Update MCOP Homepage -> WA-COP
* SSF: Insert CSS from URL var to allow Deployment Map page to be modified to be used in an iframe (eg. http://sunflower.sahanafoundation.org/eden/project/project/map?sector.name=None%2CDeployment&css=.top-bar{display:none}.sidebar{display:none}%23content{left: -250px}%23content h2{display:none}%23footer{display:none}


modules/templates/SSF/views/layout.html
```diff
@@ -33,6 +33,11 @@
  {{# Built by /static/scripts/tools/build.sahana.py }}
  <link href="/{{=appname}}/static/themes/{{=theme}}/eden.min.css" rel="stylesheet" type="text/css" />
 {{pass}}
+{{if "css" in request.vars:}}
+  <style>
+    {{=request.vars.css}}
```
nursix: This is dangerous as it allows arbitrary script injection - should not be accepted into trunk.

nursix: An acceptable method would be to parse the "css" contents, and then inject the CSS rules one by one explicitly.

michaelhowden: Fixed

PR #1114, opened 2015-08-13
FTP Message Channel with FTP subscription from S3Notification

biplovbhandari:
@flavour Please merge


modules/s3/s3xml.py
```diff
@@ -1913,6 +1913,40 @@ def get_struct(self, prefix, name,
             return e
         else:
             raise AttributeError("No table like %s" % tablename)
+        
+    #------------------------------------------------------------------------------
+    def get_data_from_url(self, url):
```
nursix: This entire function is already contained in the etree.parse() method - parse() does accept URLs directly, so it's unnecessary to re-implement that.

nursix: If you really need to read the URL yourself, then it would be as simple as:

    with urllib2.urlopen(url) as source:
            tree = etree.parse(source)

The whole rest with read() and stuff is superfluous, and re-inventing the wheel.

Also - such a function (if at all necessary) should be placed in s3utils - it's not codec-related in any way.

controllers/msg.py
```diff
@@ -834,6 +834,62 @@ def postp(r, output):
     return s3_rest_controller()
 
 # -----------------------------------------------------------------------------
+def ftp_channel():
+    """
+        RESTful CRUD controller for FTP Chaanels
+            - appears in the administration menu
+    """
+    
+    if not auth.s3_logged_in():
```
nursix: Not protocol-safe.

    if not auth.s3_logged_in():
        auth.permission.fail()

controllers/msg.py
```diff
@@ -834,6 +834,62 @@ def postp(r, output):
...
+        label_list_button = T("List FTP Channel"),
+        label_delete_button = T("Delete FTP Channel"),
+        msg_record_created = T("FTP Channel added"),
+        msg_record_modified = T("FTP Channel updated"),
+        msg_record_deleted = T("FTP Channel removed"),
+        msg_list_empty = T("No FTP Channel currently registered")
+        )
+    
+    def postp(r, output):
+        if r.interactive:
```
nursix: Should add:

    and not r.record

It's a lot of processing here which is only needed if you look at multiple records.

modules/s3/s3msg.py
```diff
@@ -1450,6 +1451,49 @@ def post_to_facebook(self, text="", channel_id=None):
...
+                                       table.password,
+                                       table.directory,
+                                       limitby=(0,1)).first()
+                                          
+        # FTP Transfer
+        import ftplib
+        try:
+            remote_ftp_connection = ftplib.FTP(row.site)
+        except ftplib.all_errors as e:
+            current.log.error(str(e))
```
nursix: Should use sys.exc_info instead - str(e) is not reliable.

modules/s3/s3msg.py
```diff
@@ -1450,6 +1451,49 @@ def post_to_facebook(self, text="", channel_id=None):
...
+            return False
+        else:
+            remote_ftp_connection.login(row.username, row.password)
+            try:
+                if len(row.directory):
+                    remote_ftp_connection.cwd(row.directory)
+                    
+                remote_ftp_connection.storbinary("STOR %s" % filename, StringIO(data))
+            except ftplib.error_perm,e:
+                current.log.error(str(e))
```
nursix: Same here - sys.exc_info is the standard

modules/s3/s3notify.py
```diff
@@ -383,6 +387,44 @@ def get_template(path, filenames):
         success = False
         errors = []
 
+        if "FTP" in methods:
```
nursix: Why is that block separate from the other methods? Is it actually misplaced and doesn't belong into that function?

modules/s3/s3notify.py
```diff
@@ -383,6 +387,44 @@ def get_template(path, filenames):
...
+            if not filename:
+                filename = settings.get_msg_upload_filename()
+            from string import Template
+            filename = Template(filename).safe_substitute(s="%(systemname_short)s",
+                                                          r="%(resource)s")
+            filename = filename % meta_data
+            filename = ("%s.%s") % (filename, representation)
+                
+            for resource_id in resource_ids:                        
+                url = "%s/%s.%s" % (page_url, resource_id, representation)
```
nursix: Not safe - the page URL can (and often does) contain a query part.

modules/s3/s3notify.py
```diff
@@ -383,6 +387,44 @@ def get_template(path, filenames):
...
+                _debug("*******************************************************")
+                _debug("Fetching data from %s" % url)
+                
+                data = current.xml.get_data_from_url(url)
```
nursix: This doesn't make any sense to me. Why fetch from a local XML instead of retrieving directly from the function that uses it? This does also circumvent authorization - in fact, that local request is not authorized at all, let alone with a subscriber-specific authentication.

modules/s3/s3notify.py
```diff
@@ -383,6 +387,44 @@ def get_template(path, filenames):
...
+                
+                ftp_sent = current.msg.send_file_by_ftp(data, filename, channel_id)
```
nursix: A much better variant would be to implement that "send XML by FTP" as a method handler for the subscription URL, so that the request to that URL would send the data right away.

nursix:
Terribly sub-standard, and using unsafe and unnecessary methods. This needs conceptual reworking.

nursix:
As I see it, this is using completely the wrong framework - or actually: it pretends to use a framework, but then actually bends it and works around it, which is a clear indicator that this is misplaced - and an invalid generalization.

The idea for the user to be able to configure an FTP URL where to get certain CAP alerts published doesn't make this a messaging protocol, IMO. It's just a file system, after all.

It may be a synchronization method (i.e. repository type), where it would not only be easier to implement, but also make a lot more sense - and for example, logging errors in the system log facility anyhow useful for the users - as opposed to Sync log, which is accessible through the UI. And since each and every FTP target URL would have to be configured as a separate channel anyway, it could just as easily be treated as a sync peer.

As a Sync adapter, it would also be re-usable for other cases - and could be extended to be two-way.

modules/s3/s3notify.py
biplovbhandari: @nursix The representation in url may contain other than XML file (pdf, xls)

nursix:
Sync adapters look like this:
https://github.com/flavour/eden/blob/master/modules/s3/sync_adapter/eden.py
or this:
https://github.com/flavour/eden/blob/master/modules/s3/sync_adapter/wrike.py

...each with its own way of login, pull or push. Shouldn't be difficult to implement one for FTP.

And FTP repositories could then be configured as synchronization peers, including authorization and synchronization schedule etc, for more details about the concept see:
http://eden.sahanafoundation.org/wiki/UserGuidelines/Admin/Synchronization

flavour:
We don't currently use Sync Adapters for any other Notification method, whereas we do use Messaging Channels which is where it seemed consistent to have FTP be just another Messaging Channel. I'm not sure we need to hold up this merge on this point.

nursix:
And what is the point with subscriptions here - what is the point with having 1 FTP channel and 100 users subscribing to it? It would only copy the same file 100 times to the same target location.

No, I think this isn't the right concept.

nursix:
And merging it will only make it harder to pull it out later when we've implemented the right concept of FTP Synchronization.

nursix:
...which would not only be much cleaner, but also indeed generalizable.

modules/s3/s3notify.py
flavour: Right, I think this should be in the
for method in methods:
    if method == "FTP":
       ....
    else:
       ... old code here

nursix:
If it would have to become a subscription concept, then this will need to personalize the FTP channel (target location). Otherwise this doesn't make any sense here.

modules/s3/s3xml.py
flavour: etree.parse() will work for XML URLs, yes good...am sure that can be worked in, but we do also need (in future) to be able to handle non-XML representations like XLS/PDF

controllers/msg.py
flavour: Ah yes, missed that ;)

controllers/msg.py
flavour: Yup, True

modules/s3/s3msg.py
flavour: Yes, better

modules/s3/s3notify.py
flavour: I'm not sure where the query comes from (the filter?), but I guess it can be removed completely if we are accessing the resource_id directly

nursix:
If you wish, I can implement that sync adapter for you so you can meet your deadlines. I'm working in that area at the moment anyway.

modules/s3/s3notify.py
flavour: Althoughh I agree with Dominic that it can come directly from the API rather than going back via a URL.
For XML formats: https://github.com/flavour/eden/blob/master/modules/s3/s3rest.py#L763
For4 PDF/XLS:
https://github.com/flavour/eden/blob/master/modules/s3/s3crud.py#L753

I had actually mentioned the authorisation issue (in depth) on an earlier commit, but then forgot it again when reviewing for the nth time and this hadn't been addressed :/
Anyway this is not required if using trhe API call instead :)

nursix: @biplovbhandari : that speaks even more for a sync adapter rather than this fetch-push combo.

modules/s3/s3xml.py
nursix: XLS/PDF will not work with etree.from_string anyway, so hardly an argument?

modules/s3/s3notify.py
flavour: An interesting idea...if using the current Channel implementation, this would be something like:
/eden/module/resourcename/resource_id.representation/send_via_ftp/channel_id

This would still require us to handle the authorisation though (both for access to the resource & for access to the FTP Channel)

The advantage of this method though is that we don't need to have an if method == "xxx" to call direct from the back-end API.
The disadvantage (other than the need for authorisation handling) is that we'll get the request overheads for every FTP send

modules/s3/s3xml.py
nursix: If it's about /any/ web contents to be fetched, then indeed fetch() should be your method of choice. In that case it is not necessary to implement any new function here.

modules/s3/s3notify.py
nursix: Not necessarily - the method handler can easily queue this and the queue can be handled async.

But all that is actually /included/ in Sync ;)

nursix: ...or yes, right - notify does work asynchronously. But sync reduces the request overheads to one per target repository, which is a lot less than one request per record.

modules/s3/s3notify.py
nursix: The interesting question is where the resource ID comes from? What do people subscribe to?

I would imagine that they want to subscribe to an alert area, not any particular alert - right? But the exported resource would not be the area, but the alerts for that area - so in this case, you wouldn't have a resource ID, but an area_id or location_id as filter.

Therefore I think that just removing the filter and replacing it by an ID isn't quite the whole story.

nursix: Could of course be that I'm wrong, and people do indeed subscribe to particular alerts...dunno. What's the concept here?

modules/s3/s3notify.py
nursix: Unfortunately, notifications do impersonate rather than really login, otherwise you could work around the authorization problem here by adding the session cookie to the request. 

Direct API calls will however be properly authorised even with impersonate, and actually this is the only way from here.

modules/s3/s3notify.py
flavour: People will filter by event_type_id & potentially by Location, yes
However ultimately this becomes a list of Alert resource_ids doesn't it?

modules/s3db/cap.py
```diff
@@ -1811,7 +1841,7 @@ def cap_alert_list_layout(list_id, item_id, resource, rfields, record):
                  _href = URL(c="cap", f="alert", args=[record_id, "profile"]),
                  )
     headline = DIV(headline,
-                   current.T("in %(location)s") % dict(location=location)
+                   current.T(" in %(location)s") % dict(location=location)
```
flavour: I would avoid the leading space inside the T(), have that as a separate line if it is required (easier, more reusable l10n)

flavour:
I don't think your example of 1 FTP channel with multiple users would ever happen in practise, why would it?
The FTP channel is indeed a personal one for the Subscription.

Currently I see no need for an FTP Sync Adapter...sure it may be useful to have such in general and if we can use Sync Adapters for Subscriptions (which seems reasonable) then we could make use of that here, but I don't see any issue with having FTP as a Messaging Channel for a consistent API from there.

modules/s3/s3notify.py
nursix: That how it is done here for the "FTP" method - in order to export every record as a separate file.

However, that's not standard - and a more than questionable generalization of a specific case.

flavour: True that XLS or PDF would work differently, as would other XML variants typically- for these we would expect a single file for all matching records

nursix: Note the way how it's done here - despite having the rows, a get_id() is run against the resource which runs a new db query. 

Instead, something like [row[resource._id] for row in rows] could be used.

nursix: But either way - URL filters will, in the majority of cases, be present in the page_url as that is the whole point about the page_url (see what notify() does to produce it). So, at least they need to be removed before doing simple string-operations with the URL.

nursix:
Yeah - I would strongly disagree with that. 

In the CAP context it may appear that this is just another messaging channel, but just because CAP is designed for messaging doesn't mean all possible CAP export paths are now suddenly and necessarily messaging channels. For file export - regardless whether push or pull - this is simply untrue, mixing apples and oranges and then calling it "consistency".

From a technical perspective, FTP is simply not a messaging protocol - and none of the typical subscription/notification features does apply. But the strong generalizable case is on the Sync-side, where we could re-use FTP for many purposes.

I know you have deadlines, but rewriting this as a sync adapter isn't as expensive as it may seem - and hence I don't think it's worth to fight for this PR.

nursix:
It's bending the specs of the notification framework (more than of the messaging framework), and I am opposed. Especially since it doesn't seem to complicate anything here - adding an FTP channel isn't quite like adding an email address to your user profile, so the inconsistency is already there. And this inconsistency is far greater than the inconsistency with Sync repositories.

All the exporting and sending code can be re-used for a sync adapter, so it's not like re-inventing the whole thing, but rather just moving it where it belongs. And exposing a simple dialog within CAP to configure another FTP repository doesn't seem like a big deal - and you do need to do something similar for FTP subscriptions anyway.

nursix:
...and would you argue that there is something like a personal FTP channel framework that is anywhere near a normal pr_contact as it is used for regular messaging? I think that this is actually much more inconsistent than Sync is - and I don't quite see a case for re-using FTP "messaging". Very far-fetched.

flavour:
OK, I'm convinced ;)
So the pr_subscription table should use repository_id instead of channel_id
& s3notiify doesn't have method = "FTP" but rasther method = "sync"
(Internally the SAMBRO template can label this as FTP if-desired)

nursix:
I would simplify that and use the repository name as label.

So, if you had multiple (and they were accessible for you), you could add multiple peer repositories to your subscription - and in fact, it doesn't really matter whether they are FTP or e.g. Wrike as long as they do accept PUSH.

The sync_repository table could have a "subscription" flag to distinguish those which are open for subscriptions from those which have a schedule and those who are inbound. That'd be easy.

Do you want me to implement the Adapter?

flavour:
> I would simplify that and use the repository name as label.

Instead of repository_id? *shudder*

>So, if you had multiple (and they were accessible for you), you could add multiple peer repositories >to your subscription

Wouldn't each of these be a separate subscription?
hmm, I see that method is a list:string currently so there would indeed just be 1 record if subscribing to the same filter, on the same frequency but via 2 different methods...this does get very ugly for the very rare case where a subscription was coming in via multiple repository_ids.
I would be tempted to remodel this...e.g. what if the user wanted to subscribe via 2 different email addresses? (Seems a much more common usecase than 2 different repository_ids)

> - and in fact, it doesn't really matter whether they are FTP or e.g. Wrike as long as they do accept >PUSH.

This is *exactly* why I said "use repository_id & just label as 'FTP' in the SAMBRO template"

>The sync_repository table could have a "subscription" flag to distinguish those which are open for >subscriptions from those which have a schedule and those who are inbound. That'd be easy.

I don't see the point of this yet, but I guess could be useful later if we have tons of repos for different purposes...def not required yet.

> Do you want me to implement the Adapter?

This seems to be a ballooning requirement which makes me wary of doing anything...I know that you are very busy with other work so asking you to volunteer to do this seems inappropriate, yet equally it is a lot to push back on us.

nursix:
Don't worry - I am currently busy with Sunc repositories anyway, so this is my focus area at the moment and not much of a task switch. Better now than some other time - and implementing it is actually less effort than reviewing it ;)

Subscribing via contact address rather than contact method very much +1 (has been discussed before, was just postponed until a case came up), and I do actually see a case where the same user (e.g. a CAP Publisher) would want to subscribe multiple target repositories to the same alert feed. 

That would especially be meaningful if the target repos used different formats (as has been suggested in this discussion).

Using the repository name as label instead of "FTP" would be equivalent to using email addresses instead of just "Email", btw. Don't get your "shudder" here.

nursix:
The best design would probably be to let the user choose a notification method, and then offer an (Ajax-fetched) list with possible target addresses (email addresses, mobile phone numbers, FTP repositories) for multiple selection.

Extending this with an "Add another one" dialog may be overkill at the moment - but it's an obvious future enhancement.

This is how I imagine subscription to work, but of course there could be other designs in specific use-cases.

nursix:
However, I would imagine the same FTP URL to be subscribed only exactly once to one particular alert feed - which makes the whole effort to employ the subscription framework /at all/ look a bit inflated. The user would actually enter a new FTP URL for every subscription, no?

And if that is indeed the scenario - then you don't actually need anything in S3Notify, but can push the FTP configuration directly to S3Sync.

nursix:
...which would be half the effort for twice the value

nursix:
torsdagen den 13 augusti 2015 09.28.50 skrev  Fran Boon:
> This is exactly why I said "use repository_id & just label as 'FTP' in the
> SAMBRO template"
Yeah yeah - I get it. Of course - I was just suggesting a generic way to 
represent it.

In principle I think when you subscribe via FTP, then you wouldn't just choose 
"FTP" from the list, and then want to enter a URL - and most likely a new one 
for each filter.

I don't quite see subscribing one and the same FTP target location to multiple 
filters as the most likely case, nor the most flexible one.

Could happen, sure - but it seems very unlikely that /all/ filters would go the 
one and the same FTP target location (nor all subscriptions to equally to all 
FTP target locations) - what would be the point? I mean - if that is your 
primary scenario, then what's the point with individual per-user subscriptions 
at all?

Dominic

nursix:
torsdagen den 13 augusti 2015 19.06.36 skrev du:
> Could happen, sure - but it seems very unlikely that all filters would go
> the  one and the same FTP target location (nor all subscriptions to equally
> to all FTP target locations) - what would be the point?

So, in fact - having a drop-down with existing target locations doesn't seem 
to be the most useful thing.

Having an "FTP" method to select for the subscription, and then show an 
(embedded) input dialog to enter the respective FTP target URL and 
credentials, that seems the more common case to me.

Just for the case where you would really want to subscribe the same target URL 
to multiple filters (as I say - I see this as an edge-case), then it may make 
sense to be able to retrieve it. Could just be an autocomplete for the URL?

Like in:

Subscribe to this feed

O EMail
    O myaddress@example.com
    O otheraddress@somethingelse.org
O SMS
    O +4936273564
    O +4973846283
O FTP
    URL: ftp://somewhere.org/path/to/repo
    Username: dumbo
    Password: ****
    Format: CAP | XLS | PDF

(If this looks ugly, copy to monospaced in your editor)

flavour:
>Subscribing via contact address rather than contact method very much +1 (has been discussed >before, was just postponed until a case came up),

Note that this isn't that usecase...just that it seems like if we're redesigning then this should be taken into consderation.
The downside of this 'hardcoded contact within subscription' is that if the user changes their contact details then they need to either update this subnscription manually or we need to ensure that onaccept does it for us which seems fiddly/unreliable...the decoupling of these 2 does have it's benefits.

>and I do actually see a case where the same user (e.g. a CAP Publisher) would want to subscribe >multiple target repositories to the same alert feed.

This isn't the way the workflow is currently defined, but yes I can imagine that in the real world it could well be like that - the FTP repos aren't belonging to the Publisher but it's easier for them to simply add a subscription directly rather than create the user acount for the real subscriber & then use that to subscribe them.

> I would imagine the same FTP URL to be subscribed only exactly once to one particular alert feed 

For CAP, yes

>which makes the whole effort to employ the subscription framework /at all/ look a bit inflated. The >user would actually enter a new FTP URL for every subscription, no?

Yes, but the point is that from the user perspective this is a Subscription just like for Email/SMS
FTP shouldn't look any different from these - other than the details within the contact/channel/repo..

>then you don't actually need anything in S3Notify, but can push the FTP configuration directly to >S3Sync

I worry this makes the debugging harder - an Admin wanting to debug notifications now has to look at 2 different sets of tables rather than having these in a common place...but I see your point.

>In principle I think when you subscribe via FTP, then you wouldn't just choose
>"FTP" from the list, and then want to enter a URL - and most likely a new one
>for each filter.

yes of course
I was just saying that the label would be 'FTP' not 'Sync'

>(embedded) input dialog to enter the respective FTP target URL and
>credentials, that seems the more common case to me.

Thats exactaly how I was proposing it be done for the Channel method - no difference in UX here

> Format: CAP | XLS | PDF

This could also appy to Email (email of PDF/XLS seems a common usecase)

nursix:
So, the planned workflow is that every user has a personal FTP channel which he can manage under "Contacts"? Now I should perhaps add a *shudder* here :D

Yes - uncoupling the contact address and the notification method of subscriptions had exactly that purpose, originally. The subscription could remain the same while you could maintain your email addresses in a central place without having to update every use-case separately. Very much agreed.

On the other hand, you would probably not store the email address itself in the subscription record but rather a reference (pr_contact_id), and if the address (value) for that pr_contact entry would change, it would equally change for all subscriptions.

Surely, if you remove a pr_contact, then you would also remove all subscriptions for that pr_contact entry - but what's wrong with that, actually? I do actually like the idea to be able to remove all subscriptions together with the email address from my account.

The trick would be to make the user aware that a particular of his pr_contact entries is used by subscriptions, and deleting it would also remove these subscriptions, but that should be an easy thing to do.

nursix:
Hmm - yes, you would have to debug FTP sync in a different place if it did not use subscriptions.

But at the same time, it would be a very convenient place to debug because all the sync adapter stuff is in one place ;) as opposed to the combination of S3Notify+S3Msg+S3Sync which is pretty much spread out.

In any case, I think that trade-off is acceptable - especially if the FTP Sync is additionally also maintained from another perspective (i.e. genuine FTP Sync).

nursix:
I need to get off this thread for a while and focus on something else - does not mean I've changed my mind or have lost interest ;) If required, we can continue to discuss this on Skype or Hangout tomorrow (or some other day that is convenient for you), but for now I need to get a job done. Sync, that one too ;)

waidyanatha:
Fran, Dominic,

Really really appreciate you discussing the engineering intricacies, so we have the most effective and efficient solution. However, we are a bit held up on offering the solution for the workshop. 

Unfortunately, the SMS and Email delivery code is also held up with the FTP code for us to test and ready for the workshop. I've asked Biplov to split the code and merge requests so we can test the individual delivery methods.

Thanks a bunch
Nuwan

nursix:
Not sure why - this PR and the entire discussion were only about FTP, and didn't contain any SMS/Email-related code. What's the dependency?

nursix:
Sorry @waidyanatha : I can't see any pending Email/SMS-delivery related code changes, neither in this pull request nor another, and I don't understand what could possibly be the dependency of those delivery methods from the changes discussed here. 

Further I don't see why pending trunk merges should hold up your demo - you can easily demo from any development branch, it does not need to be trunk for that? The other way around, of course, you may need to integrate upstream changes from trunk into your development branch - and may have trouble to resolve conflicts between the two versions. Is that the case?

Else, I'm not sure what you want us to do.

PR #1117, opened 2015-08-17
recent alerts submenu

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -99,6 +99,16 @@ def public():
     return alert()
 
 # -----------------------------------------------------------------------------
+def recent():
+    """
+        Show Recent Alerts
+    """
+    
+    s3.filter = (s3base.FS("info.expires") < request.utcnow)
+    
+    return alert()
+
+# -----------------------------------------------------------------------------
 def alert():
     """ REST controller for CAP Alerts and Components """
 
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -381,6 +381,15 @@ class S3CAPModel(S3Model):
                                  ),
                      *s3_meta_fields())
 
+        list_fields = [(T("Sent"), "sent"),
+                       "info.expires",
+                       "scope",
+                       "info.priority",
+                       "info.event_type_id",
+                       "info.sender_name",
+                       "area.name",
+                       ]
+        
         filter_widgets = [
             S3TextFilter(["identifier",
                           "sender",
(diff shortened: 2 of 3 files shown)


flavour:
Does this really need a separate menu entry?

I would have thought that the default Alerts list would make sense sorted to have the most recent alerts 1st and that it not being too busy should be sufficient, but can easily have an 'Expired/Unexpired filter on it if-required.

To me the title of the menu entry is misleading compared to the filter applied - this isn't "recent" alerts, but rather "unexpired" alerts...

biplovbhandari:
this is as per L64 of https://docs.google.com/spreadsheets/d/1N9xUqZQhignq7QM_7CtrV4zv86sq4lMdJj81IQuch3Y/edit#gid=0
"the expire date of alert less than now" isn't the recent alerts?

controllers/cap.py
```diff
@@ -99,6 +99,16 @@ def public():
     return alert()
 
 # -----------------------------------------------------------------------------
+def recent():
+    """
+        Show Recent Alerts
+    """
+    
+    s3.filter = (s3base.FS("info.expires") < request.utcnow)
```
nursix: FS is good enough (on s3base needed for FS)

nursix:
I think this is a mistake in the specs: recent alerts would be expires > today, not expires < today.

nursix:
Saying that because according to the spec item, recent alerts = active alerts - but expires < today selects expired alerts.

It also says "at least" recent alerts, but it doesn't explicitly disallow "all alerts with the most recent on top", which may be more useful.

nursix:
Meaning: you could list all alerts, with the most recent on top, and set expires >= today as filter default. 

Then the menu item would show only recent alerts at first (and the most recent on top), but the user could modify this filter to select more (or "clear filter" to show all alerts), and also alter the sorting.

By contrast, with s3.filter the user would need to switch to a different page to extend to older alerts.

nursix:
Filter default example here:
https://github.com/flavour/eden/blob/master/modules/templates/CRMT/config.py#L633

nursix:
Btw - with this definition it should be "Active Alerts" or "Current Alerts" not "Recent Alerts".

You can also just put the default filter into the menu item analogous to "Active Missions" in RDRT:
https://github.com/flavour/eden/blob/master/modules/templates/IFRC/menus.py#L393

Then you don't need a separate controller, and the user can still modify the filter to also see expired alerts.

flavour:
Yes, so Dominic is echoing what I am saying, with some further refinements:
* (The menu entry as it stands has the wrong name)
* Rather than having this menu item at all, it would be better to simply have this be the default view for the main Alerts list:
 * Sort with most recent at the top
 * Include a Filter to select expired or not
 * Have this filter default to just showing unexpired (which can be done most easily with the URL filter in the menu link)

PR #1119, opened 2015-08-19
list field and sorting for cap_alert

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -99,6 +99,16 @@ def public():
     return alert()
 
 # -----------------------------------------------------------------------------
+def recent():
+    """
+        Show Recent Alerts
+    """
+    
+    s3.filter = (s3base.FS("info.expires") < request.utcnow)
+    
+    return alert()
+
+# -----------------------------------------------------------------------------
 def alert():
     """ REST controller for CAP Alerts and Components """
 
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -381,6 +381,15 @@ class S3CAPModel(S3Model):
                                  ),
                      *s3_meta_fields())
 
+        list_fields = [(T("Sent"), "sent"),
+                       "info.expires",
+                       "scope",
+                       "info.priority",
+                       "info.event_type_id",
+                       "info.sender_name",
+                       "area.name",
+                       ]
+        
         filter_widgets = [
             S3TextFilter(["identifier",
                           "sender",
@@ -402,12 +411,18 @@ class S3CAPModel(S3Model):
             S3OptionsFilter("info.language",
                             label = T("Language"),
                             ),
+            S3OptionsFilter("info.id",
+                            label = T("Alerts"),
+                            cols = 2,
+                            ),
             ]
 
         configure(tablename,
                   context = {"location": "location.location_id",
                              },
                   filter_widgets = filter_widgets,
+                  list_fields = list_fields,
+                  orderby = "cap_info.expires desc",
                   list_layout = cap_alert_list_layout,
                   list_orderby = "cap_info.expires desc",
                   onvalidation = self.cap_alert_form_validation,
--- a/modules/templates/SAMBRO/config.py
+++ b/modules/templates/SAMBRO/config.py
@@ -150,6 +150,30 @@ def config(settings):
     settings.customise_msg_twitter_channel_resource = customise_msg_twitter_channel_resource
 
     # -------------------------------------------------------------------------
+    def default_alert_filter(selector, tablename = None):
+        """
+            Default filter for alert
+             - show unexpired alerts by default
+        """
+        
+        s3db = current.s3db
+        itable = s3db.cap_info
+        rows = current.db(itable.expires >= current.request.utcnow).select(itable.id)
+        return [row.id for row in rows]
+    
+    # -------------------------------------------------------------------------
+    def customise_cap_alert_controller(**attr):
+        
+        from s3 import s3_set_default_filter
+        s3_set_default_filter("info.id",
+                              default_alert_filter,
+                              tablename = "cap_alert")
+        
+        return attr
+    
+    settings.customise_cap_alert_controller = customise_cap_alert_controller
+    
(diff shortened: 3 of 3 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -150,6 +150,30 @@ def onaccept(form):
...
     # -------------------------------------------------------------------------
+    def default_alert_filter(selector, tablename = None):
+        """
+            Default filter for alert
+             - show unexpired alerts by default
+        """
+        
+        s3db = current.s3db
+        itable = s3db.cap_info
+        rows = current.db(itable.expires >= current.request.utcnow).select(itable.id)
```
nursix: Now that confuses me completely: 
You want to filter alerts by info.expires, but you implement a filter widget for info.id, and run a db query for info.id based on info.expires?? Why don't you implement a date filter for info.expires and set current.request.utcnow as default (=no db query)? What's useful about the info.id filter widget? Aren't you hiding from the user what you actually filter by?

Sorry, this doesn't make any sense to me.

controllers/cap.py
```diff
@@ -99,6 +99,16 @@ def public():
     return alert()
 
 # -----------------------------------------------------------------------------
+def recent():
+    """
+        Show Recent Alerts
+    """
+    
+    s3.filter = (s3base.FS("info.expires") < request.utcnow)
```
nursix: It's still the same - and still not making sense. Recent alerts, according to the specs, are alerts which are still active, not those which have expired yesterday. This s3.filter excludes all unexpired alerts - how can this be correct?

modules/s3db/cap.py
```diff
@@ -402,12 +411,18 @@ def model(self):
             S3OptionsFilter("info.language",
                             label = T("Language"),
                             ),
+            S3OptionsFilter("info.id",
```
nursix: How does this filter make sense?

modules/templates/SAMBRO/config.py
nursix: I was thinking like this:

    def customise_cap_alert_controller(**attr):

            now = current.request.utcnow()
            today_start = now.replace(hour=0, minute=0, second=0, microsecond=0)
            from s3 import s3_set_default_filter
            s3_set_default_filter("info.expires",
                                             {"ge": today_start},
                                             tablename = "cap_alert",
                                             )

And then have a filter widget like S3DateFilter("info.expires").

biplovbhandari:
sorry I closed the earlier PR https://github.com/flavour/eden/pulls?q=is%3Apr+is%3Aclosed
but again it is showing here :(

modules/templates/SAMBRO/config.py
nursix: NB If you would implement an S3DateFilter for info.expires instead, then you could perhaps see this default as generic, hence instead of calling set_default_filter in the template, you could pass the default filter as "default" parameter to the widget directly (in the model), similar to:

    S3DateFilter("info.expires",
                         default = {"ge": current.request.utcnow().replace(hour=0, minute=0, second=0, microsecond=0)},
                         )

That may make things a whole lot easier.

biplovbhandari: isn't "ge" and "le" input labels? How do use default here?

flavour: ge is an operator...which can link to a label or a default.

Have you tried this?

biplovbhandari: S3DateFilter("info.expires",
                         default = {"ge": current.request.utcnow.replace(hour=0, minute=0, second=0, microsecond=0)},
                         ),

This is not working for me?

nursix: Too little information: "is not working for me"

nursix: This may work if you change "ge" into "info.expires__ge", like:

    S3DateFilter("info.expires"
                    {"info.expires__ge": current.request.utcnow.replace(hour=0, minute=0, second=0, microsecond=0)},
                   )

Generally useful to know what exactly it does or does not - rather than "is not working for me".

biplovbhandari: I am sorry Dominic...had to go out...
The commit link is https://github.com/biplovbhandari/eden/commit/3ebb7bcf7a97816abba7bd19b8ad125b42391498 as you have suggested.
I wipe the database and start again. This is how it looks: http://i.imgur.com/D4gjrAx.png 
when I add the date then it looks like this: http://i.imgur.com/BLdOKyr.png
Should not the second one by default?

flavour: The End Date doesn't need setting.

NB I would label this filter as 'Expiry Date' as if I didn't see the code then I weould assume this was 'Date Created' or "Date Published' (Ideally the same date!)

flavour: Whilst Dominic is leaning towards this view (which I understand) I actually prefer the alternate approach of making this simpler for users by simply having a radio button to select between 'Expired', 'Unexpired' and 'All', defaulting to 'Unexpired'

nursix: That would be even better @flavour :) no real reason for the user having to fumble with the expiry date, the actual intent is pretty clear.

nursix: ...was just following the spec in the document as I thought that was what Biplov was looking at

nursix: @biplovbhandari No, if you only set a default for the "ge" operator, then that would only be for the "from" field, while the "to" field (which is "le") remains empty. 

If you want both, then you must set a default for both (it's a dict!) - but I didn't yet see that requirement?

nursix: Admittedly, this translation of "has not expired yet" into "expiry date is greater or equal than today" can be a bit of a brain twister because "expiry" itself suggests an end date hence you spontaneously conclude the filter to be "until" not "from" (which is already reflected in the requirements spec).

Since that error seems common and hence would also affect the user and leave them confused, Fran's proposed simplification doesn't just seem better, but even necessary.

PR #1130, opened 2015-08-24
Added Event as a new field - text

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -530,14 +530,6 @@ def alert():
                                                                   filterby="alert_id",
                                                                   filter_opts=(r.id,),
                                                                   ))
-            elif r.component_name == "location":
-                # Limit to those for this Alert
-                r.component.table.area_id.requires = IS_EMPTY_OR(
-                                                        IS_ONE_OF(db, "cap_area.id",
-                                                                  s3db.cap_area_represent,
-                                                                  filterby="alert_id",
-                                                                  filter_opts=(r.id,),
-                                                                  ))
 
             # @ToDo: Move inside correct component context (None?)
             post_vars = request.post_vars
@@ -1026,6 +1018,12 @@ def info_fields_comments():
           _title="%s|%s" % (
               T("Denotes the category of the subject event of the alert message"),
               T("You may select multiple categories by holding down control and then selecting the items.")))
+
+    table.event.comment = DIV(
+          _class="tooltip",
+          _title="%s|%s" % (
+              T("The text denoting the type of the subject event of the alert message"),
+              T("If not specified, will the same as the Event Type.")))
 
     table.response_type.comment = DIV(
           _class="tooltip",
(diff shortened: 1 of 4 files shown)


flavour:
What is the purpose of this separate field?

biplovbhandari:
This is as suggested by Nuwan...beside event type (as parameter in cap XML), a free text field called event (as in CAP 1.2) should be there...if the user doesn't fill in the event value should be same as event type...last time discussed with Nuwan

flavour:
What exactly is this for?
As I see it, the purpose here is to allow the Alert Editor to set the Event field to something other than an event_type_id.
The reason they need to do this is because they don't have write access to the event_type_id lookup list?
What is the usecase for this? When would event be set to something that shouldn't go into event_type_id?
Note that the list will be filled with junk anyway since we import whatever people put into event into event_type_id for imported alerts...although I guess that not all deployments may import alerts.

waidyanatha:
The distinction between an eden.event_type and a cap.event is that we have a knowledge base that depends on the event_type. One example, would be the meteorological warning_priorities that are defined on the event_type = "tropical cyclone". We would define a set of warning priorities based on the event_type. Similarly there is other knowledge such as the event_type specific cap.expire (date/time) offset that we would pre-program. Therefore, we need to include the event_type in our design.

cap.event, in general is free text. For example, for an event_type = "terrorist attack", based on the actual incident we would fill the cap.event value as "terrorist bomb explosion". Another could be "terrorist chemical warfare attack". Both of them would be associated with event_type = "terrorist attack" which certain agencies would subscribe to such alert. There might be an a non-terrorist bomb explosion because some factory accidentally diffused a bomb. National security would not care too much about that attack to activate their people and hence would not subscribe to such events (reality they would because they are curious about all bomb explosions, but trying to make a case).

For those alerts, like an asteroid threat would not have a cap template. The user would use the default template to generate an alert. In such instances, we would need to set the event_type, which would be associated with some database relationships (knowledge) as well as define the actual cap.event.

Hope this explains

nursix:
Oops, careful what you give away ;)

> 24 aug 2015 kl. 15:02 skrev Nuwan Waidyanatha <notifications@github.com>:
> 
> National security would not care too much about that attack to activate their people and hence would not subscribe to such events

nursix:
I think the essence of this is that "event" is a (short) description of the particular event, whereas event_type is a classification of the event. In the very basic form, event would be just a textual representation of event_type, but it could be more elaborate.

If this were to become more formal, a hierarchical taxonomy would be a possible solution - with "event" being the leaf category, and "event_type" the corresponding root category. The advantage of that would be that consumers could subscribe to sub-categories at any arbitrary level (and knowledge-base could link to sub-categories as well) - without giving up the basic CAP data structure.

A bit Sahana-added value, so to speak ;)

PR #1137, opened 2015-08-28
project_Details bug fix

raj454raj:
Ajax call breaks when submitting a record for staff.

### DIFF
--- a/controllers/project.py
+++ b/controllers/project.py
@@ -1018,13 +1018,6 @@ def indicator_data():
 
     return s3_rest_controller()
 
-# =============================================================================
-# Community Volunteers
-# =============================================================================
-#def human_resource():
-#    """ RESTful controller for Project <> Staff Assignments """
-#    return s3_rest_controller()
-
 def person():
     """ RESTful controller for Community Volunteers """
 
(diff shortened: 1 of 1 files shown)


raj454raj:
@flavour: Just a sec merging one more bug :p

nursix:
The project_Details bug fix is missing ;)

nursix:
Ah - it's already in trunk, so it doesn't appear in the diff anymore. Ok ok

raj454raj:
@nursix : Fran is too fast for a second :D :D

controllers/project.py
```diff
@@ -1216,4 +1216,13 @@ def campaign_response_summary():
 
     return s3_rest_controller()
 
+# -----------------------------------------------------------------------------
+def human_resource_project():
```
nursix: This used to be the "human_resource" controller, which is still defined in this file but commented. Maybe just uncomment+rename, or remove the commented version and replace with this one?

raj454raj: @nursix : Done :+1:

nursix: Fine by me, good to merge.

PR #1138, opened 2015-08-30
CAP: minor correction

biplovbhandari:
@flavour Please merge...

### DIFF
--- a/modules/s3/s3crud.py
+++ b/modules/s3/s3crud.py
@@ -1925,18 +1925,25 @@ class S3CRUD(S3Method):
                 approve = FORM(INPUT(_value=T("Approve"),
                                     _type="submit",
                                     _name="approve-btn",
-                                    _id="approve-btn"))
+                                    _id="approve-btn",
+                                    _class="action-btn"))
 
                 reject = FORM(INPUT(_value=T("Reject"),
                                     _type="submit",
                                     _name="reject-btn",
-                                    _id="reject-btn"))
+                                    _id="reject-btn",
+                                    _class="action-btn"))
+                
+                edit = A(T("Edit"),
+                         _href=r.url(id=r.id, method="update",
+                                     vars={"_next": r.url(id=r.id, method="review")}),
+                         _class="action-btn")
 
                 cancel = A(T("Cancel"),
                            _href=r.url(id=0),
                            _class="action-lnk")
 
-                output["approve_form"] = DIV(TABLE(TR(approve, reject, cancel)),
+                output["approve_form"] = DIV(TABLE(TR(approve, reject, edit, cancel)),
                                              _id="approve_form")
 
                 reviewing = False
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -502,7 +502,7 @@ def alert():
 
                     response.s3.stylesheets.append("../themes/default/cap.css")
 
-                elif r.method != "import":
+                elif r.method != "import" and not get_vars.get("_next"):
                     s3.crud.submit_style = "hide"
                     s3.crud.custom_submit = (("edit_info",
                                               T("Save and edit information"),
@@ -592,6 +592,8 @@ def alert():
             r.next = URL(c="cap", f="alert", args=[lastid, "info"])
 
         if r.interactive:
+            if get_vars.get("_next"):
+                r.next = get_vars.get("_next")
             #if r.component_name == "info":
             #    update_url = URL(f="info", args=["[id]"])
             #    s3_action_buttons(r, update_url=update_url)
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -1128,8 +1128,8 @@ class S3CAPModel(S3Model):
         oid = settings.get_cap_identifier_oid()
         suffix = settings.get_cap_identifier_suffix()
 
-        return "%s-%s-%s-%03d%s%s" % \
-                    (prefix, oid, _time, next_id, ["", "-"][bool(suffix)], suffix)
+        return "%s-%s-%s-%03d-%s" % \
+                    (prefix, oid, _time, next_id, suffix)
 
     # -------------------------------------------------------------------------
     @staticmethod
@@ -1414,7 +1414,8 @@ def cap_rheader(r):
 
                         # Display 'Submit for Approval' based on permission  
                         # and deployment settings
-                        if not r.record.approved_by and \
+                        if not current.request.get_vars.get("_next") and \
+                           not r.record.approved_by and \
                            current.deployment_settings.get_cap_authorisation() and \
                            current.auth.s3_has_permission("update", "cap_alert",
                                                           record_id=alert_id):
--- a/modules/templates/SAMBRO/menus.py
+++ b/modules/templates/SAMBRO/menus.py
@@ -52,9 +52,14 @@ class S3MainMenu(default.S3MainMenu):
             else:
                 # Publisher sees minimal options
                 # @ToDo: Add role check here once role defined
-                return [homepage(),
-                        homepage("cap"),
-                        ]
+                menus_ = [homepage(),
+                          homepage("cap"),
+                          ]
+                
+                if auth.s3_has_role("MAP_ADMIN"):
+                    menus_.append(homepage("gis"),)
+                    
+                return menus_ 
 
         # Public or CUG reader sees minimal options
         return [homepage(),


biplovbhandari:
@flavour Please review and merge

modules/s3/s3crud.py
```diff
@@ -464,6 +466,24 @@ def create(self, r, **attr):
...
+            
+            @param r: the S3Request
+            @param attr: dictionary of parameters for the method handler
+        """
+        
+        if current.s3db.resource(r.tablename,
+                                 id=r.id,
+                                 unapproved=True).approve():            
+            current.session.confirmation = current.T("Record Approved")
+            redirect(r.url(method="", id=0))
```
nursix: Please use r.next, do not redirect from here.

modules/s3/s3crud.py
```diff
@@ -464,6 +466,24 @@ def create(self, r, **attr):
...
+        else:            
+            current.session.error = current.T("Unable to approve the record")
+            redirect(r.url(method="", id=0))
```
nursix: Please use r.error, do not redirect from here.

modules/s3/s3crud.py
```diff
@@ -1925,18 +1945,25 @@ def review(self, r, **attr):
...
+                                    _class="action-btn"))
 
                 reject = FORM(INPUT(_value=T("Reject"),
                                     _type="submit",
                                     _name="reject-btn",
-                                    _id="reject-btn"))
+                                    _id="reject-btn",
+                                    _class="action-btn"))
+                
+                edit = A(T("Edit"),
```
nursix: I'm not sure that this button is really required now that unapproved records are visible for reviewers in regular CRUD. It seems a bit overcrowded with buttons in this view.

modules/s3/s3crud.py
```diff
@@ -109,6 +109,8 @@ def apply_method(self, r, **attr):
             output = self.read(r, **attr)
         elif method == "update":
             output = self.update(r, **attr)
+        elif method == "approve":
+            output = self.approve(r, **attr)
```
nursix: Approve should be a POST, not a GET (REST standard = GET must not change db status)

modules/s3/s3crud.py
```diff
@@ -464,6 +466,24 @@ def create(self, r, **attr):
...
+    def approve(self, r, **attr):
+        """ 
+            Approve the record 
+            
+            @param r: the S3Request
+            @param attr: dictionary of parameters for the method handler
+        """
+        
+        if current.s3db.resource(r.tablename,
+                                 id=r.id,
```
nursix: This doesn't even check whether there is an r.id - but that must be checked. You can not assume that.

nursix: Ideally you follow the pattern for "review" and hold those two things together.

modules/s3/s3crud.py
nursix: Correction: self.next not r.next

modules/s3/s3crud.py
```diff
@@ -464,6 +466,24 @@ def create(self, r, **attr):
...
     # -------------------------------------------------------------------------
+    def approve(self, r, **attr):
+        """ 
+            Approve the record 
+            
+            @param r: the S3Request
+            @param attr: dictionary of parameters for the method handler
+        """
+        
+        if current.s3db.resource(r.tablename,
```
nursix: This is incorrect - must be s3db.resource(self.resource), and must apply URL filters.

modules/s3/s3crud.py
nursix: Ideally, you follow the standard for next as it is implemented in other methods (e.g. UPDATE).

nursix:
I'm sorry - but this does not follow the relevant standards for REST nor CRUD, which makes it massively inconsistent and risky.

- "Approve" is a method to change the db status, and should therefore be POST not GET. 
- Post-action redirection in S3Methods uses self.next, not redirect. You should never redirect from an S3Method directly
- Error handling should go through r.error to be safe for non-interactive clients
- The target resource can not be concluded from r.tablename (which is the master table) - all S3Methods must use self.resource and self.record_id to work so that they work on both components and link tables
- You should *never* re-instantiate the target resource for manipulation, that is incorrect. Instead, if you need to operate on unapproved records, you must include them in the first place (REST)

I'm afraid this implementation is completely unacceptable (because it is incorrect).

nursix:
Note that approval does actually not work from components (obviously can't approve a component without approving the master).

That is why "review" is only available for master requests - and thus it may be sufficient to throw an error for component requests, and otherwise stick with self.resource.

Re-instantiation is actually not necessary, because if you see the record in a list-view (and hence have that "Approve" action button), then you already have "unapproved" in CRUD. If you don't have that, then the user has no review permission anyway, and hence should not be able to use approve in the first place.

GET is not correct in any case, though. As the very minimum it must be caught behind a confirmation dialog - but actually it must be RESTful a POST request, a simple URL GET should be able to approve a record.

But what is dramatically wrong with that method is that you do not check whether the user is actually permitted to approve the record. You seem to take that for granted. That combined with a GET request is a security hole big like barn door.

nursix:
With this implementation, anyone who wants to publish an alert just needs to open it and change the URL method into /approve. They do not even need update permission, nor any complicated request construction - read-access and a browser are good enough.

nursix:
The other weird question is why this is using the "approve" workflow at all.

In an earlier, similar case, we had implemented a "publish" workflow, which is why there is a PUBLISH permission. This simply used a "public" flag in the target record, and imposed a filter for anonymous users that hid all records with public=False.

This would be a lot easier and safer to implement, and all those weird modifications to the approval concept wouldn't be necessary, and interference with security could be avoided.

nursix:
This publish-workflow was for CMS posts (newsfeed). Very simple resource-specific method, that was exposed for users with auth.has_permission("publish") == True (this permission does exist and can be managed).

If all you need is to hide alerts from anonymous users, then I think all this is effort and complications are way beyond reasonable.

nursix:
...not everything needs to be implemented at the framework level ;)

nursix:
Really - this is very confusing. Why do you "publish" by approving?

If approving an alert and publishing it is one and the same workflow step - then why make it complicated and use record approval?

I would have though that these are two independent steps - approval means "authorize it", and publish means "send it out", in which case indeed you need to have both an approval-concept and a publishing workflow.

But now it looks as if that is one and the same thing - so "approving" a record means to make it public at the same time. And if that is the case, then a simple publish-workflow would have been just as good - and all those inconsistent and insecure framework modifications wouldn't have been necessary.

If though "approving" and "publishing" are two independent workflows, as originally argued, then it seems wrong to make the "Publish" button call the "approve" method.

nursix:
In my memory it was "you can only publish an alert when it has been approved" - without though making publishing mandatory at the moment the alert has been authorized. I would understand if that requirement has changed - but then it seems odd to employ the complexity of record approval for that purpose, and make major framework modifications just for that one case.

===
Generally I do see a problem with all this:

The normal meaning of record approval (in the sense of the concept that was implemented) is "to allow the record *to enter* the database" - which is why it has a reject-component. This meaning has been massively undermined by the recent changes, and records are now allowed to enter regular database workflows without first being approved.

But in the alert case it seems that it actually has the meaning "to make a record available for a wider audience" - while the record gets accepted into the database upfront and without further procedure. That is though "publishing" - and not "approval".

Both concepts can co-exist, which is why we do have separate permissions for them (APPROVE/REJECT vs. PUBLISH), and typically would implement separate functionality too.

But if you don't need both, then you should actually stick to the right concept instead of bending the opposite one to fit your needs while breaking it's original meaning - and without need, I must say.

nursix:
I think this has all been terribly mixed up due to context-specific terminology being mapped to the wrong framework concept - and unfortunately it is causing the loss of that very valuable, and much needed security concept to prevent unwanted records from entering regular workflows.

nursix:
To make that clearer:

The actual idea behind record approval is (as needed in NYCP and other templates):
General Public => Data Entry => Approval => Application Workflows

You need seems to be (SAMBRO):
Application Workflows => Approval => Publish => General Public

So, it's turning a framework concept completely upside down - so much that it actually gets lost.

And I do actually disagree that both concepts can be lumped together - because in fact they can co-exist and be separate workflows:

General Public => Data Entry => Approval => Application Workflows => Approval => Publish => General Public

...and should therefore be kept separate.

nursix:
...and since public data entry (e.g. crowdsourcing) is still an aspect in Sahana, we need to retain the workflow entry barrier option as we had it, even if that is not required for SAMBRO.

In all previous cases where we needed controlled publishing (workflow exit barrier) we did not use the approval-workflow, but a "public" flag of some kind. Michael Howden has correctly recognized this as controlled=>uncontrolled transition (=output chain), and that it is different from the record approval concept (which is for the input chain).

However, SAMBRO project terminology seems to suggest that the application environment is uncontrolled while the general public is the controlled area - but even if that is the case from a business logic perspective, it does still not make "publishing alerts" an input chain, because as Fran explained: application workflows come before that.

Unlike what I first expected, SAMBRO doesn't seem to impose any input chain controls other than trusting the users - every workflow input (e.g. a new alert) is automatically trusted and allowed to enter application workflows if the user has permission to create it.

I did expect that alerts published by other sources and automatically imported into SAMBRO would require approval before being processed further in SAMBRO (=input control), and could be rejected if they were not trustworthy.

But instead, SAMBRO aims to impose output chain controls by requiring separate authorization for workflow items to become available beyond the application workflows (e.g. alerts *to* the general public, not *from* other sources).

So, I misinterpreted the need - and hence accepted certain modifications to the approval concept even though they appeared incorrect. But now I see that it is actually a completely different kind of thing that is needed :/

modules/s3/s3crud.py
biplovbhandari: @nursix Overcrowded, yes but looks okey rather than having to enter the url...The _next helps to come to the previous view rather than again having to type in the url...The label in the button is clear and does not confuse approver ?

biplovbhandari:
@flavour Please merge

PR #1144, opened 2015-09-09
CAP: for cap_area auto assign the info_if if only one record is present

biplovbhandari:
@flavour Please merge...


biplovbhandari:
@flavour Please merge

PR #1161, opened 2015-09-27
FTP Sync Adapter

biplovbhandari:
@flavour Please review


modules/s3layouts.py
```diff
@@ -405,7 +405,7 @@ def layout(item):
             return None
 
 # =============================================================================
-class S3AddResourceLink(S3NavigationItem):
+class S3ResourceLink(S3NavigationItem):
```
nursix: -1 to renaming this. It's unnecessary and "S3ResourceLink" says nothing, it's confusing - please revert it. The main use-case is to *add* a new record.

Using the same layout for an update-link is well possible - and if for that edge-case the name seems wrong, then do something like "from s3layouts import S3AddResourceLink as S3UpdateResourceLink".

modules/templates/SAMBRO/controllers.py
```diff
@@ -370,30 +373,105 @@ def _manage_subscriptions(self, resources, filters):
...
+            if properties and json.loads(properties)["repository_id"]:            
+                user_repository_id = json.loads(properties)["repository_id"]            
+            else:
+                user_repository_id = ftp_rows.first().id
+            
+            UPDATE_REPOSITORY = T("Update Repository")
+
+            repository_id.comment = S3ResourceLink(c="sync", f="repository",
+                                                   m="update",
+                                                   args=[user_repository_id],
```
nursix: Note that this does not properly authorize the particular record!
It will be better to check permission for the user_repository_id first - and only then add the update_link

models/tasks.py
```diff
@@ -619,8 +619,10 @@ def sync_synchronize(repository_id, user_id=None, manual=False):
         """
             Run all tasks for a repository, to be called from scheduler
         """
-
-        auth.s3_impersonate(user_id)
+        
+        if user_id:
```
nursix: Actually not - shall even be run if user_id is None.

modules/templates/SAMBRO/controllers.py
```diff
@@ -370,30 +373,105 @@ def _manage_subscriptions(self, resources, filters):
...
+                                                   label=UPDATE_REPOSITORY,
+                                                   title=UPDATE_REPOSITORY,
+                                                   tooltip = UPDATE_REPOSITORY,
```
nursix: Repeating the title is not quite the point with a tooltip ;)

modules/templates/SAMBRO/controllers.py
```diff
@@ -370,30 +373,105 @@ def _manage_subscriptions(self, resources, filters):
...
-                            _class="control-group"))
-        form.append(fieldset)
+        
+        # Sync Row
+        properties = subscription["comments"]  
+        
+        synctable = s3db.sync_repository
+        query = (synctable.apitype == "ftp") & \
+                (synctable.deleted != True) & \
+                (synctable.owned_by_user == current.auth.user.id)
```
nursix: Working around authorization and construction your own?

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+        """
+
+        repository = self.repository
+
+        # Log the operation
+        log = repository.log
+        log.write(repository_id = repository.id,
+                  resource_name = task.resource_name,
+                  transmission = log.OUT,
+                  mode = log.PUSH,
```
nursix: Err? No - this is PULL

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+        
+        if not url:
+            message = "Remote URL required for FTP Push"
+            log.write(repository_id = repository.id,
+                      resource_name = resource_name,
+                      transmission = log.OUT,
+                      mode = log.PUSH,
+                      action = "send",
+                      remote = remote,
+                      result = log.ERROR,
```
nursix: this is FATAL not ERROR - because you can't even try to push.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                      
+                    for resource_id in resource_ids:
+                        resource.clear_query()
+                        resource.add_filter(~(FS("id") == resource_id))
+                        
+                        data = self._get_data(resource, representation)
+                        # FTP Transfer
+                        import ftplib
+                        remote = True                        
+                        try:
```
nursix: This should go into login()

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                    remote = True  
+                    try:
+                        remote_ftp_connection = ftplib.FTP(url)
+                    except ftplib.all_errors as e:
+                        message = str(e)
+                        result = log.ERROR
+                        output = current.xml.json_message(False, 400, message)
+                    else:
+                        try:
+                            remote_ftp_connection.login(username, password)
```
nursix: This should go into login

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                            )
+        
+        if request.transformable():
+            return resource.export_xml(stylesheet = request.stylesheet(),
+                                       pretty_print = True,
+                                       )
+        
+        else:
+            if representation == "html":
+                # @ToDo Implement this
```
nursix: Implement what exactly?
(If you don't have a spec, then remove this alternative for now)

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                                result = log.ERROR
+                                output = current.xml.json_message(False, 400, message)
+                            else:
+                                try:                    
+                                    remote_ftp_connection.storbinary("STOR %s" % filename,
+                                                                     StringIO(data))
+                                except ftplib.error_perm as e:
+                                    message = str(e)
+                                    result = log.ERROR
+                                    output = current.xml.json_message(False, 400, message)
```
nursix: Why would you return a JSON message? Standard is just the error message.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                password = repository.password
+                
+                # For log operation                
+                if task.multiple_file:
+                    if type(resource_ids) is not list:
+                        resource_ids = [resource_ids]                    
+                      
+                    for resource_id in resource_ids:
+                        resource.clear_query()
+                        resource.add_filter(~(FS("id") == resource_id))
```
nursix: I don't understand the ~ here - you do want just this record, not exclude this record, or?

nursix:
I fundamentally disagree with "S3ResourceLink" - the name was chosen to clearly document the purpose of this class, and I don't want to hide it for the sake of a single edge-case. You can still add a comment to document the deviation, or assign an alternative class in-place, if you think it would otherwise be confusing, but not rename everywhere else where "S3AddResourceLink" gives clarity and makes the code more understandable.

nursix:
I can review the FTP Sync Adapter once you've got that out of the way.

biplovbhandari:
@nursix Thank you so much for the review
I would need ftp object in the push once I move the ftp login to login method...How do I retrieve the ftp object on the push?

nursix:
You can just put it into self no? S3Sync will always use the same adapter instance it has logged in with.
So, during login - set self.ftp_connection

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+"""
+
+import sys
+import os
+
+from gluon import *
+
+from ..s3sync import S3SyncBaseAdapter
+
+from ..s3query import S3URLQuery, S3FieldSelector as FS
```
nursix: Just:
    from s3query import S3URLQuery, FS

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+    
+    # -------------------------------------------------------------------------
+    def login(self):
+        """
+            Login at the peer repository
+
+            @return: None if successful, otherwise the error
+        """
+        
+        # Need FTP object so login done at push
```
nursix: You can store the FTP object in self - S3Sync will always use the same connector instance it has logged in with.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+            log.write(repository_id = repository.id,
+                      resource_name = resource_name,
+                      transmission = log.OUT,
+                      mode = log.PUSH,
+                      action = "send",
+                      remote = remote,
+                      result = log.ERROR,
+                      message = message,
+                      )
+            output = current.xml.json_message(False, 400, message)
```
nursix: output is just the message, not json_message

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+            msince = task.last_push            
+            if msince is not None and "modified_on" in table.fields:
+                resource.add_filter(table.modified_on > msince)
+                
+            # Get the ID of the resource after filter and msince
+            resource_ids = resource.get_id()
+            
+            # No Changes since last push?
+            if resource_ids is None:
+                log.write(repository_id = repository.id,
```
nursix: You could DRY this a little bit, and have just one log.write at the end. This is easier to read and maintain

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                              mode = log.PUSH,
+                              action = "send",
+                              remote = remote,
+                              result = result,
+                              message = message,
+                              )
+        
+        # Returns after operation is complete
+        if output is not None:
+            resource.muntil = None
```
nursix: You're not supposed to write to resource.muntil here - use a local variable (mtime).

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+from ..s3query import S3URLQuery, S3FieldSelector as FS
+from ..s3rest import S3Request
+from ..s3export import S3Exporter
+
+try:
+    from cStringIO import StringIO    # Faster, where available
+except:
+    from StringIO import StringIO
+
+DEBUG = False
```
nursix: I know you copy/pasted it, but ideally - remove this.
We should not use any local _debug anymore - but always current.log.debug where it makes sense.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                else:
+                    data = self._get_data(resource, representation)
+                    
+                    # FTP Transfer
+                    import ftplib
+                    remote = True  
+                    try:
+                        remote_ftp_connection = ftplib.FTP(url)
+                    except ftplib.all_errors as e:
+                        message = str(e)
```
nursix: I think I've said that before: our standard is:

    except ftplib.all_errors:
        message = sys.exc_info()[1]

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,334 @@
...
+                            remote_ftp_connection.login(username, password)
+                        except ftplib.error_perm as e:
+                            message = str(e)
+                            result = log.ERROR
+                            output = current.xml.json_message(False, 400, message)
+                        else:
+                            try:                    
+                                remote_ftp_connection.storbinary("STOR %s" % filename,
+                                                                 StringIO(data))
+                            except ftplib.error_perm as e:
```
nursix: Same - we should use sys.exc_info in Sahana.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+            properties = json.loads(properties)
+            
+            if "resource_filter_ids" in properties:
+                sync_resource_filter_table = current.s3db.sync_resource_filter
+                query = (sync_resource_filter_table.owned_by_user == current.auth.user.id) & \
+                        (sync_resource_filter_table.task_id == properties["sync_task_id"])
+                rows = current.db(query).select(sync_resource_filter_table.id,
+                                                orderby=sync_resource_filter_table.id)
+                for row in rows:
+                    current.db(sync_resource_filter_table.id == row.id).delete()
```
nursix: Maybe you shouldn't repeat "current.db" - but bring it into scope at an appropriate point outside of the loop (maybe even outside of the "if"s if multiple of them can apply).

nursix: Note that we do normally *not* delete.
You should remove the task using S3Resource.delete(), and rely on the cascade.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+                  
+            if "sync_task_id" in properties:
+                sync_task_table = current.s3db.sync_task
+                query = (sync_task_table.owned_by_user == current.auth.user.id) & \
+                        (sync_task_table.repository_id == properties["repository_id"])
+                current.db(query).delete()
```
nursix: Same here - better use S3Resource.delete and rely on the cascade.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+            if (len(removed_filter_ids) > 0):
+                for removed_filter_id in removed_filter_ids:
+                    db(sync_resource_filter_table.id == removed_filter_id).delete()
+                resource_filter_id_list = current_filter_ids
+        else:
+            query = sync_resource_filter_table.owned_by_user == current.auth.user.id
+            rows = db(query).select(sync_resource_filter_table.id,
+                                    orderby=sync_resource_filter_table.id)
+            for row in rows:
+                db(sync_resource_filter_table.id==row.id).delete()
```
nursix: That is not how we should do it. Use S3Resource.delete

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+                # Start the scheduler
+                db(query).update(status = "QUEUED",
+                                 next_run_time = utcnow + timedelta(0,300))
+            async_task_id = json.loads(properties)["async_task_id"]
+            pass        
+        else:
+            # First Run: Start schedule_task
+            async_task_id = current.s3task.schedule_task("sync_synchronize",
+                                vars={"repository_id":selected_repository_id},
+                                period=300,  # seconds
```
nursix: Really?

nursix: I mean: really hardcoding it?

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+                # Update
+                db(sync_task_table.repository_id == old_repository_id).\
+                                update(repository_id = selected_repository_id)
+            sync_task_id = row.id        
+        else:
+            # First Run
+            sync_task_data = {"repository_id": selected_repository_id,
+                              "resource_name": resource,
+                              "mode": 2, #Push
+                              "strategy": ["create"],
```
nursix: Updated alerts wouldn't get synchronized? Ok...

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+            if "async_task_id" in properties:
+                repo_id = properties["repository_id"]
+                # Remove the scheduler_task
+                scheduler_table = current.db.scheduler_task
+                vars = '{"repository_id": %s, "user_id": %s}' % (repo_id, current.auth.user.id)
+                query = (scheduler_table.function_name == "sync_synchronize") & \
+                        (scheduler_table.vars == vars )
+                current.db(query).delete()
+                query_ = (current.s3db.sync_job.repository_id == repo_id)
+                current.db(query_).delete()
```
nursix: Again - should use S3Resource.delete.

Note: when using S3Resource.delete, deletion of the scheduler_task entries should go last.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+                sync_task_table = current.s3db.sync_task
+                query = (sync_task_table.owned_by_user == current.auth.user.id) & \
+                        (sync_task_table.repository_id == properties["repository_id"])
+                current.db(query).delete()
+                
+            if "async_task_id" in properties:
+                repo_id = properties["repository_id"]
+                # Remove the scheduler_task
+                scheduler_table = current.db.scheduler_task
+                vars = '{"repository_id": %s, "user_id": %s}' % (repo_id, current.auth.user.id)
```
nursix: Safer (and faster) to use the scheduler_task_id from the respective sync_jobs.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+                sync_job_data = {"repository_id": selected_repository_id,
+                                 "scheduler_task_id": async_task_id,
+                                 }
+                s3db.sync_job.insert(**sync_job_data)
+            else:
+                async_task_id = None
+            
+        # Dump the data
+        properties = json.loads(properties)
+        properties.update({"sync_task_id": sync_task_id})
```
nursix: I think this is unnecessarily complicated.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+        if len(filters)>0:           
+            if properties and \
+               "resource_filter_ids" in json.loads(properties):
+                resource_filter_id_list = json.loads(properties) \
+                                            ["resource_filter_ids"]
+            else:
+                resource_filter_id_list = []
+                
+            for filter in filters:
+                key = str(filter[0].split("__belongs",1)[0]).strip("[]")
```
nursix: ? this needs a comment

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
...
+                              "last_push": utcnow, # since used for notifications,
+                                                   # so don't send old alerts
+                              }
+            sync_task_id = sync_task_table.insert(**sync_task_data)        
+        
+        
+        # Sync Resource Filter
+            
+        # Filter None List
+        filters = json.loads(filters)
```
nursix: This entire block is very difficult to understand - what does it do? Can you add comments so that it becomes clear what it is good for? There may be people after you trying to maintain this - please explain it to them.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +715,230 @@ def _update_subscription(self, subscription):
         subscription["filter_id"] = filter_id
         return subscription
 
+    # -------------------------------------------------------------------------
+    def _update_sync(self, resource, filters, selected_repository_id, properties):
+        """ Update synchronization settings """
```
nursix: Parameter description?

modules/s3/sync_adapter/ftp.py
flavour: I think it's worth keeping this but commented out...

flavour:
> I fundamentally disagree with "S3ResourceLink" - the name was chosen to clearly document the          > purpose of this class, and I don't want to hide it for the sake of a single edge-case. You can still add a   > comment to document the deviation, or assign an alternative class in-place, if you think it would            > otherwise be confusing, but not rename everywhere else where "S3AddResourceLink" gives clarity and > makes the code more understandable.

hmm...we started with a subclass - but it didn't make sense to me since it was so similar.
We /could/ have an S3ResourceLink base class & then have Add & Update as subclasses, but it really seems overkill to me.
We can have S3AddResourceLink as an alias to S3ResourceLink since that defaults to m="create" & I suggested that to avoid breaking non-trunk templates/modules.

modules/templates/SAMBRO/controllers.py
flavour: Do you have an alternative idea?
It comes from needing to have the UX for the sync be the same as subscription

modules/templates/SAMBRO/controllers.py
flavour: Alert updates are always done as extra info elements

modules/templates/SAMBRO/controllers.py
flavour: Hardcoding?

nursix:
söndagen den 27 september 2015 12.17.57 skrev  Fran Boon:
> hmm...we started with a subclass - but it didn't make sense to me since it
> was so similar. We /could/ have an S3ResourceLink base class & then have
> Add & Update as subclasses, but it really seems overkill to me. We can have
> S3AddResourceLink as an alias to S3ResourceLink since that defaults to
> m="create" & I suggested that to avoid breaking non-trunk
> templates/modules.
What for?

You can well use the S3AddResourceLink class, and override the method.

My point is about the *name* of the class, not about the overriding of the 
method - that part is alright.

But using it for an "update" link is an edge case - the normal standard use-
case is to use it for *adding* a new record, and that is what the class name 
should say.

If in one case it is used for *update*, then that one case can either use 
S3AddResourceLink(method="update") and explain that in a comment, or, if by 
all means it has to be another class name, then import it under an alias.

The "Add" in the class name is what I am after - I want this to be left in 
place in order to make clear what it does.

modules/templates/SAMBRO/controllers.py
nursix: Yeah, obviously:

    properties["sync_task_id"] = sync_task_id

flavour:
> The "Add" in the class name is what I am after - I want this to be left in
place in order to make clear what it does.

For the myriad cases which use this, I'm happy for them to use this alias...but I think the core class should be renamed to allow the use for both create & update.
Update is indeed an edge case right now, but I can see it being reused & I see it as very confusing if it's called Add there...

modules/templates/SAMBRO/controllers.py
nursix: Worth a comment to code maintainers?

modules/s3/sync_adapter/ftp.py
nursix: A ToDo should be a wee bit more specific than that, that's what I mean.

+1 to commenting the branch, though.

nursix:
Right - it is the "Add" in the places where it is used, not in the class definition. But there it is important to really express what the code does *especially* if it can do multiple things. 

S3ResourceLink is generic, and for people not so familiar with the code, it is not obvious what it would do (again, in the places where it is /used/ - not where the class is defined).

nursix:
Btw, in my opinion "S3ResourceLink" is a totally misleading name for this class.

You could name it S3PopupLink, which makes it a whole world clearer - and then maybe have S3CreatePopupLink and S3UpdatePopupLink as aliases.

But S3ResourceLink is nonsense, really.

nursix:
S3AddResourceLink wasn't really my favorite name either - but at least the "Add" made clear the purpose, somehow.

S3PopupLink would make it a lot better, especially because it makes the connection to s3.popup.js.

nursix:
And with S3PopupLink, maybe we don't need the "Add" alias, if that is the default behavior.

flavour:
I'm fine with PopupLink...but feel we should keep the S3AddResourceink alias to avoid breaking non-trunk Templates/Modules

nursix:
Backwards-compatibility is important too, yes. 

I'm using S3AddResourceLink in non-trunk code too - and in a feature branch for a new module I'm currently working on. Though I can deal with that - but maybe others can't as easily.

However, my main concern is that the new name should not be misleading, or become too generic so that you can no longer understand the purpose. People who try to learn from existing models need to be able to make the connection, and "S3AddResourceLink" was already weak - but "S3ResourceLink" is even weaker. "S3PopupLink" is a lot stronger, especially because the corresponding JS script also uses "popup" in the name.

modules/templates/SAMBRO/controllers.py
nursix: Hardcoding the sync interval.

I highly doubt that new alerts would be created every 5 minutes - if that is really the case, then the recipients would be overwhelmed anyway. Who could possibly respond at that rate?

The actual idea may be to get new alerts out with a maximum delay of 5 minutes - but: wouldn't it make a lot more sense to trigger synchronize() onaccept of a new alert rather than running on a fixed schedule?

nursix: ...yes, you can trigger synchronize for a repository directly:

    task_id = s3task.async("sync_synchronize",
                           args = [repository.id],
                           vars = {"user_id": auth.user.id, "manual": True},
                           )

flavour: Had no idea you meant the interval! ;)
Yes should be sent ASAP...this is fundamentally notify & here we're talking about alerts...

nursix: ...although it might be better to implement a new task that can synchronize multiple repositories, so that you don't need to schedule one task per repo (seems unnecessary). That task could start immediately onaccept, and update all CAP target repos.

That way, as long as there is no new alert, nothing will happen at all.

PR #1162, opened 2015-09-28
SAMBRO: FTP Sync

biplovbhandari:


modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+                filename = Template(filename).safe_substitute(s="%(systemname_short)s",
+                                                              r="%(resource)s")
+                filename = filename % placeholders
+                
+                # Get Representation
+                representation = task.representation
+                filename = ("%s.%s") % (filename, representation)
+                
+                remote = True
+                import ftplib
```
flavour: I would be tempted to add ftplib as self.ftplib after importing during login
I'm not sure if this really makes any difference though ;)

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+        if request.transformable():
+            return resource.export_xml(stylesheet = request.stylesheet(),
+                                       pretty_print = True,
+                                       )
+        
+        else:            
+            if representation == "csv":
+                exporter = S3Exporter().csv
+                
+            # @ToDo either use CRUD or via URL
```
flavour: I'm sure it will be CRUD not URL ;)

modules/s3db/sync.py
```diff
@@ -283,13 +287,17 @@ def model(self):
         # Reusable Fields
         sync_repository_represent = S3Represent(lookup=tablename)
         repository_id = S3ReusableField("repository_id", "reference %s" % tablename,
+                                        comment = S3PopupLink(c="sync", f="repository",
+                                                              label=ADD_REPOSITORY,
+                                                              title=ADD_REPOSITORY,
```
flavour: Remember that Dominic already said that if title == label, means don't bother with title definition

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+                              mode = log.PUSH,
+                              action = "send",
+                              remote = remote,
+                              result = result,
+                              message = message,
+                              )
+                # Quit here
+                remote_ftp_connection.quit()
+                result = log.SUCCESS
+                message = "Quit Connection"
```
nursix: Superfluous log message, really.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+                                                             StringIO(data))
+                        except ftplib.error_perm:
+                            message = sys.exc_info()[1]
+                            result = log.ERROR
+                            output = message
+                        else:
+                            message = "data sent successfully"
+                            result = log.SUCCESS
+                        # Log each operation                        
+                        log.write(repository_id = repository.id,
```
nursix: DRY DRY DRY - log one message per push, not many.

If you need debug information, write to current.log.debug - in the sync log, this is irrelevant.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+            try:
+                remote_ftp_connection.login(username, password)
+            except ftplib.error_perm:
+                error = sys.exc_info()[1]
+                result = log.ERROR
+            else:
+                result = log.SUCCESS
+        
+        # Log the operation
+        log.write(repository_id=repository.id,
```
nursix: Not needed - S3Sync will write a log entry if there was an error, otherwise we do not log the login-action separately.

Just return the error message if there was an error, otherwise none.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+                  transmission=log.OUT,
+                  mode=log.PUSH,
+                  action="Login",
+                  remote=remote,
+                  result=result,
+                  message=error,
+                  )
+        
+        if error:
+            current.log.error(error)
```
nursix: That's not needed either - if there was an error, S3Sync will log it. If there was no error, we don't log anything.

If you need debug info, use current.log.debug.

modules/s3/sync_adapter/ftp.py
nursix: No it doesn't. Faster to read names from module than from self.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+                                                         StringIO(data))
+                    except ftplib.error_perm:
+                        message = sys.exc_info()[1]
+                        result = log.ERROR
+                        output = message
+                    else:
+                        message = "data sent successfully"
+                        result = log.SUCCESS
+                    # Log the operation                        
+                    log.write(repository_id = repository.id,
```
nursix: Way too much logging.

We want 1 log entry per push, not many - just to see whether the push was successful or not, and if not then why. Not logging every step in the process - that's what you can use current.log.debug for.

modules/templates/SAMBRO/config.py
```diff
@@ -205,6 +215,58 @@ def customise_org_organisation_resource(r, tablename):
...
+        def onaccept(form):
+            # Normal onaccept
+            s3db.sync_repository_onaccept(form)
+            # Set the owned_by_user for the ftp sync
+            db = current.db
+            table = db.sync_repository
+            query = (table.id == form.vars.id) & \
+                    (table.apitype == "ftp") & \
+                    (table.deleted != True)
+            db(query).update(owned_by_user = current.auth.user.id)
```
nursix: This is actually not necessary - owned_by_user will automatically be the user who created the record.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                query = (sync_job_table.deleted != True) &\
+                        (sync_job_table.repository_id == repo_id)
+                # Remove the scheduler_task
+                row = db(query).select(sync_job_table.scheduler_task_id,
+                                       limitby=(0, 1)).first()
+                scheduler_table = db.scheduler_task
+                query = (scheduler_table.id == row.scheduler_task_id)
+                db(query).delete()
+                # Delete the sync_job record
+                job_resource = s3db.resource("sync_job", filter=query)
```
nursix: Wrong query.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+            properties = json.loads(properties)
+            db = current.db
+            s3db = current.s3db
+                  
+            if "sync_task_id" in properties:
+                sync_task_table = s3db.sync_task
+                query = (sync_task_table.deleted != True) &\
+                        (sync_task_table.owned_by_user == current.auth.user.id) & \
+                        (sync_task_table.repository_id == properties["repository_id"])
+                resource = s3db.resource("sync_task", filter=query)
```
nursix: Use an S3ResourceQuery here:

    query = (FS("owned_by_user") == current.auth.user.id) & (FS("repository_id") == properties["repository_id"])

deleted != True is not needed with S3Resource - it will never contain deleted records unless you ask for it explicitly.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+            # Runned in onapprove cap_alert 
+            async_task_id = current.s3task.schedule_task("sync_synchronize",
+                                                args = [selected_repository_id],
+                                                vars={"user_id": user_id},
+                                                )
+            if async_task_id:
+                sync_job_data = {"repository_id": selected_repository_id,
+                                 "scheduler_task_id": async_task_id,
+                                 }
+                s3db.sync_job.insert(**sync_job_data)
```
nursix: In principle, you would have to call update_super, set_record_owner and onaccept when you insert records in a table. 

However, there is currently no onaccept and no super-entity - so it's just set_record_owner you need to call.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                        current_filter_ids.append(resource_filter_id)
+                                         
+            removed_filter_ids = list(set(resource_filter_id_list) - set(current_filter_ids))
+            
+            if (len(removed_filter_ids) > 0):
+                query = sync_resource_filter_table.id.belongs(removed_filter_ids)
+                resource = s3db.resource("sync_resource_filter", filter=query)
+                resource.delete()
+        else:
+            query = sync_resource_filter_table.owned_by_user == user_id
```
nursix: Better use FS("owned_by_user") == user_id

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                            resource_filter_id_list.append(resource_filter_id)
+                            current_filter_ids.append(resource_filter_id)                    
+                    else:
+                        # First Run                       
+                        resource_filter_data = {"task_id": sync_task_id,
+                                                "tablename": tablename,
+                                                "filter_string": filter_string,
+                                                "modified_on": utcnow,
+                                                }
+                        resource_filter_id = sync_resource_filter_table.\
```
nursix: Not calling set_record_owner here nor onaccept here :/

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                               row.filter_string == filter_string:
+                                current_filter_ids.append(row.id)
+                                break
+                        else:
+                            resource_filter_data = {"task_id": sync_task_id,
+                                                    "tablename": tablename,
+                                                    "filter_string": filter_string,
+                                                    "modified_on": utcnow,
+                                                    }
+                            resource_filter_id = sync_resource_filter_table. \
```
nursix: Not calling set_record_owner nor onaccept - but there is an onaccept to be called!

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                    tablename = "gis_location"
+                    
+                values = value_.split(",")
+    
+                for value in values:
+                    filter_string = "%s=%s" % (query_part, value)
+                    if properties and \
+                       "resource_filter_ids" in json.loads(properties):
+                        rows = db(sync_resource_filter_table.id.belongs\
+                                 (resource_filter_id_list)).select()
```
nursix: Specify the fields you want to select

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                        for row in rows:
+                            if row.owned_by_user == user_id and \
+                               row.task_id == sync_task_id and \
+                               row.tablename == tablename and \
+                               row.filter_string == filter_string:
+                                current_filter_ids.append(row.id)
+                                break
+                        else:
```
nursix: "for" has no "else"

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+        # Sync Resource Filter
+            
+        # Normally a filter looks like this
+        # [["event_type_id__belongs","24,3"],["priority__belongs",null]]
+        # Get only those that have value and ignore null one
+        filters = json.loads(filters)
+        filters = [filter for filter in filters if filter[1] is not None]
+        current_filter_ids = []
+        sync_resource_filter_table = s3db.sync_resource_filter
+        if len(filters)>0:
```
nursix: blanks around operators

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+        # Normally a filter looks like this
+        # [["event_type_id__belongs","24,3"],["priority__belongs",null]]
+        # Get only those that have value and ignore null one
+        filters = json.loads(filters)
+        filters = [filter for filter in filters if filter[1] is not None]
+        current_filter_ids = []
+        sync_resource_filter_table = s3db.sync_resource_filter
+        if len(filters)>0:           
+            if properties and \
+               "resource_filter_ids" in json.loads(properties):
```
nursix: parsing the JSON twice?? Why?

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+            sync_task_id = sync_task_table.insert(**sync_task_data)        
+        
+        
+        # Sync Resource Filter
+            
+        # Normally a filter looks like this
+        # [["event_type_id__belongs","24,3"],["priority__belongs",null]]
+        # Get only those that have value and ignore null one
+        filters = json.loads(filters)
+        filters = [filter for filter in filters if filter[1] is not None]
```
nursix: "filter" is a reserved Python keyword - use something else

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+        query = (sync_task_table.deleted != True) & \
+                (sync_task_table.owned_by_user == user_id) & \
+                (sync_task_table.repository_id == old_repository_id)
+                
+        row = db(query).select(sync_task_table.id,
+                               sync_task_table.repository_id,
+                               limitby=(0, 1)).first()
+        if row:
+            # Check if update?
+            if row.repository_id == selected_repository_id:
```
nursix: Try with:

    if row.repository_id != selected_repository_id:

There is no reason to have a pass under "if", and then an else-branch

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+            old_repository_id = json.loads(properties)["repository_id"]
+        else:
+            # First Run
+            old_repository_id = None
+        
+        # Insert/Update repository_id    
+        if old_repository_id:
+            if old_repository_id == selected_repository_id:
+                # Do Nothing
+                pass
```
nursix: Try to negate the condition?

There is no reason to have a pass under "if" (but then have an else-branch).

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,323 @@
...
+        url = repository.url
+        
+        import ftplib                        
+        try:
+            self.remote_ftp_connection = ftplib.FTP(url)
+        except ftplib.all_errors:
+            error = sys.exc_info()[1]
+            result = log.ERROR 
+        else:
+            remote_ftp_connection = self.remote_ftp_connection
```
nursix: This is quite funny, isn't it?

In the try you set self.x, just to then bring it back into scope here. Why not:

    try:
        ftp_connection = ftplib.FTP(url)
    except ftplib.all_errors:
        ....
        ftp_connection = None
    else:
        ....
    
    self.remote_ftp_connection = ftp_connection

Bit easier to grasp...

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+            resource.delete()
+            resource_filter_id_list = None
+                            
+        # Initialize Scheduler
+        
+        # Already has the job for the repository_id
+        if properties and \
+           "async_task_id" in json.loads(properties):
+            async_task_id = json.loads(properties)["async_task_id"]
+            pass
```
nursix: pass? Forgot to remove?

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+        else:
+            # First Run: Initialize, don't start
+            # Runned in onapprove cap_alert 
+            async_task_id = current.s3task.schedule_task("sync_synchronize",
+                                                args = [selected_repository_id],
+                                                vars={"user_id": user_id},
```
nursix: consistent blanks around = if multi-line.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                                                }
+                        resource_filter_id = sync_resource_filter_table.\
+                                                insert(**resource_filter_data)
+                        resource_filter_id_list.append(resource_filter_id)
+                        current_filter_ids.append(resource_filter_id)
+                                         
+            removed_filter_ids = list(set(resource_filter_id_list) - set(current_filter_ids))
+            
+            if (len(removed_filter_ids) > 0):
+                query = sync_resource_filter_table.id.belongs(removed_filter_ids)
```
nursix: Simpler:

    resource = s3db.resource("sync_resource_filter", id=removed_filter_ids)

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+                                                "modified_on": utcnow,
+                                                }
+                        resource_filter_id = sync_resource_filter_table.\
+                                                insert(**resource_filter_data)
+                        resource_filter_id_list.append(resource_filter_id)
+                        current_filter_ids.append(resource_filter_id)
+                                         
+            removed_filter_ids = list(set(resource_filter_id_list) - set(current_filter_ids))
+            
+            if (len(removed_filter_ids) > 0):
```
nursix: In Python, you don't need outer brackets around conditions.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+        sync_resource_filter_table = s3db.sync_resource_filter
+        if len(filters)>0:           
+            if properties and \
+               "resource_filter_ids" in json.loads(properties):
+                resource_filter_id_list = json.loads(properties) \
+                                            ["resource_filter_ids"]
+            else:
+                resource_filter_id_list = []
+                
+            for filter in filters:
```
nursix: I would still want to have a comment here explaining what this block does (=splitting the belongs-filter into a series of = queries), and why (yeah: why? what is this good for?)

modules/templates/SAMBRO/config.py
```diff
@@ -205,6 +215,58 @@ def customise_org_organisation_resource(r, tablename):
...
+        
+        s3db = current.s3db
+        def onapprove(record):
+            # Normal onapprove
+            s3db.cap_alert_approve(record)
+            
+            db = current.db
+            table = db.scheduler_task
+            
+            query = (table.task_name == "sync_synchronize")
```
nursix: Why still the hassle with sync_job and scheduler_task?

Why not call repository.synchronize directly from a separate async task for the relevant repositories?

It is not required to have a sync_job for a repository - it's merely an /option/ if you want to run synchronizations with regular time intervals. But you don't want to sync with regular time intervals - you only want to sync when a new alert gets approved - so best to call an async task here onapprove.

flavour: Yup, exactly what I said :)

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +718,230 @@ def _update_subscription(self, subscription):
...
+        
+        # Already has the job for the repository_id
+        if properties and \
+           "async_task_id" in json.loads(properties):
+            async_task_id = json.loads(properties)["async_task_id"]
+            pass
+        else:
+            # First Run: Initialize, don't start
+            # Runned in onapprove cap_alert 
+            async_task_id = current.s3task.schedule_task("sync_synchronize",
```
nursix: Why not just async a task to push all relevant repositories once onapprove - it doesn't quite make sense to me to create a schedule if you don't need a schedule.

The point with scheduled tasks is that they are repeated automatically after a certain amount of time - but you don't want it to be repeated automatically. You want the FTP repositories to be updated exactly once when an alter gets approved.

So - write a task function cap_ftp_sync in models/tasks.py, which finds the relevant repositories and calls repository.synchronize() for each of them. Then call this function like current.s3task.async("cap_ftp_sync", ...)

And remove all the sync_job and scheduler_task stuff.

No?

modules/templates/SAMBRO/controllers.py
```diff
@@ -370,30 +373,108 @@ def _manage_subscriptions(self, resources, filters):
...
+            if properties and json.loads(properties)["repository_id"]:            
+                user_repository_id = json.loads(properties)["repository_id"]            
+            else:
+                user_repository_id = ftp_rows.first().id
+            
+            UPDATE_REPOSITORY = T("Update Repository")
+            
+            if current.auth.s3_has_permission("update", "sync_repository", 
+                                              record_id = user_repository_id):
+                repository_id.comment = S3PopupLink(c="sync", f="repository",
```
nursix: repository_id is an S3ReusableField - it does not have a "comment" attribute. 

You set such an attribute here, just to read it again when you create the form row a few lines further down - but for that you do not need "repository_id" /at all/. You can just make this a local variable "repository_comment".

There is no other purpose of bringing repository_id into scope here, so that's completely unnecessary.

nursix: ...and wrong, actually. You're manipulating a nonexistent attribute of a template object, both of which isn't really allowed as it can lead to terrible regressions. This must be fixed.

nursix:
Thanks for updating it - but not quite ready yet, I'm afraid.

No point to review this further until you've resolved the issues indicated so far - will continue after the next update.

modules/templates/SAMBRO/controllers.py
nursix: (you are overwriting the original job query - always careful when re-using variables ;) )

PR #1164, opened 2015-09-28
add status field to cases

tirgil:


modules/s3db/dvr.py
```diff
@@ -136,13 +137,13 @@ def model(self):
...
-                     #      default = 1,
-                     #      label = T("Status"),
-                     #      represent = lambda opt: \
-                     #           dvr_status_opts.get(opt, UNKNOWN_OPT),
-                     #      requires = IS_EMPTY_OR(IS_IN_SET(dvr_status_opts)),
-                     #      ),
+                     Field("status", "integer",
+                           default = 1,
+                           label = T("Status"),
+                           represent = lambda opt: \
```
nursix: Tip:

represent = S3Represent(options=dvr_status_opts),

Saves the innermost loop when representing many rows ;)

flavour: New to me :)
Many, many examples that could be improved this way then - ideal beginner volunteer task I guess

nursix: No big win performance-wise, but a lot easier to type ;)

PR #1165, opened 2015-09-29
FTP Sync Adapter

biplovbhandari:


modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,284 @@
...
+                  resource_name = task.resource_name,
+                  transmission = log.OUT,
+                  mode = log.PULL,
+                  action = None,
+                  remote = False,
+                  result = log.FATAL,
+                  message = "Pull from FTP currently not supported",
+                  )
+
+        output = current.xml.json_message(False, 400, message)
```
nursix: Just the error, no JSON message.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,284 @@
...
+        log = repository.log
+        remote = False
+        output = None
+        
+        current.log.debug("S3SyncRepository.push(%s, %s)" % (repository.url,
+                                                             resource_name))
+        
+        # Define the resource
+        resource = current.s3db.resource(resource_name,
+                                         include_deleted=True)
```
nursix: Hmm, include_deleted=True would only make sense if you would indeed delete remote records when they have been deleted locally - but you're not doing that.

Additionally, you would have to follow task.strategy. Only if it contains "delete", then you are allowed to delete remote records. Same for "create" and "update" strategies.

Right now, you export all records (both deleted and not deleted), and you push all records which have been modified (any of created, deleted or updated) regardless of task.strategy. In SAMBRO, where you set msince manually, this might work. But for a generic adapter, this is wrong of course.

Fine with a ToDo for now, though. But re-consider include_deleted.

modules/templates/SAMBRO/controllers.py
```diff
@@ -616,4 +720,120 @@ def _update_subscription(self, subscription):
...
+                              }
+            sync_task_id = sync_task_table.insert(**sync_task_data)       
+            auth.s3_set_record_owner(sync_task_table, sync_task_id)
+            old_sync_id = properties["sync_task_id"] = sync_task_id
+        
+        # Sync Resource Filter
+        
+        # Remove Old Filter and create new
+        query = (FS("task_id") == old_sync_id)
+        s3db.resource("sync_resource_filter", filter=query).delete(cascade=True)
```
nursix: No cascade=True please!

cascade=True is to indicate recursion (of resource.delete itself!), so that it can rollback this part of the cascade in case another part of the cascade fails. But there is no such exit in this function, and you do write the new filters unconditionally, so no point to keep this transaction open.

nursix:
Makes a lot more sense now - apart from the small bits indicated above.

Now it needs testing, and it would be ideal if you could implement a bunch of unit tests for it. But I guess now I'm asking too much :/

modules/s3/sync_adapter/ftp.py
nursix: Actually - no: you can not just make it a ToDo. You /must/ follow strategy.

SAMBRO specifies task.strategy=["create"] which means:

1) Only if strategy contains "update" or "delete" then you can use modified_on > mtime. If strategy is only "create", you must use created_on > mtime.
2) include_deleted must be False unless task.strategy includes "delete" *and* you have the capability to delete remote records.

So, this must be fixed - not just ToDo.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,284 @@
...
+        tablename = resource.tablename
+        
+        if filters:
+            queries = S3URLQuery.parse(resource, filters[tablename])
+            [resource.add_filter(q) for a in queries for q in queries[a]]
+
+        # Filter to records after last push
+        msince = task.last_push            
+        if msince is not None and "modified_on" in table.fields:
+            resource.add_filter(table.modified_on > msince)
```
nursix: Here: must use created_on if strategy is only "create". Only if strategy includes "update" or "delete" you may use modified_on.

nursix: This is important even with manual mtime, because otherwise you create a race-condition. Records could be updated before the async task goes out, and then you end up with old records whilst you only want to send new ones (=violates the strategy).

biplovbhandari:
@flavour Please merge

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,284 @@
...
+                  mode = log.PUSH,
+                  action = "send",
+                  remote = remote,
+                  result = result,
+                  message = message,
+                  )
+        # Returns after operation is complete
+        if output is not None:
+            mtime = None
+        return(output, mtime)
```
nursix: Btw - in Python, we usually write:

    return output, mtime

...but at least a blank between the "return" and the value (return is not a function)

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,284 @@
...
+        table = resource.table
+        tablename = resource.tablename
+        
+        if filters:
+            queries = S3URLQuery.parse(resource, filters[tablename])
+            [resource.add_filter(q) for a in queries for q in queries[a]]
+
+        # Filter to records after last push
+        msince = task.last_push            
+        if msince is not None and "created_on" in table.fields:
```
nursix: Hmm - still not following task.strategy :/

I was thinking of something like:

    if msince:
        strategy = task.stragegy
        updated = "update" in strategy
        created = "create" in stratgegy
        if udpated and created:
            mtime_filter = table.modified_on > msince
        elif created:
            mtime_filter = table.created_on > msince
        elif updated:
            mtime_filter = (table.created_on <= msince) & (table.modified_on > msince)
        resource.add_filter(mtime_filter)

...but of course you write a nicer one.

modules/s3/sync_adapter/ftp.py
```diff
@@ -0,0 +1,284 @@
...
+        log = repository.log
+        remote = False
+        output = None
+        
+        current.log.debug("S3SyncRepository.push(%s, %s)" % (repository.url,
+                                                             resource_name))
+        
+        # Define the resource
+        resource = current.s3db.resource(resource_name,
+                                         include_deleted=False)
```
nursix: This is the default ;) Better try a useful comment:

    resource = current.s3db.resource(resource_name,
                                     # FTP remote deletion not supported yet
                                     #include_deleted = True
                                     )

modules/s3/sync_adapter/ftp.py
nursix: Note that this is specific for this adapter as FTP has no update strategy, so the export must take care of it (normally, the importer would).

biplovbhandari:
@flavour Please merge

nursix:
Looks ok to me - @flavour to do his own review, of course.

Does it work?

biplovbhandari:
Yes :-)

nursix:
Good work then :) thank you.

biplovbhandari:
@nursix Couple of problems:

1. When I add repo, it does not seem to insert into the "last_push" field as I have specified here https://github.com/flavour/eden/pull/1165/files#diff-8b02f3a5c97483e605769a048020941cR780

2. When I add FTP repo and disselect it, it removes the repo as intended https://github.com/flavour/eden/pull/1165/files#diff-8b02f3a5c97483e605769a048020941cR837, but if I add again the "owned_by_user" field does not gets updated with the auth.user.id but the deleted_fk field has attribute as {"owned_by_user": xx, "modified_by": xx, "created_by": xx} (deleted field is False ofcourse again after creating it). Do I need to set is_owned_by_user onaccept?

What am I missing here?

nursix:
onsdagen den 30 september 2015 01.36.43 skrev  Biplov Bhandari:
> @nursix Couple of problems:
> 
> 1. When I add repo, it does not seem to insert into the "last_push" field as
> I have specified here
> https://github.com/flavour/eden/pull/1165/files#diff-8b02f3a5c97483e605769a
> 048020941cR780

I think that is because you end up here:
https://github.com/flavour/eden/pull/1165/files#diff-8b02f3a5c97483e605769a048020941cR764

> 2. When I add FTP repo and disselect it, it removes the repo as intended
> https://github.com/flavour/eden/pull/1165/files#diff-8b02f3a5c97483e605769a
> 048020941cR837, but if I add again the "owned_by_user" field does not gets
> updated with the auth.user.id but the deleted_fk field has attribute as
> {"owned_by_user": xx, "modified_by": xx, "created_by": xx} (deleted field
> is False ofcourse again after creating it). Do I need to set
> is_owned_by_user onaccept?
I'm not sure I understand that. 

When you delete a repo - how can you re-add it? I don't see code that sets the 
deleted-flag to False, so how are you doing it so that deleted is False, yet 
there are deleted_fk?

And no - you don't touch owned_by_user directly, but always call 
set_record_owner.

Dominic

nursix:
onsdagen den 30 september 2015 01.36.43 skrev  Biplov Bhandari:
> 2. When I add FTP repo and disselect it, it removes the repo as intended
> https://github.com/flavour/eden/pull/1165/files#diff-8b02f3a5c97483e605769a
> 048020941cR837, but if I add again the "owned_by_user" field does not gets
> updated with the auth.user.id but the deleted_fk field has attribute as
> {"owned_by_user": xx, "modified_by": xx, "created_by": xx} (deleted field
> is False ofcourse again after creating it). Do I need to set
> is_owned_by_user onaccept?
From your code, I don't see that the same repository record is being re-used.

I can't believe that deleted=False in that record. How would that be possible? 
Once a record has been deleted, and deleted_fk has been created, you would 
have to intentionally revert that, i.e. actively set deleted=False - but 
nowhere in your code I can see that happening.

Could it be that you are looking at the deleted record while your code creates 
a new record that you are not aware of?

Dominic

biplovbhandari:
@nursix Let me test again

nursix:
onsdagen den 30 september 2015 01.36.43 skrev  Biplov Bhandari:
> What am I missing here?
NB You should run again with a fresh database. 

It could well be that after running the old version of your commit (which 
didn't quite follow framework rules), the database is now corrupt and hence 
the strange behavior.

Wipe the db, and verify again?

Dominic

nursix:
Ah no - you are right.

I guess you are using the same UUID when re-creating the repo? That indeed makes it an update rather than a create, and wouldn't re-establish the record owner.

Let me see whether I can fix that.

biplovbhandari:
@nursix The first part (last_push insertion) is working...

Yes for second one, for recreating I am giving same UUID, so it basically updates deleted field to False.

First Creation: http://i.imgur.com/ZNVWGtg.jpg

After deletion->Creation: http://i.imgur.com/cweWBt5.jpg

nursix:
Yes, I can see why.

This is an edge-case where a record update should be treated as a create in CRUD, but isn't. I'll fix it.

nursix:
The background for this problem is:

A matching UUID (or other unique key) in a CRUD form would reinstate a previously deleted record rather than creating a new one. However, this was (wrongly) treated as "update" in forms, instead of treating it properly as "create":
https://github.com/nursix/eden/commit/1548edc282263e8f43bc81dd6954baee1761ec52

"Update", though, does not set the record owner again.

This is a rare edge-case, however - because the uuid field is editable. However, the uuid is not required for passive repositores - so in the FTP case you could actually hide it.

And if you just leave this field empty when creating a repository, then it will be automatically populated with a real uuid, and then the problem won't occur. 

Well, ok - with this fix ^ it won't occur either way.

biplovbhandari:
Done :) Thanks @nursix so much

@flavour Please merge

PR #1168, opened 2015-10-07
SAMBRO: color coded cap area

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/config.py
+++ b/modules/templates/SAMBRO/config.py
@@ -262,6 +262,41 @@ def config(settings):
     settings.customise_sync_repository_controller = customise_sync_repository_controller
     
     # -------------------------------------------------------------------------
+    def customise_cap_warning_priority_resource(r, tablename):
+        
+        s3db = current.s3db
+        def onaccept(form):
+            # Normal onaccept if any
+            
+            if form.vars.color_code:
+                db = current.db
+                stable = s3db.gis_style
+                etable = s3db.gis_layer_entity
+                rows_ = db(etable.instance_type == "gis_layer_feature"). \
+                                                            select(etable.id)
+                query = (stable.layer_id.belongs([row_.id for row_ in rows_]))
+                rows = db(query).select(stable.id,
+                                        stable.style,
+                                        orderby=stable.id)
+                if rows:
+                    for row in rows:
+                        if row.style:
+                            sdata = OrderedDict(prop = "priority",
+                                                fill = form.vars.color_code,
+                                                fillOpacity = 0.4,
+                                                cat = form.vars.name
+                                                )                        
+                            if sdata not in row.style:
+                                row.style.append(sdata)
+                                db(stable.id == row.id).update(style = row.style)
+        
+        s3db.configure(tablename,
+                       create_onaccept = onaccept,
+                       )
+        
+    settings.customise_cap_warning_priority_resource = customise_cap_warning_priority_resource        
+        
+    # -------------------------------------------------------------------------
     # Comment/uncomment modules here to disable/enable them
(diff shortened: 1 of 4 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -262,6 +262,41 @@ def custom_prep(r):
...
+        def onaccept(form):
+            # Normal onaccept if any
+            
+            if form.vars.color_code:
+                db = current.db
+                stable = s3db.gis_style
+                etable = s3db.gis_layer_entity
+                rows_ = db(etable.instance_type == "gis_layer_feature"). \
+                                                            select(etable.id)
+                query = (stable.layer_id.belongs([row_.id for row_ in rows_]))
```
flavour: This is confusing as it looks like you are conflating id with layer_id - since the id is from gis_layer_entity they are in fact the same as there is no id field...I would prefer this to be more explicit & actually use the layer_id field directly

modules/templates/SAMBRO/config.py
```diff
@@ -262,6 +262,41 @@ def custom_prep(r):
...
+                rows = db(query).select(stable.id,
+                                        stable.style,
+                                        orderby=stable.id)
+                if rows:
+                    for row in rows:
+                        if row.style:
+                            sdata = OrderedDict(prop = "priority",
+                                                fill = form.vars.color_code,
```
flavour: Bring this outside the loop
(I would bring it outside the top if: )

modules/templates/SAMBRO/config.py
```diff
@@ -262,6 +262,41 @@ def custom_prep(r):
...
+                                                fillOpacity = 0.4,
+                                                cat = form.vars.name
```
flavour: Again, bring this outside the loop

modules/templates/SAMBRO/config.py
```diff
@@ -262,6 +262,41 @@ def custom_prep(r):
...
+                rows_ = db(etable.instance_type == "gis_layer_feature"). \
+                                                            select(etable.id)
+                query = (stable.layer_id.belongs([row_.id for row_ in rows_]))
+                rows = db(query).select(stable.id,
+                                        stable.style,
+                                        orderby=stable.id)
+                if rows:
+                    for row in rows:
+                        if row.style:
+                            sdata = OrderedDict(prop = "priority",
```
flavour: Why does this need to be Ordered?

modules/templates/SAMBRO/config.py
```diff
@@ -262,6 +262,41 @@ def custom_prep(r):
...
+                                                fill = form.vars.color_code,
+                                                fillOpacity = 0.4,
+                                                cat = form.vars.name
+                                                )                        
+                            if sdata not in row.style:
+                                row.style.append(sdata)
+                                db(stable.id == row.id).update(style = row.style)
```
flavour: I'm still worried by this not using IS_JSONS3 validator
SQLite & PostgreSQL behave different;y with the style field since it just uses a text field in SQLite but a real JSON field in PostgreSQL.

modules/templates/SAMBRO/config.py
```diff
@@ -262,6 +262,41 @@ def custom_prep(r):
...
+                etable = s3db.gis_layer_entity
+                rows_ = db(etable.instance_type == "gis_layer_feature"). \
+                                                            select(etable.id)
+                query = (stable.layer_id.belongs([row_.id for row_ in rows_]))
+                rows = db(query).select(stable.id,
+                                        stable.style,
+                                        orderby=stable.id)
+                if rows:
+                    for row in rows:
+                        if row.style:
```
flavour: row.style is getopt many times...I would bring this into scope up here

PR #1173, opened 2015-10-09
CAP:tweaks

biplovbhandari:
@flavour Please merge...

### DIFF
--- a/modules/templates/SAMBRO/menus.py
+++ b/modules/templates/SAMBRO/menus.py
@@ -66,13 +66,19 @@ class S3MainMenu(default.S3MainMenu):
                 return super(S3MainMenu, cls).menu_modules()
             else:
                 # Publisher sees minimal options
-                # @ToDo: Add role check here once role defined
                 menus_ = [homepage(),
-                          homepage("cap"),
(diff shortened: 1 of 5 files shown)


modules/templates/SAMBRO/menus.py
```diff
@@ -128,4 +134,65 @@ def menu_admin(cls, **attr):
...
+                    M("Predefined Alert Area", f="area",
+                      check=s3_has_role("ADMIN"))(
+                        M("Create", m="create"),
+                        M("Import from CSV", m="import", p="create"),
+                    ),
+                    M("RSS Channels", c="msg", f="rss_channel",
+                      check=s3_has_role("ADMIN"))(
+                        M("Create", m="create"),
+                    ),
+                    M("Twitter Channels", c="msg", f="twitter_channel",
```
flavour: I'll merge this as it's a step fwd, but you did you get the UX enhancement suggestions I made for this?
i.e. the fact that there needs to be a login channel
Hence I would filter the login channel from this view & provide a dedicated link to the login channel (we can create this in prepop, even with the entries blank)

PR #1178, opened 2015-10-11
import S3AddResourceLink

tirgil:


modules/s3db/dvr.py
```diff
@@ -33,7 +33,7 @@
 from gluon.storage import Storage
 from gluon.tools import callback
 from ..s3 import *
-from s3layouts import S3PopupLink
+from s3layouts import S3PopupLink, S3AddResourceLink
```
nursix: Deprecated - pls use S3PopupLink instead (drop-in replacement)

nursix: and not used in this file anyway

PR #1180, opened 2015-10-12
CAP: cleanup

biplovbhandari:


static/scripts/S3/s3.cap.js
```diff
@@ -77,22 +77,27 @@
...
-                }
-            }
+        	if (!$(this).val()) {
+        		$(this).css('border', '1px solid gray');
+        	}
+        	else {
+	            var p = S3.cap_priorities,
+	                len = p.length;
+	            if ($(this).find('option:selected').text() == 'Undefined') {
+	                $(this).css('border', '2px solid gray');
```
flavour: In general, I think this kind of thing is better done by adding/removing classes...the actual CSS is in the stylesheet.
Won't hold-up merge though...just a future @ToDo (not urgent)

PR #1181, opened 2015-10-12
CAP: fix key-value widget import + import/export cleanup + removed cap_info_labels function

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -29,7 +29,6 @@
 
 __all__ = ("S3CAPModel",
            "S3CAPAreaNameModel",
-           "cap_info_labels",
            "cap_alert_is_template",
            "cap_rheader",
            "cap_alert_list_layout",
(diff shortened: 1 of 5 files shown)


modules/s3db/cap.py
```diff
@@ -1595,32 +1620,21 @@ def model(self):
...
-                contact=T("Contact information"),
-                parameter=T("Parameters")
-                )
+    
+    if fstring == "||":
+        fstring = "[]"
+    else:
+        fstring = fstring.replace("|", "")
+        fstring = fstring.replace("}{", "},{")
+        fstring = fstring.replace("u'", "'")
```
flavour: I still think there is a bug here.
Surely we don't want u'fou' converted to 'fo' which I think this would do
What we want is 'fou'

biplovbhandari:
@flavour Please merge

modules/s3db/cap.py
```diff
@@ -1595,32 +1620,24 @@ def model(self):
...
+    
+    if fstring == "||":
+        fstring = "[]"
+    else:
+        fstring = fstring.replace(" ", "") # Remove space
+        fstring = fstring.replace("|", "")
+        fstring = fstring.replace("}{", "},{")
+        fstring = fstring.replace("{u'", "{'")
+        fstring = fstring.replace(":u'", ":'")
+        fstring = fstring.replace(",u'", ",'")
```
flavour: That looks safer :)

PR #1182, opened 2015-10-12
Migrated commit_represent to S3Represent

anurag-ks:


modules/s3db/req.py
```diff
@@ -2764,6 +2738,44 @@ def commit_ondelete(row):
...
+    # -------------------------------------------------------------------------
+    def represent_row(self, row):
+        """
+            Represent a single Row
+
+            @param row: the Row
+        """
+        table = current.s3db.req_commit
+        if row.type == 1:
+            return "%s - %s" % (table.site_id.represent(row.site_id),
```
nursix: Makes no difference with S3Represent if you call other represents per-row.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,44 @@ def commit_ondelete(row):
...
+    """Representation of Commit"""
+
+    def lookup_rows(self, key, values, fields=[]):
+        """
+            Custom lookup method
+            @param key: the key Field
+            @param values: the values
+            @param fields: the fields to retrieve
+        """
+        table = current.s3db.req_commit
```
nursix: Blank line after docstring please.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,44 @@ def commit_ondelete(row):
...
+        if len(values) == 1:
+            query = (key == values[0])
+        else:
+            query = key.belongs(values)
+        rows = current.db(query).select(table.type,
```
nursix: Separate logical code blocks by blank line => easier to read and follow the flow

modules/s3db/req.py
nursix: You should call the bulk-method of the same (!) represent instances once during lookup_rows for /all/ values.

That will store the representations in the instance, so that when you call them per-row in represent_row, they wouldn't be looked up again.

modules/s3db/req.py
nursix: Typically, use self.table in lookup_rows

anurag-ks:
@nursix done

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+class req_CommitRepresent(S3Represent):
+    """Representation of Commit"""
+
+    def __init__(self):
+        """
+            Constructor
+        """
+
+        table = current.s3db.req_commit
+        super(req_CommitRepresent, self).__init__()
```
nursix: This constructor does basically nothing - so should be removed.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+            @param fields: the fields to retrieve
+        """
+        
+        if len(values) == 1:
+            query = (key == values[0])
+        else:
+            query = key.belongs(values)
+
+        rows = current.db(query).select(self.table.type,
+                                        self.table.date,
```
nursix: Repeated lookups of self.table - should be brought into scope.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+                                        self.table.organisation_id,
+                                        self.table.site_id,
+                                        limitby=(0, len(values)))
+
+        commit_dates = [row[str(self.table.date)] for row in rows]
+        if commit_dates:
+            self.table.date.represent.bulk(commit_dates)
```
nursix: I don't think date.represent has a bulk-method?

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+        else:
+            query = key.belongs(values)
+
+        rows = current.db(query).select(self.table.type,
+                                        self.table.date,
+                                        self.table.organisation_id,
+                                        self.table.site_id,
+                                        limitby=(0, len(values)))
+
+        commit_dates = [row[str(self.table.date)] for row in rows]
```
nursix: Repeated lookup of str(self.table.date) inside the loop - should be done once outside the loop (it's constant)

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+        if commit_dates:
+            self.table.date.represent.bulk(commit_dates)
+
+        organisation_ids = [row[str(self.table.organisation_id)] for row in rows]
```
nursix: Same - repeated lookup of str(self.table.organisation_id) inside the loop. Should be done once outside the loop.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+        if organisation_ids:
+            self.table.organisation_id.represent.bulk(organisation_ids)
+
+        site_ids = [row[str(self.table.site_id)] for row in rows]
```
nursix: Here too - same repeated lookup of a constant value inside the loop.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
                 # @ToDo: Provide a way for the committer to specify this
                 db(rtable.id == req_id).update(commit_status=REQ_STATUS_NONE)
 
+# -----------------------------------------------------------------------------
+class req_CommitRepresent(S3Represent):
+    """Representation of Commit"""
```
nursix: Blanks between """ and the text pls.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+    # -------------------------------------------------------------------------
+    def represent_row(self, row):
+        """
+            Represent a single Row
+
+            @param row: the Row
+        """
+
+        if row.type == 1:
+            return "%s - %s" % (self.table.site_id.represent(row.site_id),
```
nursix: You should test this - site_id.represent may return a link (unless you suppress it explicitly), and then this string construction is actually not ok, or is it?

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+                                self.table.date.represent(row.date))
+        else:
+            return "%s - %s" % (self.table.organisation_id.represent(row.organisation_id),
```
nursix: Same here - organisation_id.represent may return a link (unless you set show_link=False), which may not work with this type of string construction.

nursix: Subject to testing, of course - I'm not certain it really does.

modules/s3db/req.py
```diff
@@ -2764,6 +2738,66 @@ def commit_ondelete(row):
...
+            @param values: the values
+            @param fields: the fields to retrieve
+        """
+        
+        if len(values) == 1:
+            query = (key == values[0])
+        else:
+            query = key.belongs(values)
+
+        rows = current.db(query).select(self.table.type,
```
nursix: Critically, this lookup does not include table.id - but that is mandatory. All rows must contain the record ID, otherwise S3Represent won't work.

nursix: (How else could we identify which representation belongs to which id?)

This not being here tells me that you did not test it!

nursix:
This is the right concept - but a sloppy implementation of it ;)

You should always *test* your code before submitting a PR - and this code can not work, so you either didn't test it, or you ignored that it doesn't work ;)

anurag-ks:
Actually @nursix I did tried to test it, but I think they way I tested it is wrong, sorry for that.

nursix:
@anurag-ks - generally no problem to ask for feedback on the concept by tagging me in a comment to your commit before submitting a PR. But for a PR, stuff should be working.

You could try to write a unit test for this (even if just as an exercise) - this would give you immediate feedback while refactoring things.

anurag-ks:
@nursix just a doubt, these represent classes/functions are used for custom rendering of data values for data-tables or exporting data, right ? But on enabling the `commit_id` field on `/eden/req/commit` gives me `None`

flavour:
for req/commit there is no 'commit_id' field, there is an 'id' field which in other  places would be seen as a commit_id

flavour:
> these represent classes/functions are used for custom rendering of data values for data-tables or exporting data, right ?

yes

anurag-ks:
Okay so `commit_id` is for components, but it's not there on the data tables.
To be honest I haven't written any unit tests before (only automation stuff), so I was checking out other ways to test it.

nursix:
A unit test would create a test req_commit record, and then run req_CommitRepresent against the record ID. Then check whether the return value is correct, and finally remove the test record.

That's all there is to it, nothing fancy.

nursix:
There are over 500 examples of unit tests in modules/unit_tests and plenty of documentation for the Python unittest module. It's not rocket science - the trick is just to understand what you're trying to prove with the test and implement a valid test.

nursix:
And I said that with the unittest only to spare you the effort of digging for a case where it is visible, because I understood that you couldn't find one. In fact, there may not be one ;)

PR #1183, opened 2015-10-13

biplovbhandari:
@flavour Please merge...

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -537,8 +537,8 @@ def alert():
                 if alert_id:
                     itable.web.default = current.deployment_settings.get_base_public_url()+\
                                          URL(c="cap", f="alert", args=alert_id)
-            
-            
+
+
             elif r.component_name == "area":
                 atable = r.component.table
                 # Limit to those for this Alert
@@ -818,7 +818,7 @@ def template():
                 field.required = False
 
             itable.category.required = False
-            
+
             alert_id = request.args(0)
             # Check for prepopulate
             if alert_id:
@@ -972,7 +972,6 @@ def priority_get():
                                         wptable.name,
                                         orderby = wptable.id)
                 result = rows.json()
-                #set_priority_js(event_type_name)
                 # Uses "others" event_type
                 # Use this according to deployment
                 #from gluon.serializers import json as jsons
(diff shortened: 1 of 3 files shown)


controllers/cap.py
```diff
@@ -1013,28 +1037,22 @@ def compose():
...
+            - used by those having the permission to publish
+    """
 
-    priority_conf = '''S3.cap_priorities=%s''' % jsons(p_settings)
-    js_global = s3.js_global
-    if not priority_conf in js_global:
-        js_global.append(priority_conf)
+    # For SAMBRO, permission is checked by the Authentication Roles but the permission
+    # should be checked if CAP module is enabled
+    alert_id = get_vars.get("cap_alert.id")
```
flavour: Why is this a separate controller?
Makes more sense to me as an S3Method

So we'd have cap/alert/[alert_id]/publish

F

modules/s3db/cap.py
```diff
@@ -1563,15 +1561,29 @@ def cap_alert_approve(record=None):
...
+
+            # Notify the owner of the record about approval
+            row = db(query).select(table.owned_by_user,
+                                   limitby=(0, 1)).first()
+            if row.owned_by_user:
+                pe_id = current.auth.s3_user_pe_id(int(row.owned_by_user))
+                subject = "%s: Alert Approved" % settings.get_system_name_short()
+                url = "%s%s" % (settings.get_base_public_url(),
+                                URL(c="cap", f="alert", args=[alert_id]))
+                message = "This alert that you requested to review has been approved:\n%s" % url
```
flavour: no i18n?

modules/s3db/cap.py
```diff
@@ -1563,15 +1561,29 @@ def cap_alert_approve(record=None):
         if alert_id:
             db = current.db
             approved_on = record["approved_on"]
-            db(db.cap_alert.id == alert_id).update(approved_on = current.request.utcnow)
-    
+            table = db.cap_alert
+            query = table.id == alert_id
+            settings = current.deployment_settings
```
flavour: should be moved inside the branch which uses it

controllers/cap.py
biplovbhandari: @flavour I am a bit confused here. As Dominic was saying earlier, the original concept was 'approve' being that the record could come into the database and 'publish' being record sent out. But in SAMBRO, we trust the user so every alert created comes into the database and 'approve' simply gets it out to the general public. So I think if we stick to workflow ATM, the approve should be implemented as I have done here, or otherwise we need to modify the workflow, which would involve two steps -> Create Alert -> Send for Approval -> Approved -> Send for Publish -> Published -> Record to general public.

Since I think we agreed to have this simple as Approved being record sent to general public, the approve method works for now? What are your thoughts?

nursix: I think Fran's point is that it should not be a separate controller, but rather an S3Method, so that it's called via cap/alert/5/publish instead of cap/publish?cap_alert.id=5.

It's not about the workflow, I don't think.

biplovbhandari: Here I am simply approving the record so that it appears in the datatable, name is different though

flavour: Dominic is right...I am NOT trying to complicate/alter the workflow
I am merely saying that the functionality should be at cap/alert/5/publish instead of cap/publish?cap_alert.id=5

PR #1184, opened 2015-10-14

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -537,8 +537,8 @@ def alert():
                 if alert_id:
                     itable.web.default = current.deployment_settings.get_base_public_url()+\
                                          URL(c="cap", f="alert", args=alert_id)
-            
-            
+
+
             elif r.component_name == "area":
                 atable = r.component.table
                 # Limit to those for this Alert
@@ -818,7 +818,7 @@ def template():
                 field.required = False
 
             itable.category.required = False
-            
+
             alert_id = request.args(0)
             # Check for prepopulate
             if alert_id:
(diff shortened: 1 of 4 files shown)


controllers/cap.py
```diff
@@ -1013,6 +1012,24 @@ def compose():
     redirect(URL(c="cap", f="alert"))
 
 # -----------------------------------------------------------------------------
+def alert_approve():
```
nursix: Hmm - yesterday Fran said he wanted this to become an S3Method of cap_alert.

Now you merely renamed the controller - what am I missing here?

flavour: Was exactly my comment earlier

biplovbhandari: This was done as separate function as in others also as here:https://github.com/flavour/eden/blob/master/controllers/vulnerability.py#L1535
Besides, S3crud always outputs view https://github.com/flavour/eden/blob/master/modules/s3/s3crud.py#L790? This is just action? What am I missing here?

flavour: > This was done as separate function as in others also as here: https://github.com/flavour/eden/blob/master/controllers/vulnerability.py#L1535

This is a custom function highly tied to the Vulnerability model, whereas I see nothing CAP-specific about the one you created. (Also the function was created bya  developer who wasn't aware of the S3Method-based approach)

flavour: > S3crud always outputs view https://github.com/flavour/eden/blob/master/modules/s3/s3crud.py#L790? This is just action?

Typically yes, but I don't see that as a complete requirement.

PR #1185, opened 2015-10-14

biplovbhandari:


biplovbhandari:
@flavour Should I add a function in controller to approve as I have done earlier?

flavour:
> Should I add a function in controller to approve as I have done earlier?

No! ;) The review page is /OK/ (although I prefer to avoid it)...it's just the label I was suggesting to change right now.  This first iteration can be refined with the POST script later (I'm sure it will be necessary as this extra step seems clumsy to me)

PR #1186, opened 2015-10-14
CAP: represent area name with local name

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -522,6 +522,19 @@ def alert():
                                    )
 
                     response.s3.stylesheets.append("../themes/default/cap.css")
+                    
+                elif r.method == "assign":
+                    # Represent each row with local name if available
+                    translate = settings.get_L10n_translate_cap_area()
+                    if translate:
+                        if session.s3.language == settings.get_L10n_default_language():
+                            translate = False
+                
+                        if translate:
+                            from s3 import S3Represent
+                            atable = s3db.cap_area
+                            cap_area_represent = cap_AssignAreaRowRepresent()
+                            atable.name.represent = S3Represent(options=cap_area_represent)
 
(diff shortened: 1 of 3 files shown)


nursix:
I don't think this is about field value representation at all - it's rather to select between two alternative fields in the view.

The elegant solution is therefore to include both fields in the select (as extra_fields), and then have a Field.Method that returns either the name_l10n (if it is not empty) or otherwise the name:

    local_name = row["cap_area_name.name_l10n"]
    return local_name if local_name else row["cap_area.name"]

The filtering of the name_l10n by current language can be achieved by adding e.g. cap_area_name as a filtered component "local_name" to cap_area, and use language as filterby. Then specify "local_name.name_l10n" as extra_field (along with "name").

That way, you do not need to conduct any lookup, not write any complicated S3Represent subclass, and yet you get a very efficient mechanism.

biplovbhandari:
I think this is about field value (cap_area.name) representation as I am doing here: https://github.com/flavour/eden/pull/1186/files#diff-deea28c62cb200676ae57b2e1b30a34bR1138 (building representation options)? Isn't it? Or am I missing something here?

nursix:
You are doing that, yes - but that doesn't mean that is what it is about. 

In principle, you want to show the localized name or, if it is not available, fall back to the original name. And that is simply choosing between two alternative name fields, not representation of the name.

Including both fields, and implementing the fallback as Field.Method is the easiest and most efficient variant. Building a dict of representation options is not required here.

nursix:
Or am I wrong that in "assign" (=cap_AssignArea), you are showing a list of areas, and in that list you want to show either the local name of the area, or, if that is not available, the original name?

nursix:
Ok ok, maybe I'm wrong and misinterpreting your code. This is closed anyway, so feel free to disregard.

PR #1187, opened 2015-10-15
CAP: represent area name with local name + SAMBRO: Added RSS BUTTON

biplovbhandari:


biplovbhandari:
@flavour Please merge

PR #1188, opened 2015-10-17
SAMBRO: warning_priority refactored + filtered options for S3OptionsFilter + SAMBRO: rss icon to the homepage

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/views/index.html
+++ b/modules/templates/SAMBRO/views/index.html
@@ -1,10 +1,11 @@
 {{extend "layout.html"}}
+<div id='home'>
 <h1>{{=title}}</h1>
 <div class='subscribe-button'>
  {{if current.auth.s3_logged_in():}}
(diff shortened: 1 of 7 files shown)


modules/templates/SAMBRO/views/index.html
```diff
@@ -1,10 +1,11 @@
 {{extend "layout.html"}}
+<div id='home'>
 <h1>{{=title}}</h1>
 <div class='subscribe-button'>
  {{if current.auth.s3_logged_in():}}
-  <a class='button alert tiny' href='/{{=appname}}/cap/alert.rss' target='_blank'>{{=T("RSS Feeds")}}</a>
+  <a href='/{{=appname}}/cap/alert.rss' target='_blank'><i class='icon-rss'></i></a>
```
flavour: I wouldn't get rid of the Text Label - just add the RSS icon too...
Also I think you can be more DRY here:
{{if current.auth.s3_logged_in(): fn = "alert" else fn = "public"}}
<a href='/{{=appname}}/cap/{{=fn}}.rss' target='_blank'><i class='icon-rss'></i></a>

PR #1191, opened 2015-10-20

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -155,6 +155,31 @@ def alert():
                 redirect(URL(c="cap", f="template",
                              args = request.args,
                              vars = request.vars))
+            if r.record.approved_by is not None:
+                # Once approved, don't allow to edit
+                # Don't allow to delete
+                s3db.configure(tablename,
+                               editable=False,
+                               deletable=False,
+                               )
+            # Uncomment or add to this according to country CAP profile
+            #if r.record.msg_type in ("Update", "Cancel", "Error"):
+                # Use case for change in msg_type
+            #    atable = r.table
+            #    for f in ("template_id",
+            #              "sender",
+            #              "status",
+            #              "msg_type",
+            #              "source",
+            #              "scope",
+            #              "restriction",
+            #              "addresses",
+            #              "codes",
+            #              "note",
+            #              "reference",
+            #              "incidents",
+            #              ):
+            #        atable[f].writable = False
         else:
             r.resource.add_filter(r.table.is_template == False)
             s3.formats["cap"] = r.url() # .have added by JS
@@ -551,6 +576,26 @@ def alert():
                     itable.web.default = current.deployment_settings.get_base_public_url()+\
                                          URL(c="cap", f="alert", args=alert_id)
 
+                if r.record.approved_by is not None:
+                    # Once approved, don't allow info segment to edit
+                    # Don't allow to delete
+                    s3db.configure("cap_info",
+                                   editable = False,
+                                   deletable = False,
+                                   )
+                # Uncomment or add to this according to country CAP profile 
+                #if r.record.msg_type in ("Update", "Cancel", "Error"):
+                    # Use case for change in msg_type
+                #    for f in ("language",
+                #              "category",
+                #              "event",
+                #              "event_type_id",
+                #              "audience",
+                #              "event_code",
+                #              "sender_name",
+                #              "parameter",
+                #              ):
+                #        itable[f].writable = False
 
             elif r.component_name == "area":
                 atable = r.component.table
@@ -584,6 +629,14 @@ def alert():
                     field.default = rows.first().id
                     field.writable = field.readable = False
 
+                if r.record.approved_by is not None:
+                    # Once approved, don't allow area segment to edit
+                    # Don't allow to delete
+                    s3db.configure("cap_area",
+                                   editable = False,
+                                   deletable = False,
+                                   )
+
             elif r.component_name == "resource":
                 atable = r.component.table
                 # Limit to those for this Alert
@@ -601,6 +654,14 @@ def alert():
                     field = atable.info_id
                     field.default = rows.first().id
                     field.writable = field.readable = False
+
+                if r.record.approved_by is not None:
+                    # Once approved, don't allow resource segment to edit
+                    # Don't allow to delete
+                    s3db.configure("cap_resource",
+                                   editable = False,
+                                   deletable = False,
+                                   )
 
             # @ToDo: Move inside correct component context (None?)
             post_vars = request.post_vars
@@ -666,20 +727,7 @@ def alert():
                 rows = db(itable.alert_id == lastid).select(itable.id)
 
                 rtable = s3db.cap_resource
-                r_unwanted_fields = set(("deleted_rb",
-                                         "owned_by_user",
-                                         "approved_by",
-                                         "mci",
-                                         "deleted",
-                                         "modified_on",
-                                         "realm_entity",
-                                         "uuid",
-                                         "created_on",
-                                         "deleted_fk",
-                                         "created_by",
-                                         "modified_by",
-                                         "owned_by_group",
-                                         ))
+                r_unwanted_fields = s3base.s3_all_meta_field_names()
                 rfields = [rtable[f] for f in rtable.fields
                                      if f not in r_unwanted_fields]
                 rows_ = db(rtable.alert_id == alert.template_id).select(*rfields)
@@ -1068,5 +1116,122 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
                                        row_["cap_area.name"])
                             
         return cap_area_options
+
+# -----------------------------------------------------------------------------
+def clone_alert():
+    """
+        Function to clone the cap_alert along with its components
+            Used for the change the message status
+    """
+    
+    from s3 import s3_utc
+    caller = get_vars.get("caller")
+    alert_id = request.args(0)
+    
+    alert_table = s3db.cap_alert
+    info_table = s3db.cap_info
+    area_table = s3db.cap_area
+    location_table = s3db.cap_area_location
+    tag_table = s3db.cap_area_tag
+    resource_table = s3db.cap_resource
+    unwanted_fields = s3base.s3_all_meta_field_names()
+    
+    # Copy the alert segment
+    alert_fields = [alert_table[f] for f in alert_table.fields
+                    if f not in unwanted_fields]
+    alert_row = db(alert_table.id == alert_id).select(*alert_fields,
+                                                      limitby=(0, 1)).first()    
+    alert_row_clone = alert_row.as_dict()
+    del alert_row_clone["id"]
+    del alert_row_clone["identifier"]
+    alert_row_clone["msg_type"] = caller
+    alert_row_clone["reference"] = ("%s,%s,%s") % (alert_row.sender,
+                                                   alert_row.identifier,
+                                  str(s3_utc(alert_row.sent)).replace(" ", "T"),
+                                                   )
+    new_alert_id = alert_table.insert(**alert_row_clone)
+    auth.s3_set_record_owner(alert_table, new_alert_id)
+    s3db.onaccept(alert_table, {"alert_id": new_alert_id})
+    
+    # Copy the info segment
+    info_fields = [info_table[f] for f in info_table.fields
+                   if f not in unwanted_fields]
+    info_rows = db(info_table.alert_id == alert_id).select(*info_fields)
+    if info_rows:
+        for info_row in info_rows:
+            info_id = info_row.id
+            info_row_clone = info_row.as_dict()
+            del info_row_clone["id"]
+            info_row_clone["alert_id"] = new_alert_id
+            new_info_id = info_table.insert(**info_row_clone)
+            auth.s3_set_record_owner(info_table, new_info_id)
+            s3db.onaccept(info_table, {"info_id": new_info_id})
+            
+            # Copy the area segment
+            query = (area_table.info_id == info_id) &\
+                    (area_table.alert_id == alert_id)
+            area_fields = [area_table[f] for f in area_table.fields
+                           if f not in unwanted_fields]        
+            area_rows = db(query).select(*area_fields)
+            if area_rows:   
+                for area_row in area_rows:
+                    area_id = area_row.id
+                    area_row_clone = area_row.as_dict()
+                    del area_row_clone["id"]
+                    area_row_clone["alert_id"] = new_alert_id
+                    area_row_clone["info_id"] = new_info_id
+                    new_area_id = area_table.insert(**area_row_clone)
+                    auth.s3_set_record_owner(area_table, new_area_id)
+                    s3db.onaccept(area_table, {"area_id": new_area_id})
+                    
+                    # Copy the area_location
+                    location_fields = [location_table[f] for f in location_table.fields
+                                       if f not in unwanted_fields]
+                    location_rows = db(location_table.area_id == area_id).\
+                                                    select(*location_fields)
+                    if location_rows:
+                        for location_row in location_rows:
+                            location_row_clone = location_row.as_dict()
+                            del location_row_clone["id"]
+                            location_row_clone["alert_id"] = new_alert_id
+                            location_row_clone["area_id"] = new_area_id
+                            new_location_id = location_table.insert(**location_row_clone)
+                            auth.s3_set_record_owner(location_table, new_location_id)
+                            s3db.onaccept(location_table,
+                                          {"location_id": new_location_id})
+                            
+                    # Copy the area_tag
+                    tag_fields = [tag_table[f] for f in tag_table.fields
+                                  if f not in unwanted_fields]
+                    tag_rows = db(tag_table.area_id == area_id).\
+                                                        select(*tag_fields)
+                    if tag_rows:
+                        for tag_row in tag_rows:
+                            tag_row_clone = tag_row.as_dict()
+                            del tag_row_clone["id"]
+                            tag_row_clone["alert_id"] = new_alert_id
+                            tag_row_clone["area_id"] = new_area_id
+                            new_tag_id = tag_table.insert(**tag_row_clone)
+                            auth.s3_set_record_owner(tag_table, new_tag_id)
+                            s3db.onaccept(tag_table, {"tag_id": new_tag_id})
+                
+                # Copy the resource segment
+                query = (resource_table.info_id == info_id) &\
+                        (resource_table.alert_id == alert_id)
+                resource_fields = [resource_table[f] for f in resource_table.fields
+                                   if f not in unwanted_fields]                
+                resource_rows = db(query).select(*resource_fields)
+                if resource_rows:
+                    for resource_row in resource_rows:
+                        resource_row_clone = resource_row.as_dict()
+                        del resource_row_clone["id"]
+                        resource_row_clone["info_id"] = new_info_id
+                        resource_row_clone["alert_id"] = new_alert_id
+                        rid = resource_table.insert(**resource_row_clone)
+                        auth.s3_set_record_owner(resource_table, rid)
+                        s3db.onaccept(resource_table,
+                                      {"resource_id": new_resource_id})
+                    
+        redirect(URL(c="cap", f="alert", args=[new_alert_id]))
 
 # END =========================================================================
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -398,13 +398,15 @@ class S3CAPModel(S3Model):
                                          _title="%s|%s" % (T("The text describing the purpose or significance of the alert message"),
                                                            T("The message note is primarily intended for use with status 'Exercise' and message type 'Error'"))),
                            ),
-                     Field("reference", "list:reference cap_alert",
+                     Field("reference", #"list:reference cap_alert",
                            label = T("Reference"),
-                           represent = S3Represent(lookup = tablename,
-                                                   fields = ["msg_type", "sent", "sender"],
-                                                   field_sep = " - ",
-                                                   multiple = True,
-                                                   ),
+                           writable = False,
+                           readable = False,
+                           #represent = S3Represent(lookup = tablename,
+                           #                        fields = ["msg_type", "sent", "sender"],
+                           #                        field_sep = " - ",
+                           #                        multiple = True,
+                           #                        ),
                            comment = DIV(_class="tooltip",
                                          _title="%s|%s" % (T("The group listing identifying earlier message(s) referenced by the alert message"),
                                                            T("The extended message identifier(s) (in the form sender,identifier,sent) of an earlier CAP message or messages referenced by this one."))),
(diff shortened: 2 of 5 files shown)


controllers/cap.py
```diff
@@ -1069,4 +1117,122 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
                             
         return cap_area_options
 
+# -----------------------------------------------------------------------------
+def cap_msgTypeChangeFn():
```
nursix: This violates our naming conventions - function names are all-lowercase (CamelCase), and underscores as word separator.

And in th controller, you do not need the cap_ prefix.

controllers/cap.py
```diff
@@ -1069,4 +1117,122 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+    """
+        Function to change the message status
```
nursix: This is a very poor docstring, not the slightest explaining why you need such a hyper-complex procedure just to change the message status!

nursix: The truth is that this function is cloning an alert, including all its components - which is not exactly what "changing status" means to me. Why not call it "clone" and say that in the docstring - if that is what it does?

controllers/cap.py
```diff
@@ -1069,4 +1117,122 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                                                    select(*location_fields)
+                    if location_rows:
+                        for location_row in location_rows:
+                            location_row_clone = location_row.as_dict()
+                            del location_row_clone["id"]
+                            location_row_clone["alert_id"] = new_alert_id
+                            location_row_clone["area_id"] = new_area_id
+                            new_location_id = location_table.insert(**location_row_clone)
+                            auth.s3_set_record_owner(location_table, new_location_id)
+                            row = db(location_table.id == new_location_id).\
```
nursix: You do not need to re-select the new row, it is good enough if you add the new record id to the dict, and then pass that dict.

controllers/cap.py
nursix: Moreover, the function name and the docstring tell two different stories, and the actual function yet another one. Somehow this should be sorted out.

Why is it necessary to clone the entire alert?

modules/s3db/cap.py
```diff
@@ -301,7 +301,7 @@ def model(self):
                            default = "{}",
                            readable = False,
                            ),
-                     Field("identifier", unique=True, length=128,
+                     Field("identifier", length=128, #unique = True
```
nursix: Such changes are evil (because they do not auto-migrate, and hence require each and every existing Sahana system to be manually migrated).

Please explain why you're doing this, why you suddenly need to allow (in fact: deliberately produce) duplicates (clones), and how you want to deduplicate alerts.

nursix:
IMHO this is the wrong concept, and has always been, which is why I recommended to not to use the approval-framework in the first place.

The alert should be copied into a separate table ("published alerts") when it gets approved and published, and that table should generally be locked for all writes except by the publishing-workflow. That extra table can have additional meta-fields (for message tracking), and a different access level.

It *is* a publishing workflow after all, exactly like I said - and this requirement here shows why that is different from the input-approval.

biplovbhandari:
@nursix The workflow here is that after an alert has been published; at time we want to send the update or error if it is(well we want to minimize this, but we cannot control human error). So in that case we have the same alert(with same identifier small components-info, area and resource) and we update certain fields (like msg_type, instruction, headline etc.)(this depends on country profile so I have commented them out here: https://github.com/flavour/eden/pull/1191/files#diff-deea28c62cb200676ae57b2e1b30a34bR165
So this need of cloning the alert along with its components.

nursix:
I would rewrite the whole thing, and introduce a cap_published_alert table that holds those alerts that have been published. 

Copy the cap_alert into cap_published_alert when it gets approved for dissemination, and link all published copies per link table to the original cap_alert.

The cap_alert table would then be restricted to alert editors, whilst the cap_published_alert table would be open to everyone, but read-only. And all subscriptions/notifications would work from the cap_published_alert table only.

That is a *much* cleaner and safer concept, and it allows tracking of all published versions easily. Additionally, the cap_published_alert table can hold the EDXL-DE meta fields for message tracking, if that is still a relevant concept for SAMBRO.

nursix:
Regardless the concept, though - there are some violations of coding conventions you should fix.

And for making the alert identifier non-unique - I think this is debatable. Is it valid, anyway? I'm not sure it is for CAP - I think alert identifiers are meant to be unique and must not have multiple versions with the same ID.

nursix:
AFAIK you must send an update referencing the original rather than sending another version of the original - hence you must use a new identifier. Please validate that.

nursix:
Yup - correct. Re-using the same identifier violates the CAP standard - if you want to send an update, then you must issue a new alert (new identifier) with a *reference* to the original. Please read in the CAP 1.2 specification about how to rectify alerting errors.

Alert identifiers must be unique - reusing them will give consuming systems a hard time, and they may even refuse to accept the update if it carries the same identifier. Sending multiple versions with the same identifier will also undermine the credibility of the Sahana system as alert source.

This is the minimum you must change.

I would reconsider the whole concept, however.

nursix:
Like in the example:

    <alert xmlns = "urn:oasis:names:tc:emergency:cap:1.2">
        
        <identifier>TRI13970876.2</identifier> 
        
        <msgType>Update</msgType>
        
        <references>trinet@caltech.edu,TRI13970876.1,2003-06-11T20:30:00-07:00</references>

This is what the standard defines how you can update inappropriate alerts - and not by sending out a new version with the same identifier (I like the version-numbering in this example btw.).

I think it is important to adhere to this standard, because consuming systems will expect it.

biplovbhandari:
@nursix I think you are correct, the identifier should be different like http://publicalert.pagasa.dost.gov.ph/output/gfa/29ebde5a-0809-4d75-9866-fb84d11bf42c.cap and http://publicalert.pagasa.dost.gov.ph/output/gfa/45827385-ce01-40ec-87d6-d94fe9d75d79.cap. That way we can fill in the reference field to the earlier alert.

@waidyanatha your thought on this?

controllers/cap.py
biplovbhandari: @nursix Sorry my bad...will fix that

controllers/cap.py
biplovbhandari: @nursix Sorry again...I forget to update the docstring :(

waidyanatha:
The reference field is used if an Alerting Authority relays a CAP message
that was originated by another Alerting Authority. For example, when the
Maldives Met Services issues a CUG Alert that is received by the Maldives
Nat'l Disaster Management Center and then NDMC decides to relay that
message to the public. In this case, the identifier will get replaced with
the NDMC OID and the MMS originated identifier gets added to the
<reference> element.

In general it is a good habit to always refer to the CAP specifications
document when coding:
http://docs.oasis-open.org/emergency/cap/v1.2/CAP-v1.2-os.html

nursix:
That does not answer the question, @waidyanatha 

> 2.3.5     Repudiating a False Alarm
> Inadvertently the integrated alerting network has been activated with an inaccurate warning 
> message. This activation comes to officials' attention immediately through their own monitoring 
> facilities (e.g., 2.3.3 above).  Having determined that the alert is, in fact, inappropriate, the officials 
> issue a cancellation message that refers directly to the erroneous prior alert.  Alerting systems that 
> are still in the process of delivering the alert (e.g., telephone dialing systems) stop doing so.  
> Broadcast systems deliver the cancellation message. Other systems (e.g., highway signs) simply 
> reset to their normal state.

I don't think it is correct to just publish an updated version of the original alert using the same identifier - this would contradict the general principles of a messaging protocol, and CAP in particular.

The original alert will have triggered responses, and just re-publishing the alert would not take back those responses. Moreover, as it uses the same identifier, systems will treat it as "already processed" (that is what the identifier is used for!), and hence the update may not trigger anything at all, and all the changes go unnoticed.

Hence, you must issue a new alert of msgType "Update" with a new identifier, that references the original alert, and explicitly triggers response updates.

Additionally, it will undermine the credibility of the alert source if there are multiple versions of the same alert, without explicitly indicating an update. That looks dodgy.

nursix:
No - sent is sent, you can not re-publish an updated alert under the same identifier. No messaging protocol in the world allows that, and neither does CAP (if it would, it would be fundamentally flawed).

If you want to update an already-published alert, then you must send an update message with a new identifier, and a reference to the original alert.

Using version-numbering like in the example above makes it human-comprehensible, which would provide an additional means for verification - but since CAP is meant to be machine-processed in the first line, this is secondary.

Most, if not all, automated systems will track which alerts they already have responded to, so if we want them to respond to an update (and indeed we do want that), we must use a new identifier and an explicit indication that this is an update of a previous alert. That's how CAP is designed, is it not?

nursix:
This is why I keep proposing to not just use the Sahana approval-mechanism, but to also explicitly copy any published alert into a separate table (like the "sent messages" box in your email program), and generally lock that table against interactive writes of any kind.

And then, you should only ever assign a unique alert identifier at the moment when you actually publish the alert (=send the message). Perhaps the alert table could hold a base identifier, and when copying/sending it, you would append a version number - that way preventing that there can every be two messages with exactly the same identifier.

But even with the "cloning" concept, you still need to make sure that an update create a new message.

The point is that CAP is a *messaging* protocol, and unlike with representational protocols (e.g. REST), the sender is not in control of the object - the message is the object you've sent away, not the object you have in the database!

And since you have no control of the message object - you can not update it. You can only ever generate a new message object that overrides the previous one - and being a new message object, it has to carry a new identifier (otherwise it wouldn't be a new object, that's the point with identifiers).

waidyanatha:
Hi Dominic,

Thanks for these insights and for raising these valid points. Let me
investigate this a bit further.

I get your point but curious whether it is illegal or unethical to send a
massage with the same ID more than once in this world?

Nuwan

nursix:
@waidyanatha : is is neither "illegal" nor "unethical" - it is a protocol violation, and thus an incorrect implementation.

The only question is whether the changed message is still the same message - because only then it can use the same identifier again due to the uniqueness-requirement specified by the CAP standard.

And by definition a message is only ever the same message if all possible responses are invariable to the change (e.g. a different data format, or a different messaging channel) - but the very reason to send an updated version in the first place is *because* we want to alter responses, and that does make it a new message.

And a new message MUST have a new identifier.

Come on - this is information technology basics, high school level - why the debate?

waidyanatha:
agree with you, what my worry is that the definition of the identifier
might be wrong if it is supposed to refer to the specific event and not the
message. Specs use the term "this warming message". I need to clarify from
the guru's that it does not refer to the event.

waidyanatha:
BTW - I skipped high school and went fishing with my friends instead

nursix:
:D Okay, I see - fishing is bad, because you can indeed catch the same fish twice and each time it's different. So I understand that.

Anyway - CAP is about alerting, not about event management, so does not define any "event" entity, and consequently it does not hold any event object reference. I can tell you as much ;) 

But yes, feel free to double-check with the TC.

waidyanatha:
This what the Canadians do - We create a new identifier each time.

Every element in that first group of elements, prior to the <info> block,
attributes the same thing and that thing is the Alert Message (not the
Alert).

We know this because for most of the other elements it is obvious they are
attributing the Alert Message (<status>, <sent>, <references>, etc...) so
<identifier> must too. Also, none of those elements are required to be
"persistent" across the series of messages that make up the alert.

Also, the spec says... "A number or string uniquely identifying this
message, assigned by the sender."

waidyanatha:
Philippines also does the same, Lester said - "We create a new
identifier then reference the previous identifier."

I hope the data structure has another kind of ID that allows us to
bundle the messages that are related to a singe event? This is
necessary for presentation and for audit trails. Dominic or Biplov
might have thought of it or explained the strategy. If you yes, then
good.

nursix:
@waidyanatha CAP does not provide any such identifier, because, as I said before, it doesn't deal with event management at all and hence does not define any event entity.

Sahana, however, does have an event management module, and it seems trivial to me to add an event_incident_id() to cap_alert, so that alerts can a) be filtered by incident, b) appear on a tab under the respective incident and c) can be reported on by incident.

But again - even if we implement that in SAMBRO/Sahana, there will be no way to express the relationship in CAP alerts and hence communicate that to message recipients - because the CAP standard does not define any such thing like an "event".

The reason behind this lack of a CAP data structure is very simple: there is no common understanding of what "event" actually refers to - every country, every organisation use their own definition. That makes it impossible to incorporate that into the CAP standard - apart of course from the fact that it isn't actually relevant for the alerting as such.

Note that what you call "event" is called "incident" in Sahana (whereas a Sahana "event" can have many "incidents").

nursix:
NB CAP is XML, and like with any XML, you can easily embed additional elements under a separate namespace - and hence it will very well be possible to add the incident reference to the alert message.

Just not with CAP elements.

waidyanatha:
The CAP module's tables carry more information than a typical cap message
does. For example, we have introduced a table called the warning_priority.
We also use the Incident values from the IRS module (or whatever it is
called). This is for our own controls, presentation, ease-of-use, and
usefulness. Basially, SAMBRO offers a comprehensive multi-agency alerting
system and it is also CAP compliant, meaning we can offer a CAP RSS Feed
that adheres to the CAP standard.

waidyanatha:
Correct, that is one approach. The other would be to use the <parameter>
element was introduced for, to embed other data elements and values.

nursix:
Obviously, yes - question is though what purpose it would serve to put the Sahana incident reference into the CAP feed? Are there *any* consuming systems who would process it, and if yes - what for?

There is no "official" (authoritative) registry of incident identifiers - nor do alert responders normally manage alerts by incidents (but rather responses by alert). I think such a feature would be very specific for individual peer systems - and should maybe only be added if/when actually required for a particular integration.

However, within Sahana it would easily be possible to add a link between alerts and incidents (as described above), with all the standard functionality that would add. And yet - that's a possible solution, now: can you specify the possible problem that it would solve (or would it just be "cool")?

nursix:
...and are you requesting implementation from me, or are we just discussing Biplov's work?

nursix:
onsdagen den 21 oktober 2015 04.06.11 skrev  Nuwan Waidyanatha:
> The CAP module's tables carry more information than a typical cap message
> does. For example, we have introduced a table called the warning_priority.
> We also use the Incident values from the IRS module (or whatever it is
> called). This is for our own controls, presentation, ease-of-use, and
> usefulness. Basially, SAMBRO offers a comprehensive multi-agency alerting
> system and it is also CAP compliant, meaning we can offer a CAP RSS Feed
> that adheres to the CAP standard.
So? Rehearsing the presentation - or is this meant to specify a requirement?

Struggling to understand what you're after here - maybe we should switch to a 
voice channel for better comms.

nursix:
@waidyanatha As much as I understand the SAMBRO project, event/incident management would be a significant extension of scope - but even if logical, I don't yet see a sufficient use-case definition anywhere in the documentation (...the parts I have visibility of).

Even a simple, workflow-less incident registry would add technical requirements (such as data entry), and right now, that is not described anywhere. Who would register incidents? Where do the data come from? What data exactly? Is filtering by incident the only relevant user task in this regard?

I see the value, but not enough information yet to action it.

waidyanatha:
​Just as much as I would like you to be involved in the design of SAMBRO,
it is forbidden!. These are simply brainstorming session and I am not
requesting any implementation because that would be a violation.

SAMBRO - it is a vision beyond CAP and warning. The design would serve as
an Early Warning System and an Incident Command System; mainly working
within the response phase of the disaster management cycle. It does bleed
into preparedness with risk mapping because we need to understand the
risks. I have limited the wiki to CAP, for now, to avoid confusing the
principals and developers, and having to explain a lot of the intricacies.
The need for a comprehensive SAMBRO was confirmed during the August
workshop, read the last paragraph:
http://sahanafoundation.org/cap-on-a-map-tot-week-one/ . When I find the
time, I intend to revamp the Blueprint. Here's a set of incomplete slides
that may paint a picture of the full scope:
https://docs.google.com/presentation/d/11GbmMmMJVcp8wGfrjQbm02_eWuFTaTnRuIKAxjCUrI4/edit#slide=id.g7085e25d6_0_0

Incidents - As you know CAP Alert block already carries the Incident
element. Naturally, we thought of using the Sahana Incident model to
provide the look up values for this field. That is how it got linked to the
CAP module. Users would edit the Incident Model, through the views, to
their liking. Unfortunately there is no "authoritative Incident dataset"
defined for CAP, in general. The Australians have their own and the
Canadians/Americans have their own. Canadians and Americans (peer system)
share the Incident values in their CAP cross-border messages. It is mainly
for systems, more so than humans, that use these values (or codes) for
applying system logic. One aspect I'm trying to introduce/test in the next
phase of this project is cross-boarder alerting and peer systems. I don't
think our partners understand the importance of it or the concept.

nursix:
CAP refers to incident /types/, not to individual incidents - and that part is indeed implemented using the Sahana event management module.

The question is whether we additionally need to reference individual incidents - which in turn would require to also *track* those individual incidents in SAMBRO. 

And this tracking of individual incidents is not currently described as a SAMBRO requirement, and in fact, not in these overview documents either. It seems that the documentation simply /assumes/ that such incident tracking is happening - and doesn't define whether that happens within SAMBRO or outside of it.

If it was to happen within SAMBRO, it should use the event_incident entity - which is part of the "event" module (not IRS). The "event" module is already used by SAMBRO (cap_alert references the event_incident_type taxonomy, which is the same taxonomy used by event_incident), so it would more or less just mean to:

a) expose the incident tracking views (menu item, user permissions)
b) link cap_alert to event_incident (instead of or in addition to event_incident_type)
c) add the incident tracking workflow to the specs

This link would add all the CRUD and mapping functionality around incidents, and, the cool thing is that this easily extends into both sitreps (which are attached to event_incident) and dispatch (also attached to event_incident), thereby providing the full cycle you're after.

And yet - nowhere in the documentation or presentations it is actually mentioned that SAMBRO would need to track incidents. 

This is simply *the* missing link between all the components, and yet it's not mentioned anywhere. Almost as if it would happen outside of SAMBRO. Does it?

nursix:
It will be obvious to you that if you want to reference individual incidents, then you will have to track incidents. As a very minimum, you need to register them.

But SAMBRO, so far, has been built around alerts - all workflows start and end at alerts, and no individual incidents are even recorded.

Generally, as I understand it, SAMBRO is supposed to be a comprehensive messaging and alerting system - but not response management. But incident tracking is actually about response management rather than messaging/alerting, so I still wonder how this would fit into the scope.

It is very well possible that the incident tracking happens outside of SAMBRO (e.g. in a separate ICS), we have that situation in other places. And if that is the case, then the question goes to how to integrate that separate response management.

If you, however, envisage response management to be integrated in SAMBRO, then incident tracking is, obviously, not the end of the story - it will also require resource management. For instance, if you want to dispatch a unit (e.g. a fire engine) to an incident, then you must track both - the incident, and obviously also the unit.

So, this is a bit unclear still.

Doesn't mean a simple incident tracking couldn't just be enabled for SAMBRO ;)

If you are not involved with the SAMBRO design, then who is? Who does specify the requirements?

waidyanatha:
> CAP refers to incident /types/, not to individual incidents - and that
> part is indeed implemented using the Sahana event management module.
>

​That is correct, I was loosely referring to this as incident​

​but we are using the incident types or something of that order.​

The current implementation or developments of SAMBRO does not imply it is
using any incident management work flows. These are yet to be designed and
developed; i.e. a
s mentioned I have yet to define the incident tracking.

As of now my knowledge in
​the incident management aspects are
 based on the SITREP research work we did in 2012
​. It involved​
integrating
​the Freedom Fone
IVR with
​ ​
Sahana-Eden
​ (integration was not from an engineering prospective but a work flow
prospective)​
.​

In these workflows we experimented how we would first alert CERT of an
event
​(using EDXL-CAP messages) ​
and then have the CERT
​respond to the event to ​
provide incident reports
​. For example, we alert the CERT of a Landslide and then they visit the
site to provide the ground situation.; i.e. registering incidents.

Registering Incidents was ​
namely
​ sending​
​EDXL-SITREP ​
field-observation and casualty-illness reports
​ to the EOC​
. Then
​ the EOC would extend
those reports
​to
situational-information reports. Emergency managers use
​d​
the situational-information to derive the response resources
​ reports​
. That is where we stopped
​our
workflow
​s​
.
​Subsequently, the workflow
would extend to resource messaging to dispatch the required resources to
respond to those incidents.

If it was to happen within SAMBRO, it should use the event_incident entity
> - which is part of the "event" module (not IRS). The "event" module is
> already used by SAMBRO (cap_alert references the event_incident_type
> taxonomy, which is the same taxonomy used by event_incident), so it would
> more or less just mean to:
>
> a) expose the incident tracking views (menu item, user permissions)
> b) link cap_alert to event_incident (instead of or in addition to
> event_incident_type)
> c) add the incident tracking workflow to the specs
>

​Thanks for lining this out.​


> This link would add all the CRUD and mapping functionality around
> incidents, and, the cool thing is that this easily extends into both
> sitreps (which are attached to event_incident) and dispatch (also attached
> to event_incident), thereby providing the full cycle you're after.
>

​Superb! Does this mean there is already a EDXL-SITREP API that can
generate SITREP messages from the Event-Incident?​

And yet - nowhere in the documentation or presentations it is actually
> mentioned that SAMBRO would need to track incidents.
>
> This is simply *the* missing link between all the components, and yet
> it's not mentioned anywhere. Almost as if it would happen outside of
> SAMBRO. Does it?
>
​For now I only have this image describing the essential components of
SAMBRO: http://eden.sahanafoundation.org/wiki/Deployments/SAMBRO#DESIGN ​
​But it does not spell out the actual design using Sahana language.​

Incident tracking would be distinct from early warning. They are somewhat
intertwined. An incident tracking workflow may originate from an alert
(e.g. landslide alert instigates responding to the incident) and an alert
may originate from an incident (reporting a train derailment incident may
trigger alerting all First-Responders).

The separation may occur by distinguishing Alerting Authorities from
Incident Managers. Police, for example, could be both an Alerting Authority
reporting of some civil unrest and also a Incident Mangers responding to
the civil unrest to contain the situation. Until such time I come across a
real situation, I am not sure whether we want to assign the roles
associated with alerting and incident management to a single user?

​Regarding the scope of SAMBRO and the terminology of "alerting and
messaging" - life is event driven. We make plans but certain events shape
the course of our lives. Similarly SAMBRO is also event driven. When I
first coined the term it was inclined towards alerting. But now I want to
extend that to incident command and control. In a very non-technical and
way I am associating the word messaging with Incident Command and Control
messaging. Once we mature we might change the term from SAMBRO to something
else. For now that is what the users are familiar with and would stick to
it for a while.​

​Regarding the SAMBRO specification, it is very much a research and
development approach. Developing countries in Asia are slowly developing
their disaster management systems. They have an idea what to do but don't
always know how to do it. What I am offering is a simple and basic platform
that allows them to build on. Strategy is to start with an EWS. Then slowly
migrate them in to ​ICS. The initial requirements comes from my research
experience. With that we have developed an egg. We will slowly mature the
egg into a chicken through a community-centric participatory design and
development approach. The subsequent set of requirements will come from the
users (or adopting community).

flavour:
> Does this mean there is already a EDXL-SITREP API that can
generate SITREP messages from the Event-Incident?​

No...supporting a workflow doesn't mean being able to expose that via a specific Standard without additional work....experience of EDXL standards suggests significant work ;)

PR #1192, opened 2015-10-21

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -155,6 +155,31 @@ def alert():
                 redirect(URL(c="cap", f="template",
                              args = request.args,
                              vars = request.vars))
+            if r.record.approved_by is not None:
+                # Once approved, don't allow to edit
+                # Don't allow to delete
+                s3db.configure(tablename,
+                               editable=False,
+                               deletable=False,
+                               )
+            # Uncomment or add to this according to country CAP profile
+            #if r.record.msg_type in ("Update", "Cancel", "Error"):
+                # Use case for change in msg_type
+            #    atable = r.table
+            #    for f in ("template_id",
+            #              "sender",
+            #              "status",
+            #              "msg_type",
+            #              "source",
+            #              "scope",
+            #              "restriction",
+            #              "addresses",
+            #              "codes",
+            #              "note",
+            #              "reference",
+            #              "incidents",
+            #              ):
+            #        atable[f].writable = False
         else:
             r.resource.add_filter(r.table.is_template == False)
             s3.formats["cap"] = r.url() # .have added by JS
@@ -551,6 +576,26 @@ def alert():
                     itable.web.default = current.deployment_settings.get_base_public_url()+\
                                          URL(c="cap", f="alert", args=alert_id)
 
+                if r.record.approved_by is not None:
+                    # Once approved, don't allow info segment to edit
+                    # Don't allow to delete
+                    s3db.configure("cap_info",
+                                   editable = False,
+                                   deletable = False,
+                                   )
+                # Uncomment or add to this according to country CAP profile 
+                #if r.record.msg_type in ("Update", "Cancel", "Error"):
+                    # Use case for change in msg_type
+                #    for f in ("language",
+                #              "category",
+                #              "event",
+                #              "event_type_id",
+                #              "audience",
+                #              "event_code",
+                #              "sender_name",
+                #              "parameter",
+                #              ):
+                #        itable[f].writable = False
 
             elif r.component_name == "area":
                 atable = r.component.table
@@ -584,6 +629,14 @@ def alert():
                     field.default = rows.first().id
                     field.writable = field.readable = False
 
+                if r.record.approved_by is not None:
+                    # Once approved, don't allow area segment to edit
+                    # Don't allow to delete
+                    s3db.configure("cap_area",
+                                   editable = False,
+                                   deletable = False,
+                                   )
+
             elif r.component_name == "resource":
                 atable = r.component.table
                 # Limit to those for this Alert
@@ -601,6 +654,14 @@ def alert():
                     field = atable.info_id
                     field.default = rows.first().id
                     field.writable = field.readable = False
+
+                if r.record.approved_by is not None:
+                    # Once approved, don't allow resource segment to edit
+                    # Don't allow to delete
+                    s3db.configure("cap_resource",
+                                   editable = False,
+                                   deletable = False,
+                                   )
 
             # @ToDo: Move inside correct component context (None?)
             post_vars = request.post_vars
@@ -666,20 +727,7 @@ def alert():
                 rows = db(itable.alert_id == lastid).select(itable.id)
 
                 rtable = s3db.cap_resource
-                r_unwanted_fields = set(("deleted_rb",
-                                         "owned_by_user",
-                                         "approved_by",
-                                         "mci",
-                                         "deleted",
-                                         "modified_on",
-                                         "realm_entity",
-                                         "uuid",
-                                         "created_on",
-                                         "deleted_fk",
-                                         "created_by",
-                                         "modified_by",
-                                         "owned_by_group",
-                                         ))
+                r_unwanted_fields = s3base.s3_all_meta_field_names()
                 rfields = [rtable[f] for f in rtable.fields
                                      if f not in r_unwanted_fields]
                 rows_ = db(rtable.alert_id == alert.template_id).select(*rfields)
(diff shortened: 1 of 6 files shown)


controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                                                      limitby=(0, 1)).first()    
+    alert_row_clone = alert_row.as_dict()
+    del alert_row_clone["id"]
+    del alert_row_clone["identifier"]
+    alert_row_clone["msg_type"] = caller
+    alert_row_clone["reference"] = ("%s,%s,%s") % (alert_row.sender,
+                                                   alert_row.identifier,
+                                  str(s3_utc(alert_row.sent)).replace(" ", "T"),
+                                                   )
+    new_alert_id = alert_table.insert(**alert_row_clone)
```
nursix: You are aware that this function is open for everyone, hence even anonymous users can clone alerts. Is that correct? Wouldn't it be better to make this an S3Method and call it via the alert() controller?

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+    
+    # Copy the alert segment
+    alert_fields = [alert_table[f] for f in alert_table.fields
+                    if f not in unwanted_fields]
+    alert_row = db(alert_table.id == alert_id).select(*alert_fields,
+                                                      limitby=(0, 1)).first()    
+    alert_row_clone = alert_row.as_dict()
+    del alert_row_clone["id"]
+    del alert_row_clone["identifier"]
+    alert_row_clone["msg_type"] = caller
```
nursix: You are not validating this - you just trust the URL variable??

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+    tag_table = s3db.cap_area_tag
+    resource_table = s3db.cap_resource
+    unwanted_fields = s3base.s3_all_meta_field_names()
+    
+    # Copy the alert segment
+    alert_fields = [alert_table[f] for f in alert_table.fields
+                    if f not in unwanted_fields]
+    alert_row = db(alert_table.id == alert_id).select(*alert_fields,
+                                                      limitby=(0, 1)).first()    
+    alert_row_clone = alert_row.as_dict()
```
nursix: You are not checking whether alert_row actually exists - just trusting it?

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                                                    select(*location_fields)
+                    if location_rows:
+                        for location_row in location_rows:
+                            location_row_clone = location_row.as_dict()
+                            del location_row_clone["id"]
+                            location_row_clone["alert_id"] = new_alert_id
+                            location_row_clone["area_id"] = new_area_id
+                            new_location_id = location_table.insert(**location_row_clone)
+                            auth.s3_set_record_owner(location_table, new_location_id)
+                            s3db.onaccept(location_table,
```
nursix: Not quite correct - onaccept takes the form vars, i.e.:

    location_row_clone["id"] = new_location_id
    s3db.onaccept(location_table, location_row_clone)

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                resource_rows = db(query).select(*resource_fields)
+                if resource_rows:
+                    for resource_row in resource_rows:
+                        resource_row_clone = resource_row.as_dict()
+                        del resource_row_clone["id"]
+                        resource_row_clone["info_id"] = new_info_id
+                        resource_row_clone["alert_id"] = new_alert_id
+                        rid = resource_table.insert(**resource_row_clone)
+                        auth.s3_set_record_owner(resource_table, rid)
+                        s3db.onaccept(resource_table, {"resource_id": rid})
```
nursix: Same here: update the form vars with the "id" (not "resource_id"!) and pass them into onaccept.

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                                                        select(*tag_fields)
+                    if tag_rows:
+                        for tag_row in tag_rows:
+                            tag_row_clone = tag_row.as_dict()
+                            del tag_row_clone["id"]
+                            tag_row_clone["alert_id"] = new_alert_id
+                            tag_row_clone["area_id"] = new_area_id
+                            new_tag_id = tag_table.insert(**tag_row_clone)
+                            auth.s3_set_record_owner(tag_table, new_tag_id)
+                            s3db.onaccept(tag_table, {"tag_id": new_tag_id})
```
nursix: Same here - it's "id" not "tag_id", and you should pass the entire dict, not just the id (onaccept callbacks expect /all/ form vars, may otherwise not do anything)

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+            if area_rows:   
+                for area_row in area_rows:
+                    area_id = area_row.id
+                    area_row_clone = area_row.as_dict()
+                    del area_row_clone["id"]
+                    area_row_clone["alert_id"] = new_alert_id
+                    area_row_clone["info_id"] = new_info_id
+                    new_area_id = area_table.insert(**area_row_clone)
+                    auth.s3_set_record_owner(area_table, new_area_id)
+                    s3db.onaccept(area_table, {"area_id": new_area_id})
```
nursix: Same here - "id" not "area_id", and pass the entire dict, not just the id.

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+    info_rows = db(info_table.alert_id == alert_id).select(*info_fields)
+    if info_rows:
+        for info_row in info_rows:
+            info_id = info_row.id
+            info_row_clone = info_row.as_dict()
+            del info_row_clone["id"]
+            info_row_clone["alert_id"] = new_alert_id
+            new_info_id = info_table.insert(**info_row_clone)
+            auth.s3_set_record_owner(info_table, new_info_id)
+            s3db.onaccept(info_table, {"info_id": new_info_id})
```
nursix: Same here - not "info_id" but "id", and pass the entire dict.

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                resource_fields = [resource_table[f] for f in resource_table.fields
+                                   if f not in unwanted_fields]                
+                resource_rows = db(query).select(*resource_fields)
+                if resource_rows:
+                    for resource_row in resource_rows:
+                        resource_row_clone = resource_row.as_dict()
+                        del resource_row_clone["id"]
+                        resource_row_clone["info_id"] = new_info_id
+                        resource_row_clone["alert_id"] = new_alert_id
+                        rid = resource_table.insert(**resource_row_clone)
```
nursix: You're not calling update_super, but cap_resource is a doc_entity instance, so you must call update_super after insert. Otherwise you can not attach documents to this resource.

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                            tag_row_clone["alert_id"] = new_alert_id
+                            tag_row_clone["area_id"] = new_area_id
+                            new_tag_id = tag_table.insert(**tag_row_clone)
+                            auth.s3_set_record_owner(tag_table, new_tag_id)
+                            s3db.onaccept(tag_table, {"tag_id": new_tag_id})
+                
+                # Copy the resource segment
+                query = (resource_table.info_id == info_id) &\
+                        (resource_table.alert_id == alert_id)
+                resource_fields = [resource_table[f] for f in resource_table.fields
```
nursix: This is not good enough - you must not copy the doc_id (or any super-keys) either.

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
 
+# -----------------------------------------------------------------------------
+def clone_alert():
+    """
+        Function to clone the cap_alert along with its components
+            Used for the change the message status
+    """
+    
+    from s3 import s3_utc
+    caller = get_vars.get("caller")
```
nursix: You are not validating this - can set whatever I want

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+    alert_id = request.args(0)
```
nursix: You are not validating this - I can pick whatever alert I want, and don't even need any permissions. Even anonymous can clone an alert

controllers/cap.py
```diff
@@ -1069,4 +1117,120 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
                             
         return cap_area_options
 
+# -----------------------------------------------------------------------------
+def clone_alert():
```
nursix: It is, again, wrong to alter the database state with a GET request. This entire function should:

1) be an S3Method, called as /cap/alert/[id]/clone
2) be properly authorized, checking both READ permission on the record, and CREATE permission on all tables that are being written to
3) respond to POST (or PUT), but report an error for GET
4) call audit!
5) validate all inputs, especially the "caller" (which is actually: "msgtype")
6) make sure that super-keys are not duplicated, but instead update_super is called properly for new records
7) Make sure that for gis_locations, onvalidation is called (location_onvalidation is one of the few form validators that alters the data, so you must call it)

Sorry, I know - quite a lot of work before this can be merged :S

nursix:
I don't have the time right now to review everything, so just commenting on the most critical things.

Will review again once you've fixed these.

controllers/cap.py
biplovbhandari: @nursix Can you point me to the example how do we alter using POST request in eden, so that I can get started?

nursix: I think the safest variant is to make these Ajax buttons, and attach a script that produces the POST request - and if the request succeeds, then opens the new record.

An Ajax-button is like an action button, except that it has a "_ajaxurl" property instead of "url". You can set a "_class" property to attach the script (e.g. "action-btn cap-clone-update").

The script would send an Ajax-request to cap/alert/[id]/clone - and the S3Method would respond to it by cloning the record (if permitted) and, if successful, returning the new record ID to the script. Then the script would open the new record in a regular cap/alert/[id]/update view (i.e. client-side forwarding rather than server-side redirect).

Examples for S3Methods are easy to find, e.g.:
https://github.com/flavour/eden/blob/master/modules/s3db/work.py#L466 
This example is also a method handler responding to POST requests (in this case: POST work/job/[id]/signup.json and POST work/job/[id]/cancel.json). It responds with a simple JSON message, and you can put the new record ID into that JSON message by just adding it as another keyword argument.

This method handler is bound to the respective resource (work/job) here:
https://github.com/flavour/eden/blob/master/modules/s3db/work.py#L257

The corresponding button script is here:
https://github.com/flavour/eden/blob/master/static/scripts/S3/s3.work.js#L25
Of course the lookup of the record ID is a bit different because it works from a data list not data table. However, in data tables it's all the easier because you have a "db_id" attribute in the action-button that holds the record ID.

This example is still a bit incomplete (WIP) - lacks the audit-part. But you can find that in CRUD Forms:
https://github.com/flavour/eden/blob/master/modules/s3/s3forms.py#L648

So, when you have created the clone, then you should call current.audit("create", ...).

Obviously, if you use db().select, then you must check permissions manually (using current.auth.s3_accessible_query). If you use S3Resource.select instead, then you do not need to do that - the permission check would be included.

For the insert, however, you must check create-permission on the respective table (using current.auth.s3_has_permission) - there is no way to automate that.

Note that if you make this an S3Method of the cap_alert, then read-permission on the master-record is checked before the S3Method is even reached - as is the permission to access the cap/alert controller. This is one of the arguments /for/ using an S3Method.

(the other argument for making this a REST method is that it would call s3.prep (so you can configure method-specific resource settings), and can be customised in templates (using settings.customise_cap_alert_controller))

PR #1193, opened 2015-10-23

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -155,6 +155,31 @@ def alert():
                 redirect(URL(c="cap", f="template",
                              args = request.args,
                              vars = request.vars))
+            if r.record.approved_by is not None:
+                # Once approved, don't allow to edit
+                # Don't allow to delete
+                s3db.configure(tablename,
+                               editable=False,
+                               deletable=False,
+                               )
+            # Uncomment or add to this according to country CAP profile
+            #if r.record.msg_type in ("Update", "Cancel", "Error"):
+                # Use case for change in msg_type
+            #    atable = r.table
+            #    for f in ("template_id",
+            #              "sender",
+            #              "status",
+            #              "msg_type",
+            #              "source",
+            #              "scope",
+            #              "restriction",
+            #              "addresses",
+            #              "codes",
+            #              "note",
+            #              "reference",
+            #              "incidents",
+            #              ):
+            #        atable[f].writable = False
         else:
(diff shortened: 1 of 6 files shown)


controllers/cap.py
```diff
@@ -155,6 +155,31 @@ def prep(r):
...
                              args = request.args,
                              vars = request.vars))
+            if r.record.approved_by is not None:
+                # Once approved, don't allow to edit
+                # Don't allow to delete
+                s3db.configure(tablename,
+                               editable=False,
+                               deletable=False,
+                               )
+            # Uncomment or add to this according to country CAP profile
```
flavour: Won't hold up merge, but 'uncomment' won't be an easy option...either this needs to be in a template fragment (once we have cascading templates) or else a deployment_setting

modules/s3db/cap.py
```diff
@@ -1842,6 +1865,11 @@ def cap_rheader(r):
                     if action_btn:
                         rheader.insert(1, TR(TD(action_btn)))
 
+                    if msg_type_btn:
```
flavour: I won't hold up merge, but I don 't see any need for this extra var?
Why not just:
if msg_error_btn:

modules/templates/SAMBRO/views/index.html
```diff
@@ -2,11 +2,8 @@
 <div id='home'>
 <h1>{{=title}}</h1>
 <div class='subscribe-button'>
- {{if current.auth.s3_logged_in():}}
-  <a href='/{{=appname}}/cap/alert.rss' target='_blank'><i class='icon-rss'></i></a>
- {{else:}}
-  <a href='/{{=appname}}/cap/public.rss' target='_blank'><i class='icon-rss'></i></a>
- {{pass}}
+{{fn = "alert.rss" if current.auth.s3_logged_in() else "public.rss"}}
```
flavour: Slightly DRYer if the .rss is not part of the fn variable (also matches the meaning of the fn better!)

biplovbhandari:
Thanks @flavour will work on the suggestions :)

modules/s3db/cap.py
nursix: I said that using any of the buttons msg_*_btn can lead to regression if you remove one of them.

But I also said that the condition itself can be assigned to a boolean variable, not to create a separate one.

PR #1194, opened 2015-10-25
add site_id to dvr_case

tirgil:
remove organisation_id it was writes two times
add table permissions to aut_roles.csv for further using
change some rows in volunteers.csv and users.csv

Conflicts:
	modules/s3db/dvr.py


modules/s3db/dvr.py
```diff
@@ -73,7 +73,14 @@ def model(self):
...
             3: T("Closed"),
         }
+<<<<<<< HEAD
 
+=======
+        
+        org_site_represent = self.org_site_represent
+        auth = current.auth
+        
+>>>>>>> 8470873... add site_id to dvr_case
```
flavour: This conflict needs resolving before it can be merged

modules/s3db/dvr.py
```diff
@@ -88,7 +102,11 @@ def model(self):
                         represent = self.pr_PersonRepresent(show_link=True),
                         requires = IS_ADD_PERSON_WIDGET2(),
                         widget = S3AddPersonWidget2(controller="pr"),
+<<<<<<< HEAD
                      ),
+=======
+                     ),                     
+>>>>>>> 8470873... add site_id to dvr_case
```
flavour: And this one

modules/templates/Turkey/Demo/users.csv
```diff
@@ -1,5 +1,4 @@
 First Name,Last Name,Email,Password,Role,Organisation
-CISST,Admin,cisst_admin@example.com,eden,ORG_ADMIN,Ceza İnfaz Sisteminde Sivil Toplum Derneği
```
flavour: I deliberately added this one to test out x-org functionality in Volunteer Requests - I think it is useful to keep.
I changed the passwords to eden since all other templates have that as the default password for non-Admin

modules/s3db/dvr.py
```diff
@@ -73,7 +73,14 @@ def model(self):
             2: T("Pending"),
             3: T("Closed"),
         }
+<<<<<<< HEAD
 
+=======
+        
+        org_site_represent = self.org_site_represent
+        auth = current.auth
```
flavour: Since these are just used once, no need to bring into scope - just use them inline

PR #1195, opened 2015-10-25
use auth and org_site_represent inline

tirgil:
add label to site_id
add new translations

Conflicts:
	languages/tr.py
	modules/s3db/dvr.py

### DIFF
--- a/modules/s3db/dvr.py
+++ b/modules/s3db/dvr.py
@@ -81,6 +81,14 @@ class S3DVRModel(S3Model):
                            label = T("Case Number"),
                            ),
                      self.org_organisation_id(),
+                     self.super_link("site_id", "org_site",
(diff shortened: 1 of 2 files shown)


modules/s3db/dvr.py
```diff
@@ -81,6 +81,14 @@ def model(self):
                            label = T("Case Number"),
                            ),
                      self.org_organisation_id(),
+                     self.super_link("site_id", "org_site",
+                                    filterby="id",
+                                    filter_opts=current.auth.permitted_facilities(),
+                                    label=T("Place"),
```
flavour: This term isn't fitting with the rest of the system (only CRMT template uses this term).
Better to respect the deployment_setting:
label=settings.get_org_site_label()
- this defaults to Facility

PR #1196, opened 2015-10-25
change label name

tirgil:
change some translations
move dvr_need define before dvr_case
Conflicts:
	languages/tr.py
	modules/s3db/dvr.py

### DIFF
--- a/modules/s3db/dvr.py
+++ b/modules/s3db/dvr.py
@@ -57,6 +57,50 @@ class S3DVRModel(S3Model):
         crud_strings = current.response.s3.crud_strings
         define_table = self.define_table
         configure = self.configure
+        
+        # ---------------------------------------------------------------------
+        # Needs
+        #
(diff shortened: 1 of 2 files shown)


modules/s3db/dvr.py
```diff
@@ -57,6 +57,50 @@ def model(self):
         crud_strings = current.response.s3.crud_strings
         define_table = self.define_table
         configure = self.configure
+        
+        # ---------------------------------------------------------------------
+        # Needs
+        #
+        tablename = "dvr_need"
```
flavour: Not sure why you moved, but not a problem, will merge :)

flavour:
For some reason your multi-commit PRs always send up with just the commit message of the last one which is not ideal
This commit says "changed label" which isn't what it does at all (that was just an interim step).
The actual change is "Add site to dvr_case"

PR #1197, opened 2015-10-25
SAMBRO: deployment_settings for msg_type field + minor tweaks

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -162,24 +162,24 @@ def alert():
                                editable=False,
                                deletable=False,
                                )
-            # Uncomment or add to this according to country CAP profile
-            #if r.record.msg_type in ("Update", "Cancel", "Error"):
-                # Use case for change in msg_type
-            #    atable = r.table
-            #    for f in ("template_id",
-            #              "sender",
-            #              "status",
-            #              "msg_type",
-            #              "source",
-            #              "scope",
-            #              "restriction",
-            #              "addresses",
-            #              "codes",
-            #              "note",
-            #              "reference",
-            #              "incidents",
-            #              ):
-            #        atable[f].writable = False
+            if settings.get_cap_restrict_fields():
+                if r.record.msg_type in ("Update", "Cancel", "Error"):
+                    # Use case for change in msg_type
+                    atable = r.table
+                    for f in ("template_id",
+                              "sender",
+                              "status",
+                              "msg_type",
+                              "source",
+                              "scope",
+                              "restriction",
+                              "addresses",
+                              "codes",
+                              "note",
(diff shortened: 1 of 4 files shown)


modules/s3db/cap.py
```diff
@@ -1865,7 +1864,8 @@ def cap_rheader(r):
                     if action_btn:
                         rheader.insert(1, TR(TD(action_btn)))
 
-                    if msg_type_btn:
+                    if msg_error_btn:
```
nursix: Sorry but...

...your previous solution was more regression-safe.

If in subsequent versions, or under certain conditions, the "Error" button would be removed, renamed or whatever, then you will now have to fix a lot of code - whilst with the separate boolean, you can do with particular buttons whatever you want without having to worry about regression.

I would revert to the boolean. Much better style.

modules/templates/SAMBRO/views/index.html
```diff
@@ -2,8 +2,8 @@
 <div id='home'>
 <h1>{{=title}}</h1>
 <div class='subscribe-button'>
-{{fn = "alert.rss" if current.auth.s3_logged_in() else "public.rss"}}
-  <a href='/{{=appname}}/cap/{{=fn}}' target='_blank'><i class='icon-rss'></i>{{=T("RSS Feeds")}}</a>
+{{fn = "alert" if current.auth.s3_logged_in() else "public"}}
+  <a href='/{{=appname}}/cap/{{=fn}}/.rss' target='_blank'><i class='icon-rss'></i>{{=T("RSS Feeds")}}</a>
```
nursix: Should it really be "/.rss" - not just ".rss" ?

biplovbhandari: @nursix Oops sorry...

modules/s3db/cap.py
nursix: Apart from that it is technically not defined what a truthy value of an HTML object is, so correctly it would have to be: "if msg_error_btn is not None".

biplovbhandari:
@flavour Please merge

controllers/cap.py
```diff
@@ -573,7 +573,7 @@ def custom_widget_fn_3(r, **attr):
                 alert_id = request.args(0)
                 # Check for prepopulate
                 if alert_id:
-                    itable.web.default = current.deployment_settings.get_base_public_url()+\
+                    itable.web.default = deployment_settings.get_base_public_url()+\
```
flavour: We normally use the shortcut 'settings'

biplovbhandari:
@flavour Done

PR #1198, opened 2015-10-26
CAP: Avoid msg_type button regression

biplovbhandari:
@flavour Please merge


flavour:
I see no need to add this extra variable

nursix:
I see no extra variable here?

flavour:
ok, this is even 2 less :)

nursix:
Yup, and it looks cleaner than before - now all variables are initialized for all branches.

May have been a bit premature to close this.

PR #1201, opened 2015-10-29
SAMBRO: admin subscriptions

biplovbhandari:
@flavour @nursix Please review...

### DIFF
--- a/modules/templates/SAMBRO/controllers.py
+++ b/modules/templates/SAMBRO/controllers.py
@@ -214,6 +214,7 @@ class subscriptions(S3CustomController):
         if not auth.s3_logged_in():
             auth.permission.fail()
 
+        has_role = auth.s3_has_role
         # Available resources
         resources = [dict(resource="cap_alert",
                           url="cap/alert",
@@ -226,15 +227,15 @@ class subscriptions(S3CustomController):
         #        for filter widgets which need it.
         filters = [S3OptionsFilter("event_type_id",
                                    label = T("Event Type"),
-                                   options=self._options("event_type_id"),
-                                   widget="multiselect",
+                                   options = self._options("event_type_id"),
+                                   widget = "multiselect",
                                    resource = "cap_info",
                                    _name = "event-filter",
                                    ),
                    S3OptionsFilter("priority",
                                    label = T("Priority"),
-                                   options=self._options("priority"),
-                                   widget="multiselect",
+                                   options = self._options("priority"),
+                                   widget = "multiselect",
                                    resource = "cap_info",
                                    _name = "priority-filter",
                                    ),
@@ -253,6 +254,27 @@ class subscriptions(S3CustomController):
                                    ),
                    ]
 
+        if has_role("ALERT_EDITOR") or \
+           has_role("ALERT_APPROVER"):
+            from s3 import S3Represent
+            recipient_filters = [S3OptionsFilter("id",
+                                       label = T("Person"),
+                                       represent = S3Represent(lookup="auth_user",
+                                            fields = ["first_name", "last_name"],
+                                            field_sep = " ",
+                                            ),
+                                       widget = "multiselect",
+                                       resource = "auth_user",
+                                       _name = "person-filter",
+                                       ),
+                                 ]
+            # Title
+            title = T("Manage Recipients")
+        else:
+            recipient_filters = None
+            # Title
+            title = T("Subscriptions")
+
         filter_script = '''$.filterOptionsS3({
                              'trigger':'event-filter',
                              'target':'priority-filter',
@@ -262,14 +284,13 @@ class subscriptions(S3CustomController):
                              'showEmptyField': 'false'
                              })'''
         current.response.s3.jquery_ready.append(filter_script)
-        
-        # Title and view
-        title = T("Subscriptions")
+
+
+        # View
         self._view(THEME, "subscriptions.html")
 
         # Form
-        form = self._manage_subscriptions(resources, filters)
-
+        form = self._manage_subscriptions(resources, filters, recipient_filters)
         return dict(title = title,
                     form = form,
                     )
@@ -308,12 +329,13 @@ class subscriptions(S3CustomController):
         return options
 
     # -------------------------------------------------------------------------
-    def _manage_subscriptions(self, resources, filters):
+    def _manage_subscriptions(self, resources, filters, recipient_filters):
         """
             Custom form to manage subscriptions
 
             @param resources: available resources config
-            @param filters: filter widgets
+            @param filters: subscription filter widgets
+            @param recipient_filters: recipient filter widgets
         """
 
         # Uses Default Eden formstyle
@@ -326,7 +348,10 @@ class subscriptions(S3CustomController):
         T = current.T
         db = current.db
         s3db = current.s3db
+        auth = current.auth
         response = current.response
+        has_role = auth.s3_has_role
+        current_user_id = auth.user.id
         labels = Storage(
             #RESOURCES = T("Subscribe To"),
             #NOTIFY_ON = T("Notify On"),
@@ -339,20 +364,35 @@ class subscriptions(S3CustomController):
             ERROR = T("Error: could not update notification settings"),
             SUCCESS = T("Notification settings updated"),
         )
-
-        # Get current subscription settings resp. form defaults
-        subscription = self._get_subscription()
-
         # Initialize form
         form = FORM(_id="subscription-form",
                     hidden={"subscription-filters": ""})
 
-        # Filters
-        filter_form = S3FilterForm(filters, clear=False)
-        fieldset = FIELDSET(filter_form.fields(None,
-                                               subscription["get_vars"]),
-                            _id="subscription-filter-form")
-        form.append(fieldset)
+        if recipient_filters is not None:
+            # Admin Filters
+            subscription = self._get_admin_subscription()
+            recipient_subscription = self._get_recipients()
+            filter_form = S3FilterForm(filters, clear=False)
+            fieldset = FIELDSET(filter_form.fields(None,
+                                                   subscription["get_vars"]),
+                                _id="subscription-filter-form")
+            recipient_filter_form = S3FilterForm(recipient_filters, clear=False)
+            recipient_fieldset = FIELDSET(recipient_filter_form.fields(None,
+                                            recipient_subscription["get_vars"]),
+                                          _id="recipient-filter-form")
+            form.append(recipient_fieldset)
+            form.append(fieldset)
+        else:
+            # Get current subscription settings resp. form defaults
+            # Normal User
+            subscription = self._get_subscription()
+
+            # Filters
+            filter_form = S3FilterForm(filters, clear=False)
+            fieldset = FIELDSET(filter_form.fields(None,
+                                                   subscription["get_vars"]),
+                                _id="subscription-filter-form")
+            form.append(fieldset)
 
         # Notification options
         rows = []
@@ -390,64 +430,67 @@ class subscriptions(S3CustomController):
                               _id="method_selector"),
                      ""))
 
-        # Sync Row
-        properties = subscription["comments"]
-        if properties:
-            properties = json.loads(properties)
-
-        synctable = s3db.sync_repository
-        query = (synctable.apitype == "ftp") & \
-                (synctable.deleted != True) & \
-                (synctable.owned_by_user == current.auth.user.id)
-
-        ftp_rows = db(query).select(synctable.id,
-                                    synctable.name,
-                                    orderby = synctable.id)
-
-        multiselect = S3MultiSelectWidget(header = False,
-                                          multiple = False,
-                                          create = {"c": "sync",
-                                                    "f": "repository",
-                                                    "label": "Create Repository",
-                                                    },
-                                          )
-        if ftp_rows:
+        if not (has_role("ALERT_EDITOR") or \
+                has_role("ALERT_APPROVER")):
+            # Normal User
+            # Sync Row
+            properties = subscription["comments"]
             if properties:
-                user_repository_id = properties["repository_id"]
+                properties = json.loads(properties)
+
+            synctable = s3db.sync_repository
+            query = (synctable.apitype == "ftp") & \
+                    (synctable.deleted != True) & \
+                    (synctable.owned_by_user == current_user_id)
+
+            ftp_rows = db(query).select(synctable.id,
+                                        synctable.name,
+                                        orderby = synctable.id)
+
+            multiselect = S3MultiSelectWidget(header = False,
+                                              multiple = False,
+                                              create = {"c": "sync",
+                                                        "f": "repository",
+                                                        "label": "Create Repository",
+                                                        },
+                                              )
+            if ftp_rows:
+                if properties:
+                    user_repository_id = properties["repository_id"]
+                else:
+                    user_repository_id = ftp_rows.first().id
+
+                if auth.s3_has_permission("update", "sync_repository",
+                                          record_id = user_repository_id):
+                    repository_comment = S3PopupLink(c = "sync",
+                                                     f = "repository",
+                                                     m = "update",
+                                                     args = [user_repository_id],
+                                                     title = T("Update Repository"),
+                                                     tooltip = T("You can edit your FTP repository here"),
+                                                     )
+                field = s3db.sync_task.repository_id
+                ftp_ids = [(r.id, T(r.name)) for r in ftp_rows]
+                field.requires = IS_IN_SET(ftp_ids)
+
+                rows.append(("sync_task_repository_id__row",
+                             "",
+                             multiselect(field,
+                                         user_repository_id,
+                                         _id="sync_task_repository_id"),
+                             repository_comment))
             else:
-                user_repository_id = ftp_rows.first().id
+                if auth.s3_has_permission("create", "sync_repository"):
+                    repository_comment = S3PopupLink(c = "sync",
+                                                     f = "repository",
+                                                     title = T("Create Repository"),
+                                                     tooltip = T("Click on the link to begin creating your FTP repository"),
+                                                     )
 
-            if current.auth.s3_has_permission("update", "sync_repository",
-                                              record_id = user_repository_id):
-                repository_comment = S3PopupLink(c = "sync",
-                                                 f = "repository",
-                                                 m = "update",
-                                                 args = [user_repository_id],
-                                                 title = T("Update Repository"),
-                                                 tooltip = T("You can edit your FTP repository here"),
-                                                 )
-            field = s3db.sync_task.repository_id
-            ftp_ids = [(r.id, T(r.name)) for r in ftp_rows]
-            field.requires = IS_IN_SET(ftp_ids)
-
-            rows.append(("sync_task_repository_id__row",
-                         "",
-                         multiselect(field,
-                                     user_repository_id,
-                                     _id="sync_task_repository_id"),
-                         repository_comment))
-        else:
-            if current.auth.s3_has_permission("create", "sync_repository"):
-                repository_comment = S3PopupLink(c = "sync",
-                                                 f = "repository",
-                                                 title = T("Create Repository"),
-                                                 tooltip = T("Click on the link to begin creating your FTP repository"),
-                                                 )
-
-            rows.append(("sync_task_repository_id__row",
-                         "",
-                         "",
-                         repository_comment))
+                rows.append(("sync_task_repository_id__row",
+                             "",
+                             "",
+                             repository_comment))
 
         parent = FIELDSET()
 
@@ -502,7 +545,7 @@ $('#method_selector').change(function(){
 
             formvars = form.vars
 
-            listify = lambda x: None if not x else x if type(x) is list else [x]
+            #listify = lambda x: None if not x else x if type(x) is list else [x]
 
             # Fixed resource selection:
             subscription["subscribe"] = [resources[0]]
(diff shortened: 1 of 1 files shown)


modules/templates/SAMBRO/controllers.py
```diff
@@ -251,6 +252,27 @@ def __call__(self):
                                    _name = "language-filter",
                                    ),
                    ]
+        # Title
+        title = T("Subscriptions")
+
+        recipient_filters = None
+        if has_role("ALERT_EDITOR") or \
+           has_role("ALERT_APPROVER") or \
+           has_role("ADMIN"):
```
flavour: No need to check for ADMIN role here as ADMIN will get True for all roles

modules/templates/SAMBRO/controllers.py
```diff
@@ -251,6 +252,27 @@ def __call__(self):
...
+            from s3 import S3Represent
+            recipient_filters = [S3OptionsFilter("id",
+                                       label = T("Person"),
```
flavour: Indent

modules/templates/SAMBRO/controllers.py
```diff
@@ -251,6 +252,27 @@ def __call__(self):
                                    _name = "language-filter",
                                    ),
                    ]
+        # Title
+        title = T("Subscriptions")
+
+        recipient_filters = None
```
flavour: Easy enough to move this to else

modules/templates/SAMBRO/controllers.py
```diff
@@ -338,20 +362,35 @@ def _manage_subscriptions(self, resources, filters):
...
+            subscription = self._get_admin_subscription()
+            recipient_subscription = self._get_recipients()
+            filter_form = S3FilterForm(filters, clear=False)
+            fieldset = FIELDSET(filter_form.fields(None,
+                                                   subscription["get_vars"]),
+                                _id="subscription-filter-form")
+            recipient_filter_form = S3FilterForm(recipient_filters, clear=False)
+            recipient_fieldset = FIELDSET(recipient_filter_form.fields(None,
+                                            recipient_subscription["get_vars"]),
+                                      _id="recipient-filter-form")
```
flavour: indent

modules/templates/SAMBRO/controllers.py
```diff
@@ -389,64 +428,68 @@ def _manage_subscriptions(self, resources, filters):
...
-                                          multiple = False,
-                                          create = {"c": "sync",
-                                                    "f": "repository",
-                                                    "label": "Create Repository",
-                                                    },
-                                          )
-        if ftp_rows:
+        if not (has_role("ALERT_EDITOR") or \
+                has_role("ALERT_APPROVER") or \
+                has_role("ADMIN")):
```
flavour: Same as earlier - no need for ADMIN check

modules/templates/SAMBRO/controllers.py
```diff
@@ -625,6 +736,108 @@ def _get_subscription(self):
...
+                 (agtable.role == "Alert Approver") | \
+                 (agtable.role == "Administrator")
+        rows_ = db(query_).select(agtable.id)
+        editors_id = []
+        for row_ in rows_:
+            user_id = current.auth.s3_group_members(row_.id)
+            editors_id.append(user_id)
+
+        editor_ids = [item for sublist in editors_id for item in sublist]
+        query = (stable.owned_by_user.belongs(editor_ids)) & \
```
flavour: This is quite complex - I wonder if you can't use owned_by_group instead?
Seems to make for much simpler queries.
(You may well need to set the owned_by_group)

modules/templates/SAMBRO/controllers.py
```diff
@@ -736,6 +949,136 @@ def _update_subscription(self, subscription):
...
+                success = stable.insert(pe_id=pe_id,
+                                    filter_id=filter_id,
+                                    notify_on=subscription["notify_on"],
+                                    frequency=frequency,
+                                    method=subscription["method"],
+                                    comments=recipient_filter,
+                                    )
+            else:
+                success = stable.insert(pe_id=pe_id,
+                                    filter_id=filter_id,
```
flavour: indent

modules/templates/SAMBRO/controllers.py
```diff
@@ -736,6 +949,136 @@ def _update_subscription(self, subscription):
...
+                return None
+
+        # Save subscription settings
+        stable = s3db.pr_subscription
+        frequency = subscription["frequency"]
+        if subscription_id is None:
+            if recipient_filter is not None:
+                recipient_filter = json.dumps(recipient_filter)
+                success = stable.insert(pe_id=pe_id,
+                                    filter_id=filter_id,
```
flavour: indent

modules/templates/SAMBRO/controllers.py
```diff
@@ -736,6 +949,136 @@ def _update_subscription(self, subscription):
...
+            fk = '''{"subscription_id": %s}''' % subscription_id
+            for new in subscribe:
+                resource, url = new["resource"], new["url"]
+                if (resource, url) not in subscribed:
+                    # Restore subscription if previously unsubscribed, else
+                    # insert new record
+                    unsubscribed = {"deleted": True,
+                                    "deleted_fk": fk,
+                                    "resource": resource,
+                                    "url": url}
```
flavour: generally prefer to end last line with comma & have closing parenthesis on next line
(allows easier adding/commenting entries)

modules/templates/SAMBRO/controllers.py
```diff
@@ -389,64 +428,68 @@ def _manage_subscriptions(self, resources, filters):
...
-                                                    "f": "repository",
-                                                    "label": "Create Repository",
-                                                    },
-                                          )
-        if ftp_rows:
+        if not (has_role("ALERT_EDITOR") or \
+                has_role("ALERT_APPROVER") or \
+                has_role("ADMIN")):
+            # Normal User
+            # Sync Row
```
flavour: So Admins can't susbcribe via FTP?
I think this is OK...requirements I've seen didn't suggest otherwise but good to make this very explicit in the docs

modules/templates/SAMBRO/controllers.py
```diff
@@ -521,21 +564,91 @@ def _manage_subscriptions(self, resources, filters):
...
+                            # Check and delete
+                            user_id = auth.s3_get_user_id(pe_id=unselected_pe_id)
+                            rows = db(stable.pe_id == unselected_pe_id).select(\
+                                                            stable.id,
+                                                            stable.owned_by_user,
+                                                            orderby=stable.id)
+                            for row in rows:
+                                if row.owned_by_user != user_id:
+                                    # Delete the row as created by Alert Editor
+                                    s3db.resource("pr_subscription",
```
flavour: avoid get_opt in nested loop:
define_resource = s3db.resource outside of outer loop

modules/templates/SAMBRO/controllers.py
```diff
@@ -521,21 +564,91 @@ def _manage_subscriptions(self, resources, filters):
...
+                                                  id=row.id).delete()
+
+                    if pe_ids:
+                        for pe_id in pe_ids:
+                            user_id = auth.s3_get_user_id(pe_id=pe_id)
+                            query = (stable.owned_by_user != user_id) & \
+                                    (stable.pe_id == pe_id)
+                            row = db(query).select(limitby=(0, 1)).first()
+                            if row:
+                                success_subscription = self._update_admin_subscription(\
```
flavour: Again, avoid repeated getopt by bringing update_admin_subscription into scope outside loop

flavour:
Other than the relatively small tweaks suggested, this looks good - nice and DRY :)
I assume it all works when tested?

biplovbhandari:
@flavour Please review and merge

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
-                                               int(formvars.repository_id),
-                                               properties)
-                properties = json.dumps(properties)
-                db(stable.pe_id == current.auth.user.pe_id).update(comments=properties)
+            # Recipient IDs
+            user_ids = formvars["person-filter"]
+
+            if user_ids is not None:
+                current_pe_id = auth.user.pe_id
+                current_user_id = auth.s3_get_user_id(pe_id=current_pe_id)
```
flavour: current_user_id = auth.user.id is simpler

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
+                resource = s3db.resource
```
flavour: We generally use the shortcut define_resource for this...todistinguish from an instantiation:
resource = define_resource(...)

modules/templates/SAMBRO/controllers.py
flavour: Not yet actioned ;)

modules/templates/SAMBRO/controllers.py
flavour: I see auth.user.id read earlier too, so can be brought into scope up there

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
+                # Admin subscriptions
+                if len(user_ids) > 0:
+                    # Added to subscription
+                    pe_ids = [auth.s3_user_pe_id(int(user_id)) for user_id in user_ids] # Current selection
```
flavour: We probably won't have too many user_ids in the list but still better to bring the s3_user_pe_id into scope outside the list

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
+                    rows = db(stable).select(stable.pe_id)
+                    existing_pe_ids = [row.pe_id for row in rows]
+                    unselected_pe_ids = set(existing_pe_ids) - set(pe_ids)
+                    if unselected_pe_ids:
+                        for unselected_pe_id in unselected_pe_ids:
+                            # Check and delete
+                            user_id = auth.s3_get_user_id(pe_id=unselected_pe_id)
```
flavour: Again, s3_get_user_id can be brought into scope outside the loop

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
+                                elif row.owned_by_user == current_user_id:
+                                    # Self removal
+                                    resource("pr_subscription",
+                                             id=row.id).delete()
+
+                    if pe_ids:
+                        update_admin_subscription = self._update_admin_subscription
+                        for pe_id in pe_ids:
+                            if pe_id != current_pe_id:
+                                user_id = auth.s3_get_user_id(pe_id=pe_id)
```
flavour: So ensure the in scope outside loop also works for this one

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
+                                query = (stable.owned_by_user != user_id) & \
+                                        (stable.pe_id == pe_id) & \
+                                        (stable.deleted != True)
+                                row = db(query).select(limitby=(0, 1)).first()
```
flavour: I think you can limit the fields you wish to read? I only see id & filter_id used

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
+                                    success_subscription = update_admin_subscription(\
+                                                                    subscription,
+                                                                    pe_id)
+                else:
+                    # Removed from subscription
+                    rows = db(stable.deleted != True).select(stable.pe_id)
+                    pe_ids = [row.pe_id for row in rows] # All existing
+                    for pe_id in pe_ids:
+                        # Check and delete
+                        user_id = auth.s3_get_user_id(pe_id=pe_id)
```
flavour: & again...

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +563,104 @@ def _manage_subscriptions(self, resources, filters):
...
+                        rows = db(stable.pe_id == pe_id).select(stable.id,
+                                                         stable.owned_by_user)
+                        for row in rows:
+                            if row.owned_by_user != user_id:
+                                # Delete the row
+                                resource("pr_subscription",
+                                         id=row.id).delete()
+                            if row.owned_by_user == current_user_id:
```
flavour: Why have these 2 options which do exactly the same thing?

modules/templates/SAMBRO/controllers.py
```diff
@@ -737,6 +952,153 @@ def _update_subscription(self, subscription):
...
+            
+        if has_role(ADMIN):
+            group_row = db(agtable.role == "Administrator").select(agtable.id,
+                                                        limitby=(0, 1)).first()        
+        elif has_role("ALERT_EDITOR"):
+            group_row = db(agtable.role == "Alert Editor").select(agtable.id,
+                                                        limitby=(0, 1)).first()
+        elif has_role("ALERT_APPROVER"):
+            group_row = db(agtable.role == "Alert Approver").select(agtable.id,
+                                                        limitby=(0, 1)).first()
```
flavour: I don't think we need to set the group differently per role doing it...this is over-complicated.
I would just use the ALERT_EDITOR group no matter which of the 3 updates the subscription?

modules/templates/SAMBRO/controllers.py
biplovbhandari: @flavour The alert editor can also have own subscriptions...However they need to add the FTP through url...This one is for that check

biplovbhandari:
@flavour Why the check has failed? Any error in my code?

flavour:
raise ValueError("User not found")
This is admin@example.com
I did make changes in roughly this area, but I wouldn't have thought this could happen...am digging.

Doesn't appear to be your code

flavour:
User works fine when I prepop here - no idea why Travis env is different

biplovbhandari:
@flavour Yes works fine for me too

nursix:
Travis wasn't updated for the new prepop pattern, so didn't configure correctly.
Fixed.

biplovbhandari:
@flavour Please merge

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,107 @@ def _manage_subscriptions(self, resources, filters):
...
-                                               int(formvars.repository_id),
-                                               properties)
-                properties = json.dumps(properties)
-                db(stable.pe_id == current.auth.user.pe_id).update(comments=properties)
+            # Recipient IDs
+            user_ids = formvars["person-filter"]
+
+            if user_ids is not None:
+                define_resource = s3db.resource
+                user_pe_id = auth.s3_user_pe_id
```
flavour: This is no longer actually used anywhere

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,107 @@ def _manage_subscriptions(self, resources, filters):
...
+                            rows = db(stable.pe_id == unselected_pe_id).select(\
+                                                            stable.id,
+                                                            stable.owned_by_user,
+                                                            orderby=stable.id)
+                            for row in rows:
+                                if row.owned_by_user != user_id:
+                                    # Delete the row as created by Alert Editor
+                                    define_resource("pr_subscription",
+                                                    id=row.id).delete()
+                                elif row.owned_by_user == current_user_id:
```
flavour: I still don't understand this if/elif?
Both branches are the same

Seems to me that it will be more efficient to collect the IDs to delete in the loop & then do a single resource.delete() outside the loop?

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,107 @@ def _manage_subscriptions(self, resources, filters):
...
+                if len(user_ids) > 0:
+                    # Added to subscription
+                    pe_ids = [user_pe_id(int(user_id)) for user_id in user_ids] # Current selection
+                    rows = db(stable).select(stable.pe_id)
+                    existing_pe_ids = [row.pe_id for row in rows]
+                    unselected_pe_ids = set(existing_pe_ids) - set(pe_ids)
+                    if unselected_pe_ids:
+                        for unselected_pe_id in unselected_pe_ids:
+                            # Check and delete
+                            user_id = auth.get_user_id(pe_id=unselected_pe_id)
```
flavour: You bring get_user_id into scope outside the loop, but then don't actually use it inside ;)

flavour: In general if such a wrapper doesn't have a bulk option, then you may be more efficient not using it, but rather doing your own bulk method (or adding a bulk method to core)

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,107 @@ def _manage_subscriptions(self, resources, filters):
...
+            user_ids = formvars["person-filter"]
+
+            if user_ids is not None:
+                define_resource = s3db.resource
+                user_pe_id = auth.s3_user_pe_id
+                get_user_id = auth.s3_get_user_id
+                # Admin subscriptions
+                if len(user_ids) > 0:
+                    # Added to subscription
+                    pe_ids = [user_pe_id(int(user_id)) for user_id in user_ids] # Current selection
```
flavour: Likewise, I suspect a bulk conversion would be much faster than doing it 1 by 1 through the wrapper

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,107 @@ def _manage_subscriptions(self, resources, filters):
...
+                                                         stable.owned_by_user)
+                        for row in rows:
+                            if row.owned_by_user != user_id:
+                                # Delete the row
+                                define_resource("pr_subscription",
+                                                id=row.id).delete()
+                            elif row.owned_by_user == current_user_id:
+                                # Self Removal
+                                define_resource("pr_subscription",
+                                                id=row.id).delete()
```
flavour: Again, I would collate the IDs to delete inside the loop & then delete them all together outside the loop

modules/templates/SAMBRO/controllers.py
```diff
@@ -626,6 +753,91 @@ def _get_subscription(self):
...
+
+        db = current.db
+        s3db = current.s3db
+        stable = s3db.pr_subscription
+        ftable = s3db.pr_filter
+        agtable = db.auth_group
+
+        rows_ = db(agtable.role == "Alert Editor").select(agtable.id)
+        group_id = [row_.id for row_ in rows_]
+        query = (stable.owned_by_group.belongs(group_id)) & \
```
flavour: We only expect a single group_id don't we?
row = ...limitby(0,1)
owned_by_group == group_id

modules/templates/SAMBRO/controllers.py
```diff
@@ -626,6 +753,91 @@ def _get_subscription(self):
...
+    def _get_recipients(self):
+        """ Get current recipient """
+
+        db = current.db
+        agtable = db.auth_group
+        stable = current.s3db.pr_subscription
+
+        rows_ = db(agtable.role == "Alert Editor").select(agtable.id)
+        group_id = [row_.id for row_ in rows_]
+        query = (stable.owned_by_group.belongs(group_id)) & \
```
flavour: As above

modules/templates/SAMBRO/controllers.py
```diff
@@ -737,6 +950,149 @@ def _update_subscription(self, subscription):
...
+            filters_ = json.dumps(filters_)
+            if filter_id is None:
+                success = ftable.insert(pe_id=pe_id, query=filters_)
+                filter_id = success
+            else:
+                success = db(ftable.id == filter_id).update(query=filters_)
+            if not success:
+                return None
+            
+        if has_role(ADMIN) or \
```
flavour: No need for the explicit check for ADMIN

modules/templates/SAMBRO/controllers.py
```diff
@@ -737,6 +950,149 @@ def _update_subscription(self, subscription):
...
+           has_role("ALERT_EDITOR") or \
+           has_role("ALERT_APPROVER"):
+            group_row = db(agtable.role == "Alert Editor").select(agtable.id,
```
flavour: group_row only defined in this branch, so would crash outside it...hmm, this fn only gets called if the users are set already? So why check again?

flavour: Since this is a wrapper called multiple times in a loop, try to minimise the work done inside it.
Here we lookup a group_id every time!
At the minimum, cache the result with cache=s3db.cache in the select(), but I think here we can easily avoid it by doing the lookup outside (actually that is already done!) & passing group_id into the fn

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,122 @@ def _manage_subscriptions(self, resources, filters):
...
+                                                         stable.owned_by_user)
+                        # Collect IDs for those which are subscribed by 
+                        # Alert Editor
+                        r_ids = [row.id for row in rows if row.owned_by_user != user_id]
+                        # Delete these IDs
+                        define_resource("pr_subscription",
+                                        id=r_ids).delete()
+                        # Collect ID for the row self subscribed by
+                        # Alert Editor
+                        s_id = [row.id for row in rows if row.owned_by_user == current_user_id]
```
biplovbhandari: @flavour THis one is for self subscription for alert editor. He can subscribe and remove himself, but sees the admin subscription page. FOr FTP configure, need to go through url?

biplovbhandari:
@flavour Please merge

biplovbhandari:
@flavour Please merge

PR #1202, opened 2015-11-04
SAMBRO: admin subscriptions

biplovbhandari:


biplovbhandari:
@flavour Please merge

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+            user_ids = formvars["person-filter"]
+
+            if user_ids is not None:
+                agtable = db.auth_group
+                group_row = db(agtable.role == "Alert Editor").select(
+                                                        agtable.id,
+                                                        limitby=(0, 1)).first()
+                define_resource = s3db.resource
+                user_pe_id = auth.s3_user_pe_id
+                get_user_id = auth.s3_get_user_id
```
flavour: These 2 are no longer used now that you do this using your own bulk lookup fn :)

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+                if len(user_ids) > 0:
+                    # Added to subscription
+                    pe_ids = [user_pe_id(int(user_id)) for user_id in user_ids] # Current selection
+                    rows = db(stable).select(stable.pe_id)
+                    existing_pe_ids = [row.pe_id for row in rows]
+                    unselected_pe_ids = set(existing_pe_ids) - set(pe_ids)
+                    if unselected_pe_ids:
+                        for unselected_pe_id in unselected_pe_ids:
+                            # Check and delete
+                            user_id = get_user_id(pe_id=unselected_pe_id)
```
flavour: I was suggesting doing this using a bulk lookup rather than this wrapper which is ineffecient when done inside a loop

flavour: In fact you're not using this at all anyway?

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+                group_row = db(agtable.role == "Alert Editor").select(
+                                                        agtable.id,
+                                                        limitby=(0, 1)).first()
+                define_resource = s3db.resource
+                user_pe_id = auth.s3_user_pe_id
+                get_user_id = auth.s3_get_user_id
+                # Admin subscriptions
+                if len(user_ids) > 0:
+                    # Added to subscription
+                    pe_ids = [user_pe_id(int(user_id)) for user_id in user_ids] # Current selection
```
flavour: I suggested doing this using a bulk lookup fn instead of this wrapper, which is inefficient when done in a loop (it's not complex what it does, so very easy to write a bulk version...which would be usefully added to s3aaa.py)

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+                            # Delete these IDs
+                            define_resource("pr_subscription",
+                                            id=r_ids).delete()
+                    if pe_ids:
+                        current_pe_id = auth.user.pe_id
+                        update_admin_subscription = self._update_admin_subscription
+                        for pe_id in pe_ids:
+                            query = (stable.owned_by_group == group_row.id) & \
+                                    (stable.pe_id == pe_id) & \
+                                    (stable.deleted != True)
```
flavour: You can make this more efficient by having a base_query outside the loop:
base_query = (stable.owned_by_group == group_row.id) & (stable.deleted != True)
for pe_id in pe_ids:
    query = base_query & (stable.pe_id == pe_id)

modules/templates/SAMBRO/controllers.py
biplovbhandari: @flavour Any examples on bulk lookup?

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+                            row = db(query).select(stable.id,
+                                                       stable.filter_id,
```
flavour: indent

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+                                                                        pe_id,
+                                                                        group_row.id,
+                                                                        )
+                else:
+                    # Removed from subscription
+                    rows = db(stable.deleted != True).select(stable.pe_id)
+                    pe_ids = [row.pe_id for row in rows] # All existing
+                    for pe_id in pe_ids:
+                        # Check and delete
+                        user_id = get_user_id(pe_id=pe_id)
```
flavour: Again, completely unused (if it was should be using the bulk version, but seems can just be removed)

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+                        query = (stable.pe_id == pe_id) & \
+                                (stable.deleted != True)
```
flavour: base_query outside loop can be used here (less useful than the 2 element version above, but still useful)

modules/templates/SAMBRO/controllers.py
```diff
@@ -522,21 +565,102 @@ def _manage_subscriptions(self, resources, filters):
...
+                success_subscription = self._update_subscription(subscription)
+
+                # Process Sync FTP Subscription
+                if "Sync" in subscription["method"] and formvars.repository_id:
+                    properties = self._update_sync(subscription["subscribe"][0]['resource'],
+                                                   subscription.get("filters"),
+                                                   int(formvars.repository_id),
+                                                   properties)
+                    properties = json.dumps(properties)
+                    db(stable.pe_id == auth.user.pe_id).update(comments=properties)
```
flavour: current_pe_id is defined earlier...admittedly ina  branch...but why not define along with current_user_id for reuse?

modules/templates/SAMBRO/controllers.py
```diff
@@ -626,6 +748,91 @@ def _get_subscription(self):
...
+
+        row_ = db(agtable.role == "Alert Editor").select(agtable.id,
+                                                         limitby=(0, 1)).first()
+        query = (stable.owned_by_group == row_.id) & \
+                (stable.deleted != True)
+        left = ftable.on(ftable.id == stable.filter_id)
+        row = db(query).select(stable.method,
+                               ftable.query,
+                               left=left,
+                               limitby=(0, 1)).first()
```
flavour: All Admin Subscriptions are held in a single record?

biplovbhandari: @flavour They are same in each record, isn't it?

flavour: This seems odd. I would have thought that different Subscriptions would be done for different Groups?

modules/templates/SAMBRO/controllers.py
flavour: As I explained before, the fn doesn't wexist yet, but if you look at what this fn does then it should be pretty easy to write a bulk equivalent

modules/templates/SAMBRO/controllers.py
biplovbhandari: @flavour How to have groups, add members to group? As S3OptionsFilter?

modules/templates/SAMBRO/controllers.py
biplovbhandari: @flavour Sorry, but I was referring if there was any in other templates?

modules/templates/SAMBRO/controllers.py
flavour: Groups could be the answer, but we wouldn't want to create auth_group for these...pr_group would be used...which also has a pe_id in it already - so the subscriptions work directly with that rather than having to convert user_id <> pe_id.
Obviously if we're using pr_group for this then we'll need to give Admins a menu entry to manage them

flavour: NB Check with Nuwan whether he wants Groups...as I understood it, the norm was to subscribe individuals...i.e. most groups would contain 1 mmber which means that the groups concept is unnecessarily clumsy.
But each user should have separate subscriptions, right? Your query always does a limitby=(0,1) for the subscriptions table based purely on owned_by_group...which seems wrong...it should also be restricted to the relevant pe_id.
So fotr each pe_id (Person in this case) we can have 2 subscriptions: 1 which is user-controlled (owned_by_group=None) & 1 which is Admin-controlled (owned_by_group="ALERT_EDITOR") [id, not name of course ;)]

modules/templates/SAMBRO/controllers.py
```diff
@@ -626,6 +748,91 @@ def _get_subscription(self):
...
+
+        db = current.db
+        s3db = current.s3db
+        stable = s3db.pr_subscription
+        ftable = s3db.pr_filter
+        agtable = db.auth_group
+
+        row_ = db(agtable.role == "Alert Editor").select(agtable.id,
+                                                         limitby=(0, 1)).first()
+        query = (stable.owned_by_group == row_.id) & \
```
flavour: Why do this as 2 separate queries instead of a join? Join is generally more efficient for this

modules/templates/SAMBRO/controllers.py
flavour: The query looks correct here - not just owned_by & deleted, but also pe_id

modules/templates/SAMBRO/controllers.py
biplovbhandari: @flavour Actually this function is only called for admin subscriptions as <a href="https://github.com/biplovbhandari/eden/blob/master/modules/templates/SAMBRO/controllers.py#L373">here</a>

flavour: I don't even see where this fn is called to try & understand it?

modules/templates/SAMBRO/controllers.py
```diff
@@ -339,20 +364,35 @@ def _manage_subscriptions(self, resources, filters):
...
 
-        # Filters
-        filter_form = S3FilterForm(filters, clear=False)
-        fieldset = FIELDSET(filter_form.fields(None,
-                                               subscription["get_vars"]),
-                            _id="subscription-filter-form")
-        form.append(fieldset)
+        if recipient_filters is not None:
+            # Admin Filters
+            subscription = self._get_admin_subscription()
```
flavour: So here is where you call the fn which returns a single Admin Subscription.
But this is not what we need - an Admin should be able to set subscriptions for each user individually, shouldn't they?

biplovbhandari: This is how it appears to Alert Editors. They selects user and the assign the filter that is used for all selected users. Ofcourse, there is no group in here

flavour: So how do they have different users for different subscriptions?
e.g. User 1 for Area 1 & User 2 for Area 2

biplovbhandari: @flavour So we need group?

flavour: As I said before: Group is 1 option, however I think it's unnecessarily complex for this usecase where most groups would have 1 member.
& anyway that wouldn't ultimately change anything to use groups here instead of users - you still have just a single subscription option for all Admins for all Users

biplovbhandari:
@flavour Sorry for coming this late...Is it possible to allow to add more records as below in the custom form of subscriptions? This would then allow different subscriptions for different users.
ss

flavour:
I think you'll have to do something like this indeed...the create/update form should look like what you have already, but you also need a list view

biplovbhandari:
@flavour Custom Listview? Any examples in trunk?

flavour:
Why 'custom'?
Make it a std one - i.e. use the /eden/pr/subscription REST controller
You can make a little custom wrapper around it to filter to just admin subscriptions & patch the create button to the custom form (postp)
Simple wrappers abound:
https://github.com/flavour/eden/blob/master/controllers/vol.py#L30
https://github.com/flavour/eden/blob/master/controllers/vol.py#L56

modules/templates/SAMBRO/controllers.py
biplovbhandari: @flavour Sorry  for this, but when you say bulk version, what kinda fn do you mean? Also are there any examples of bulk lookup in the trunk?

flavour: bulk = not 1 by 1
So instead of:
for x in y:
a = db(table.id==x.id).select()

do:

rows = db(table.id.belongs([x.id for x in y])).select()

PR #1206, opened 2015-11-27
Changed literacy boolean into enumeration

Adendakis:


modules/s3db/pr.py
```diff
@@ -3541,6 +3541,20 @@ def model(self):
...
+        pr_literacy_opts = {
+            1: T("unknown"),
+            2: T("illiterate"),
+            3: T("literate"),
+        }
+
+        pr_literacy_status = S3ReusableField("literacy_status", "integer",
+                                            default = 1,
+                                            label = T("Literacy Status"),
+                                            represent = lambda opt: \
```
nursix: Could be:

    represent = S3Represent(options = pr_literacy_opts)

Good enough for now, can come with next push.

flavour:
I updated the Import XSL - see next commit

PR #1211, opened 2016-02-01
SAMBRO: user_info API

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/controllers.py
+++ b/modules/templates/SAMBRO/controllers.py
@@ -1425,4 +1425,62 @@ $('#method_selector').change(function(){
             current.s3db.resource("sync_repository",
                                   id=properties["repository_id"]).delete()
 
+# =============================================================================
+class user_info(S3CustomController):
+    """
+        User Info API
+        Used by Mobile Client
+    """
+
+    def __call__(self):
+
+        from gluon.http import HTTP
+        auth = current.auth
+        user = auth.user
+        if not user:
+            raise HTTP(401, body = auth.permission.fail())
+        else:
+            db = current.db
+            s3db = current.s3db
+            atable = s3db.cap_alert
+            gtable = s3db.auth_group
+            group_ids = auth.s3_get_roles(user["id"])
+            rows = db(gtable.id.belongs(group_ids)).select(gtable.uuid)
+            
+            # See the organisation of the auth.user
+            organisation = None
+            if user["organisation_id"] is not None:
+                otable = s3db.org_organisation
+                org_row = db(otable.id==user["organisation_id"]).select(\
+                                                        otable.name,
+                                                        limitby=(0, 1)).first()
+                organisation = org_row.name
+            
+            # Check the permission
+            has_permission = auth.s3_has_permission
+            permission = "0"
+            if has_permission("create", atable):
+                permission = "1"
+            if has_permission("read", atable):
+                permission += "2"
+            if has_permission("update", atable):
+                permission += "4"
+            if has_permission("delete", atable):
+                permission += "8"
+            
+            # Find the roles
+            roles = []
+            for row in rows:
+                role = {"rolename": str(row.uuid)}
+                roles.append(role)
+
+            response = {"fname": user["first_name"],
+                        "lname": user["last_name"],
+                        "organisation": organisation,
+                        "permissions": int(permission),
+                        "roles": roles
+                        }
+            current.response.headers["Content-Type"] = "application/json"
+            return json.dumps(response)
+
 # END =========================================================================
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -130,12 +130,21 @@ def alert():
 
         elif r.representation == "json":
             # @ToDo: fix JSON representation's ability to use component list_fields
-            list_fields = ["info.headline",
-                           "area.name",
-                           "info.priority",
+            list_fields = ["id",
+                           "identifier",
+                           "sender",
+                           "sent",
                            "status",
+                           "msg_type",
                            "scope",
-                           "info.event_type_id",
+                           "info.category",
+                           "info.event_type_id$name",
+                           "info.priority",
+                           "info.urgency",
+                           "info.severity",
+                           "info.certainty",
+                           "info.headline",
+                           "area.name",
                            ]
 
             s3db.configure(tablename,
@@ -694,12 +703,20 @@ def alert():
         lastid = r.resource.lastid
         if lastid and request.post_vars.get("edit_info", False):
             table = db.cap_alert
+            itable = s3db.cap_info
             alert = db(table.id == lastid).select(table.template_id,
                                                   limitby=(0, 1)).first()
+            iquery = (itable.alert_id == alert.template_id) & \
+                     (itable.deleted != True)
+            irows = db(iquery).select(itable.id)
+            iquery_ = (itable.alert_id == lastid) & \
+                      (itable.deleted != True)
+            irows_ = db(iquery_).select(itable.template_info_id)
 
-            if alert:
+            if alert and not \
+               (set([irow.id for irow in irows]) == set([irow_.template_info_id for irow_ in irows_])):
                 # Clone all cap_info entries from the alert template
-                itable = s3db.cap_info
+                # If already created dont copy again
                 unwanted_fields = set(("deleted_rb",
                                        "owned_by_user",
                                        "approved_by",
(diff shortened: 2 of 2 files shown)


modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,49 @@ def _remove_sync(self, properties):
...
+        from gluon.http import HTTP
+        request = current.request
+        if not request.env.REQUEST_METHOD == "POST":
+            raise HTTP(400, body = "Bad Request")
+        else:
+            db = current.db
+            auth = current.auth
+            utable = auth.settings.table_user
+            email = request.post_vars.get("email", None)
+            pwd = request.post_vars.get("password", None)
```
nursix: You don't need to send nor check the password with the POST vars.

If you send an HTTP BasicAuth header with the request, this will happen automatically, so you can simply check whether current.auth.user is not None (i.e. the BasicAuth was successful), and then user current.auth.user to obtain the user info.

This here is all re-inventing the wheel - and in a rather unsafe and non-standard way. I gave you the exact code structure, why don't you just follow it?

nursix:
I agree with Fran - this is a simple read action, so GET it should be.

There is no need to send any POST data to the server here - the HTTP BasicAuth header will trigger proper validation, and all you need to check in the custom controller is whether there is a current.auth.user. If that is the case, then return the data for *that* user (i.e. current.auth.user), and that's it. If there is no current.auth.user, return nothing - or if absolutely necessary, return a 401 challenge using current.auth.permission.fail() (not a 400 status - that's incorrect!).

Instead of user roles, you should send concrete permissions as booleans, like:

{...., "create_alerts": true, "approve_alerts": false, ...}

or a bitmap that encodes all relevant permissions in a single integer:

{..., "permissions": 428, ...}

This is more flexible and doesn't need to be adapted in case the role names/ids/uuids change. And anyway - the client doesn't need to know about roles, it needs to know about concrete permissions only. Hence, send only permissions - not roles.

nursix:
If you have doubts: there is nothing you need to do about the BasicAuth header on the server-side, that happens automatically. If the header is valid (i.e. username and password correct), then at the time when the custom controller is executed that user will be the current.auth.user.

If the header is not valid, then current.auth.user is None, and you can simply respond to that with a 401 challenge (auth.permission.fail()).

You do not need to implement anything else in this regard - no password check. And you shouldn't.

flavour:
> Instead of user roles, you should send concrete permissions

Actually I don't think that is needed for this particular usecase...it's about workflow rather than permissions really (the server will enforce permission checks)

nursix:
Oh ok - how do you know that? The original question said that the client application needed to know whether the user has permission to create alerts - it said nothing about roles.

And indeed if the mobile app is meant to align its workflows with what the user could do on the server, then necessarily you can not assume that a role will always mean the same if it is possible to change it in a live system.

You can very well deny a role permission to create alerts through the admin UI, but that wouldn't have any effect on the mobile app - and woops, the whole idea /why/ you request the user info is suddenly pointless because your mobile knows just as much without it :P

flavour:
> Oh ok - how do you know that?

Because I have reviewed the code in depth multiple times

> you can not assume that a role will always mean the same if it is possible to change it in a live system.

No-one will be doing that in this usecase

nursix:
Ok, so where is the code of the mobile app so that it can be reviewed in context?

And since when are roles fix for a use-case? For a deployment, maybe - or even for a certain organisational context. But for an entire use-case? Hardly.

I think that is simply inconsistent design that I can not really agree with - and you wouldn't either if I'd committed this. It's simply a breach of concept, confusing for developers, and a barrier for future changes if all server-side functionality checks permissions not roles, but a mobile client checks roles not permissions.

If that is meant to be a permanent solution, then -1 from me. Not a blocker, but strong disagreement.

The more important point was though about using a regular authentication mechanism instead of implementing a new one in that custom controller - that is more something I would insist on.

nursix:
Just out of curiousity: what does a role mean for that particular mobile workflow that is not reflected by server-side permissions? And if so, then why do the web GUI workflows not check roles, but permissions?

flavour:
> The more important point was though about using a regular authentication mechanism instead of implementing a new one in that custom controller - that is more something I would insist on.

& one I had already suggested in my review

biplovbhandari:
The roles for the mobile are just to determine whether the particular UI should be visible in mobile or not. Anyway the real permission check is always done in the server side.

biplovbhandari:
@flavour Please merge

modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,62 @@ def _remove_sync(self, properties):
...
+                                                        limitby=(0, 1)).first()
+                organisation = org_row.name
+            
+            # Check the permission
+            has_permission = auth.s3_has_permission
+            permission = "0"
+            if has_permission("create", atable):
+                permission = "1"
+            if has_permission("read", atable):
+                permission += "2"
```
flavour: Do any users not have this right?

modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,62 @@ def _remove_sync(self, properties):
...
+            if has_permission("update", atable):
+                permission += "4"
+            if has_permission("delete", atable):
+                permission += "8"
```
flavour: Do any users with Update not have Delete?
This seems a little overly-generic to me, but perhaps you do need it?

modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,62 @@ def _remove_sync(self, properties):
...
+            
+            # Find the roles
+            roles = []
+            for row in rows:
+                role = {"rolename": str(row.uuid)}
```
flavour: What value does a dict add over a simple list?
Seems to just add bloat to me...

modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,62 @@ def _remove_sync(self, properties):
...
+                roles.append(role)
+
+            response = {"fname": user["first_name"],
+                        "lname": user["last_name"],
+                        "organisation": organisation,
+                        "permissions": int(permission),
+                        "roles": roles
```
flavour: keys here are still overly long for my tastes...why not use just single letters?

modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,62 @@ def _remove_sync(self, properties):
...
+                permission += "8"
+            
+            # Find the roles
+            roles = []
+            for row in rows:
+                role = {"rolename": str(row.uuid)}
+                roles.append(role)
+
+            response = {"fname": user["first_name"],
+                        "lname": user["last_name"],
```
flavour: Why are first_name/last_name needed? What are they used for?

modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,62 @@ def _remove_sync(self, properties):
...
+                        "organisation": organisation,
```
flavour: Why is organisation name needed? What is this used for?
If there is a field in the form which gets filled from this, then why not do that server-side rather than passing back to client & thence back to server...

modules/templates/SAMBRO/controllers.py
flavour: & do we really need all Roles?
Seems to me that Authenticated isn't useful to pass back, nor is Administrator (as no Admin tasks are done by the mobile client?)
The only roles I am aware of as useful are Alert Editor & Approver (assuming approval is in-scope for the mobile client, which I'm not sure of)
I would have thought that a top-level signle letter key with a boolean 0/1 for whether they have the role would be plenty sufficient here...

modules/templates/SAMBRO/controllers.py
flavour: What do these permissions tell you which the roles don't?
Seems overkill to me for such a use-case specific client/function.

Building a generic toolkit sounds great, but then that would need to not just query 1 table, but all tables and would therefore carry far more information than you want...and it wouldn't live in SAMBRO/controllers.py but somewhere else.
So either build a generic toolkit or a SAMBRO-specific client, but don't do half-and-half.
My understanding is that this is SAMBRO-specific hence it living here & hence can be simplified to just saying whether the user is:
* Just Authenticated (i.e. we get a 200 here)
* Has Alert Editor or Publisher

so Potentially we could just have 4 possible responses:
* 401 (Authentication Required)
* 200 but empty (User is Authenticated but has no special rights in the system...so can just read Alerts (if this is even in-scope for the mobile app))
* 200 with an 'e' returned (User is an Alert Editor)
* 200 with an 'a' returned (User is an Alert Approver)

Sure, if there really are more things that are required back to the client then send them by all means, but start with the very simplest data structure & work up from there on an as-needed basis.

PR #1212, opened 2016-02-02
SAMBRO: user_info API

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/controllers.py
+++ b/modules/templates/SAMBRO/controllers.py
@@ -1425,4 +1425,39 @@ $('#method_selector').change(function(){
             current.s3db.resource("sync_repository",
                                   id=properties["repository_id"]).delete()
 
+# =============================================================================
+class user_info(S3CustomController):
+    """
+        User Info API
+        Used by Mobile Client
+    """
+
+    def __call__(self):
+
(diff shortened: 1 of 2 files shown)


modules/templates/SAMBRO/controllers.py
```diff
@@ -1425,4 +1425,40 @@ def _remove_sync(self, properties):
...
+        from gluon.http import HTTP
+        auth = current.auth
+        user = auth.user
+        if not user:
+            raise HTTP(401, body = auth.permission.fail())
+        else:
+            s3db = current.s3db
+            gtable = s3db.auth_group
+            mtable = s3db.auth_membership
+            join = gtable.on(gtable.id == mtable.group_id)
```
flavour: I've only seen this syntax before for left joins, which isn't required in this case.
Simpler to just put in the query:
query = (mtable.deleted != True) & \
             (mtable.group_id == gtable.id) & \
             (mtable.user_id == user["id"])

NB Note also that I am checking for the deleted memberships here not roles (I can't see roles being deleted & not an issue anyway as we filter these later...but deleted memberships on the other hand are critical)

biplovbhandari:
@flavour Done :)

PR #1213, opened 2016-02-04
SAMBRO: tweet alerts

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/s3cfg.py
+++ b/modules/s3cfg.py
@@ -2423,6 +2423,13 @@ class S3Config(Storage):
         """
 
         return self.cap.get("restrict_fields", False)
+    
+    def get_cap_post_to_twitter(self):
+        """
+            Whether to post the alerts in twitter
+        """
+        
+        return self.cap.get("post_to_twitter", False)
 
     # -------------------------------------------------------------------------
     # CMS: Content Management System
--- a/modules/templates/SAMBRO/config.py
+++ b/modules/templates/SAMBRO/config.py
@@ -83,6 +83,9 @@ def config(settings):
     # https://en.wikipedia.org/wiki/Filename
     # http://docs.attachmate.com/reflection/ftp/15.6/guide/en/index.htm?toc.htm?6503.htm
     settings.sync.upload_filename = "$s-%s" % ("recent_alert")
+    
+    # Whether to tweet alerts
+    settings.cap.post_to_twitter = True
 
     # -------------------------------------------------------------------------
     # L10n (Localization) settings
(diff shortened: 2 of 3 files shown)


biplovbhandari:
@flavour Done :) Please merge

biplovbhandari:
@flavour Done :) Please merge

modules/s3cfg.py
```diff
@@ -2423,6 +2423,13 @@ def get_cap_restrict_fields(self):
         """
 
         return self.cap.get("restrict_fields", False)
+    
+    def get_cap_post_to_twitter(self):
+        """
+            Whether to post the alerts in twitter
+        """
+        
+        return self.cap.get("post_to_twitter", False)
```
flavour: @ToDo: enhance this by as well as True, being able to specify a specific Twitter channel to tweet on (True => use default channel)

modules/templates/SAMBRO/config.py
```diff
@@ -233,6 +236,49 @@ def onapprove(record):
...
+                try:
+                    import tweepy
+                except ImportError:
+                    current.log.debug("tweepy module needed for sending tweets")
+                else:
+                    T = current.T
+                    db = current.db
+                    atable = s3db.cap_alert
+                    itable = s3db.cap_info
+                    send_tweet = current.msg.send_tweet
```
flavour: Still have this unnecessary intermediate variable

modules/templates/SAMBRO/config.py
```diff
@@ -233,6 +236,49 @@ def onapprove(record):
...
+                                                            atable.sent,
+                                                            limitby=(0, 1)).first()
+                    # Using English Info Segment for now
+                    # Because tweet has Info Headline
+                    iquery = (itable.alert_id == alert_id) & \
+                             (itable.language == "en-US")
+                    irow = db(iquery).select(itable.headline,
+                                             itable.web,
+                                             limitby=(0, 1)).first()
+                    twitter_text = \
```
flavour: No @ToDo: on handling the multi-message nicely? ;)

modules/s3cfg.py
biplovbhandari: @flavour If False, is there a deployment setting to assign the channel id?
for tweet, can we do send_tweet(self, text="", recipient=None, channel_id=None) to  send_tweet(self, text="", recipient=None, channel_id=None) or is there other way?

flavour: Options:
False = No tweeting
True = tweet using default channel
<name of channel>

Currently msg.send_tweet doesn't support that option, so if we get to that requirement then we would need to add it there too (easy to add, just like is already there for msg.poll_twitter)

As I say, this is an @ToDo anyway

PR #1215, opened 2016-02-07
Some more translations for German DRK projekt

armin11:
Hope this is ok for the translation - first time with git - i normally only use svn ;-)


languages/de.py
```diff
@@ -2875,7 +2917,7 @@
 'Opacity': 'Opazität (Undurchsichtigkeit)',
 'Open area': 'Offener Bereich',
 'Open recent': 'Kürzlich Bearbeitetes öffnen',
-'Open': 'Öffnen',
+'Open': 'Offen/Öffnen',
```
nursix: Since this is primarily shown on an action button, it should remain "Öffnen" - the double translation is very confusing.

languages/de.py
```diff
@@ -3991,6 +4038,7 @@
 'Subsistence Cost': 'Verpflegungskosten',
 'Suburb': 'Vorort',
 'Suggest not changing this field unless you know what you are doing.': 'Bitte ändern sie diesen Bereich nur, wenn sie ganz genau wissen was sie da tun!!!!',
+'Suitable': 'geeignet',
```
nursix: Uppercase should be uppercase.

languages/de.py
```diff
@@ -4333,6 +4382,7 @@
 'Unsafe': 'Unsicher',
 'Unselect to disable the modem': 'Abwählen um das Modem zu deaktivieren',
 'Unsent': 'Nicht gesendet',
+'Unspecified': 'unspezifiziert',
```
nursix: Uppercase should be uppercase.

nursix:
Looks good - 3 minors to fix which I can do after merge. Thanks Armin.

languages/de.py
armin11: we should alter the english title for the button "Open" into "Edit" - and
use the translation Offen for Open, cause the status "Open" is used in many
other cases ;-)

2016-02-07 18:36 GMT+01:00 Dominic König <notifications@github.com>:

nursix: söndagen den 7 februari 2016 09.41.49 skrev  Armin Retterath:
> we should alter the english title for the button "Open" into "Edit" - and
> use the translation Offen for Open, cause the status "Open" is used in many
> other cases

Easier said than done: "Open" is used as generalization for both READ and EDIT 
actions, so it can't just be changed into "EDIT" as it could also mean "READ".

More generally speaking, any differentiation between "to open" and "being open" 
is not easy to do in English - both meanings are needed in various places, 
with "to open" being the primary meaning.

And also, the fact that the English language requires context interpretation 
to understand a word while other languages use different words, is not limited 
to "open" - there are many other examples.

To make things worse, German has more than 30 different translations for the 
various meanings of "open" ;)

http://www.dict.cc/?s=open

So - however you do it, you can't do it right.

However, in Sahana the primary meaning of "open" is "to open", so we should 
not double-translate the word, and ideally use a different term in other 
contexts where possible - for example, in the shelter context, you could use 
something like "operating", "in service", "available", ...whilst there are no 
real alternatives to "to open" a record.

Dominic

nursix: Read: better to not OPEN a can of worms ;)

armin11: Ok - 'in service' seems to be a good idea ;-)
Am 07.02.2016 19:22 schrieb "Dominic König" <notifications@github.com>:

nursix:
@armin11 : seems GitHub worked just fine for you, hopefully we see more from you here? If you are interested to work on the DRK code - I do have a couple of open requirements for that which I could share with you. Pls contact me off-list for that.

PR #1216, opened 2016-02-08
1. SAMBRO: tweet alerts 2. SAMBRO: authenticated users can add multiple subscriptions  3. SAMBRO: adjust restriction textarea size 4. SAMBRO: put spaces in between multiple response type 5.CAP: move copy info button to the template

biplovbhandari:
@flavour Done :) Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -337,7 +337,7 @@ def alert():
                                    DIV(SPAN("%s :: " % T("Response Type"),
                                             _class="cap-label upper"
                                             ),
-                                       SPAN(info.response_type,
+                                       SPAN(" ".join(info.response_type),
                                             _class="cap-strong"
                                             ),
                                        ),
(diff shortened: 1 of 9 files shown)


controllers/cap.py
```diff
@@ -337,7 +337,7 @@ def custom_widget_fn_1(r, **attr):
                                    DIV(SPAN("%s :: " % T("Response Type"),
                                             _class="cap-label upper"
                                             ),
-                                       SPAN(info.response_type,
+                                       SPAN(["%s " % response_type for response_type in info.response_type],
```
flavour: A better way to do this (which doesn't end up with a trailing space):
" ".join(info.response_type)

modules/templates/SAMBRO/config.py
```diff
@@ -276,9 +320,13 @@ def customise_pr_subscription_controller(**attr):
...
         from s3 import S3CRUD
         s3 = current.response.s3
-
-        # Filter admin based subscription
-        s3.filter = (current.s3db.pr_subscription.owned_by_group != None)
+        has_role = current.auth.s3_has_role
+        # Filter admin based subscription based on roles
+        if not (has_role("ALERT_EDITOR") or \
+                has_role("ALERT_APPROVER")):
+            s3.filter = (current.s3db.pr_subscription.owned_by_user == current.auth.user.id)
```
flavour: auth = current.auth worth bringing into scope above has_role =

biplovbhandari:
@flavour Please merge

PR #1217, opened 2016-02-10
CAP:circle tool

biplovbhandari:
@flavour Done ...Please merge


static/scripts/S3/s3.gis.js
```diff
@@ -5034,6 +5067,150 @@ OpenLayers.ProxyHost = S3.Ap.concat('/gis/proxy?url=');
...
+                // Remove previous circle
+                if (map.s3.lastDraftFeature) {
+                    map.s3.lastDraftFeature.destroy();
+                } else if (draftLayer.features.length > 1) {
+                    // Clear the one from the Current Location in S3LocationSelector
+                    draftLayer.features[0].destroy();
+                }
+                /*var wkt_field = $('#gis_location_wkt');
+                if (wkt_field.length) {
+                    // Update form fields in S3LocationSelectorWidget
```
flavour: This is all still commented-out?

flavour: I guess that's OK because S3LocationSelectorWidget doesn't support Circles (yet or ever)
Then we should have appropriate comments on this: "Enable this if adding Circle tool support to S3LocationSelectorWidget"

I think the CIRCULARSTRING support should be line-commented too with another comment above that section & also a comment in the commented-out helper function

& the CIRCULAR STRING

static/scripts/S3/s3.gis.js
```diff
@@ -5034,6 +5067,150 @@ OpenLayers.ProxyHost = S3.Ap.concat('/gis/proxy?url=');
...
+                    // (S3LocationSelectorWidget2 does this in s3.locationselector.widget2.js, which is a better design)
```
flavour: & then this comment can be updated for S3LocationSelector & s3.ui.locationselector.js
Here & the other location(s): always take the opportunity to housekeep to help the next person...who may very well be ourselves in a few months' time! ;)

biplovbhandari:
@flavour Please merge

PR #1219, opened 2016-02-10
Added Feed to menu

devinbalkind:


modules/templates/CAD/menus.py
```diff
@@ -64,6 +64,9 @@ def menu_modules(cls):
         """ Custom Modules Menu """
 
         return [homepage(),
+                MM("Feed", c="cms", f="newsfeed", args="datalist",
```
nursix: Sorry - you had a bad example to copy from here: cll these args="xxx" should actually be m="xxx".

nursix: I'll cleanup after merge, so go ahead @flavour

PR #1221, opened 2016-02-11
SAMBRO: show unexpired alerts on map => match the alerts with the datalist

biplovbhandari:
@flavour Please MERGE


modules/templates/SAMBRO/controllers.py
```diff
@@ -55,9 +55,8 @@ def __call__(self):
         feature_resources = [{"name"      : T("Alerts"),
                               "id"        : "search_results",
                               "layer_id"  : layer_id,
-                              "tablename" : "cap_alert",
-                              "url"       : URL(c="cap", f=fn,
-                                                extension="geojson"),
+                              # @ToDo: Make the layer filter in the client side
```
flavour: # @ToDo: Make the filter update timestamp on refresh in the client side

PR #1222, opened 2016-02-11
Some new trasnlations for DRK

armin11:


nursix:
Looks ok to me.

Regarding that "P0 Number" string: can you tell where you see that in the UI? It's not in the code, so it would be in a database - maybe we can fix it if we understand where to look.

PR #1224, opened 2016-02-17
CAP: update other info segment field if one info segment is updated + SAMBRO: fix permission + SAMBRO: update url fix for subscription

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -609,6 +609,14 @@ def alert():
                             itable[f].writable = False
 
             elif r.component_name == "area":
+                list_fields = ["info_id",
+                               "name",
+                               "altitude",
+                               "ceiling",
+                               ]
+                s3db.configure("cap_area",
+                               list_fields = list_fields,
+                               )
                 atable = r.component.table
                 # Limit to those for this Alert
                 atable.info_id.requires = IS_EMPTY_OR(
@@ -617,7 +625,7 @@ def alert():
                                                           filterby="alert_id",
                                                           filter_opts=(r.id,),
                                                           ))
-                for f in ("event_type_id", "priority"):
+                for f in ("info_id", "event_type_id", "priority"):
                     # Do not show for the actual area
                     field = atable[f]
                     field.writable = field.readable = False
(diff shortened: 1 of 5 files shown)


controllers/cap.py
```diff
@@ -774,6 +774,41 @@ def postp(r, output):
         if r.interactive:
             if get_vars.get("_next"):
                 r.next = get_vars.get("_next")
+            if r.component_name == "info" and \
+               r.component_id is not None and \
+               len(r.post_vars) > 0:
+                # When an info segment is saved,
+                # update all other segments with non-language specific data
```
flavour: Why is this being done in postp & not onaccept?

controllers/cap.py
```diff
@@ -774,6 +774,41 @@ def postp(r, output):
...
+                                       else "%s:%02d" % (post_vars_get("onset"), 0),
+                         "expires"   : None if post_vars_get("expires") in ("", None)
+                                       else "%s:%02d" % (post_vars_get("expires"), 0),
+                         "web"       : web,
+                         }
+                
+                current_info_id = int(r.component_id)
+                table = db.cap_alert
+                itable = s3db.cap_info
+                rows = db(itable.alert_id == r.id).select(itable.id)
```
flavour: I guess we should filter out deleted segments here too?

static/themes/SAMBRO/js/homepage.js
```diff
@@ -18,5 +18,8 @@ $(document).ready(function(){
         $('#alert-filter-form, #alert-filter-form-hide').hide();
         $('#alert-filter-form-show').show();
     });
+    
+    // Change the default message for empty datalist
+    $('#cap_alert_datalist .dl-empty').html("No Current Alerts");
```
flavour: This is not localisable, so seems a very poor way to do this!
I think you can add msg_no_match here:
https://github.com/flavour/eden/blob/master/modules/s3db/cap.py#L562

This might be best done in the custom controller as the message is specific to this usecase. NB Might also be best to say T("No Current Alerts match these filters") or something...as there may well be current alerts which we choose to filter out...the 'No Current Alerts' is the filter which is hardcoded & not user-visible/selectable

PR #1225, opened 2016-02-18
little bugfix

armin11:


nursix:
Thanks, but no need for this micro-commit. Will include it in my next commit.

flavour:
Coming in the next update anyway, thanks

nursix:
:D Ok, you fix it @flavour

PR #1226, opened 2016-02-19
SAMBRO: tweaks

biplovbhandari:


modules/templates/SAMBRO/config.py
```diff
@@ -337,9 +337,13 @@ def custom_postp(r, output):
 
             if r.interactive and isinstance(output, dict):
                 # Modify Update Button
```
flavour: This now should be 'Modify Open button' of course (won't delay merge)

static/styles/S3/cap.css
```diff
@@ -24,4 +24,5 @@
 #cap_alert_sent__row1,
 #cap_alert_sent__row{display:none}
 .cap_template_form #cap_alert_template_title__row1,
-.cap_template_form #cap_alert_template_title__row{display:block}
\ No newline at end of file
+.cap_template_form #cap_alert_template_title__row{display:block}
+.embeddedComponent .map_wrapper {min-width: 350px;}
```
flavour: Nothing CAP-specific about this is there?
Wouldn't it make more sense in gis.css?

static/styles/S3/cap.css
```diff
@@ -24,4 +24,5 @@
 #cap_alert_sent__row1,
 #cap_alert_sent__row{display:none}
 .cap_template_form #cap_alert_template_title__row1,
-.cap_template_form #cap_alert_template_title__row{display:block}
\ No newline at end of file
+.cap_template_form #cap_alert_template_title__row{display:block}
```
flavour: space ;)

PR #1227, opened 2016-02-19
SAMBRO: add organisation root oid

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/config.py
+++ b/modules/templates/SAMBRO/config.py
@@ -227,6 +227,31 @@ def config(settings):
     settings.customise_org_organisation_resource = customise_org_organisation_resource
 
     # -------------------------------------------------------------------------
+    def customise_org_organisation_controller(**attr):
+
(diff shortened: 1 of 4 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -227,6 +230,30 @@ def customise_org_organisation_resource(r, tablename):
...
+        standard_prep = s3.prep
+        def custom_prep(r):
+            # Call standard prep
+            if callable(standard_prep):
+                result = standard_prep(r)
+            else:
+                result = True
+
+            table = current.s3db.org_organisation_tag
+            table.value.default = settings.get_cap_organisation_root_oid()
```
flavour: I wonder if we shouldn't enhance this by also checking that the user didn't just press return & set their org_oid to the country_oid...which I assume would be bad?

biplovbhandari: @flavour +1

PR #1228, opened 2016-02-21
CAP: link area and resource only to alert; info_id kept for importing alert

biplovbhandari:
@flavour Please MERGE

### DIFF
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -312,7 +312,7 @@ class S3CAPModel(S3Model):
                            writable = False,
                            comment = DIV(_class="tooltip",
                                          _title="%s|%s" % (T("A unique identifier of the alert message"),
-                                                           T("A number or string uniquely identifying this message, assigned by the sender. Must notnclude spaces, commas or restricted characters (< and &)."))),
+                                                           T("A number or string uniquely identifying this message, assigned by the sender. Must not include spaces, commas or restricted characters (< and &)."))),
                            ),
                      # @ToDo: Switch to using event_incident_type_id
                      Field("incidents", "list:string",
@@ -689,7 +689,7 @@ class S3CAPModel(S3Model):
                            ),
                      *s3_meta_fields())
 
-        priority_represent = S3Represent(lookup=tablename)
+        priority_represent = S3Represent(lookup=tablename, translate=True)
 
         crud_strings[tablename] = Storage(
             label_create = T("Create Warning Priority"),
@@ -995,7 +995,11 @@ class S3CAPModel(S3Model):
         define_table(tablename,
                      alert_id(writable = False,
                               ),
-                     info_id(),
+                     # Only used for imports
+                     # in CAP, resource are linked to alert
+                     info_id(readable = False,
+                             writable = False,
+                             ),
                      Field("is_template", "boolean",
                            default = False,
                            readable = False,
@@ -1008,6 +1012,7 @@ class S3CAPModel(S3Model):
                                          _title="%s|%s" % (T("The type and content of the resource file"),
                                                            T("The human-readable text describing the type and content, such as 'map' or 'photo', of the resource file."))),
                            ),
+                     # Using image field instead of doc_id because the CropWidget doesn't work in inline forms
                      Field("image", "upload",
                            label = T("Image"),
                            length = current.MAX_FILENAME_LENGTH,
@@ -1108,10 +1113,9 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
                        ]
 
         configure(tablename,
-                  # Shouldn't be required if all UI actions go through alert controller & XSLT configured appropriately
-                  create_onaccept = update_alert_id(tablename),
                   crud_form = crud_form,
                   list_fields = list_fields,
+                  onaccept = update_alert_id(tablename),
                   onvalidation = self.cap_resource_onvalidation,
                   super_entity = "doc_entity",
                   )
@@ -1161,9 +1165,10 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
         tablename = "cap_area"
         define_table(tablename,
                      alert_id(),
-                     info_id(comment = DIV(_class="tooltip",
-                                           _title="%s|%s" % (T("Information segment for this Area segment"),
-                                                             T("To which Information segment is this Area segment related. Note an Information segment can have multiple Area segments."))),
+                     # Only used for imports
+                     # in CAP, area are linked to alert
+                     info_id(readable = False,
+                             writable = False,
                              ),
                      # From which template area is the area assigned from
                      # Used for internationalisation
@@ -1276,9 +1281,10 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
 
         configure(tablename,
                   #create_next = URL(f="area", args=["[id]", "location"]),
-                  create_onaccept = update_alert_id(tablename),
-                  onvalidation = self.cap_area_onvalidation,
                   crud_form = crud_form,
+                  deduplicate = S3Duplicate(primary=("name", "alert_id")),
+                  onaccept = update_alert_id(tablename),
+                  onvalidation = self.cap_area_onvalidation,
                   )
 
         # Components
@@ -1341,8 +1347,9 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
             msg_list_empty = T("No locations currently defined for this alert"))
 
         configure(tablename,
-                  # Shouldn't be required if all UI actions go through alert controller & XSLT configured appropriately
-                  create_onaccept = update_alert_id(tablename),
+                  # @ToDo: see duplicate for alert_id & location_id
+                  deduplicate = S3Duplicate(primary=("area_id", "location_id")),
+                  onaccept = update_alert_id(tablename),
                   )
 
         # ---------------------------------------------------------------------
@@ -1386,7 +1393,7 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
                      *s3_meta_fields())
 
         configure(tablename,
-                  create_onaccept = update_alert_id(tablename),
+                  onaccept = update_alert_id(tablename),
         #         deduplicate = self.cap_area_tag_deduplicate,
                   )
 
@@ -1849,6 +1856,7 @@ def cap_rheader(r):
                                        _target="_blank",
                                        )
 
+                        has_permission = current.auth.s3_has_permission
                         # Display 'Submit for Approval', 'Publish Alert' or
                         # 'Review Alert' based on permission and deployment settings
                         if not current.request.get_vars.get("_next") and \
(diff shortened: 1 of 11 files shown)


modules/s3db/cap.py
```diff
@@ -3034,91 +3023,98 @@ def clone(self, r, **attr):
...
+                               if f not in unwanted_fields]
+            resource_query = (resource_table.alert_id == alert_id) &\
+                             accessible_query("read", resource_table)
+            resource_rows = db(resource_query).\
+                                        select(*resource_fields)
+            if resource_rows:
+                for resource_row in resource_rows:
+                    resource_row_clone = resource_row.as_dict()
+                    resource_row_clone["alert_id"] = new_alert_id
+                    rid = resource_table.insert(**resource_row_clone)
```
flavour: resource_table.insert can be brought into scope outside the loop

modules/s3db/cap.py
```diff
@@ -3034,91 +3023,98 @@ def clone(self, r, **attr):
...
+                        location_query = (location_table.area_id == area_id) &\
+                                         location_accessible
+                        location_rows = db(location_query).\
+                                            select(*location_fields)
+                        if location_rows:
+                            for location_row in location_rows:
+                                location_row_clone = location_row.as_dict()
+                                location_row_clone["alert_id"] = new_alert_id
+                                location_row_clone["area_id"] = new_area_id
+                                s3db.onvalidation(s3db.gis_location,
```
flavour: onvalidation can be brought into scope outside the loop

modules/s3db/cap.py
```diff
@@ -3034,91 +3023,98 @@ def clone(self, r, **attr):
...
+                        # Copy the area_location
+                        location_query = (location_table.area_id == area_id) &\
+                                         location_accessible
+                        location_rows = db(location_query).\
+                                            select(*location_fields)
+                        if location_rows:
+                            for location_row in location_rows:
+                                location_row_clone = location_row.as_dict()
+                                location_row_clone["alert_id"] = new_alert_id
+                                location_row_clone["area_id"] = new_area_id
```
flavour: location_row.clone.update(alert_id = new_alert_id,
                                           area_id = new_area_id)

modules/s3db/cap.py
flavour: outside outer loop

modules/s3db/cap.py
```diff
@@ -3034,91 +3023,98 @@ def clone(self, r, **attr):
...
+                                s3db.onvalidation(s3db.gis_location,
+                                                  location_row_clone)
+                                new_location_id = location_table.\
+                                                insert(**location_row_clone)
```
flavour: location_table.insert can be brought into scope outside outer loop

modules/s3db/cap.py
```diff
@@ -3034,91 +3023,98 @@ def clone(self, r, **attr):
...
+        if has_permission("create", area_table):
+            # Copy the area segment
+            area_fields = [area_table[f] for f in area_table.fields
+                           if f not in unwanted_fields_]
+            area_query = (area_table.alert_id == alert_id) & \
+                         accessible_query("read", area_table)
+            area_rows = db(area_query).select(*area_fields)
+            if area_rows:
+                has_location_permission = has_permission("create", location_table)
+                has_tag_permission = has_permission("create", tag_table)
```
flavour: Are these ever going to be different?
I think that we can use business-knowledge to avoid such multiple tests.
In the real world a user will either have create access to both or Neither

flavour: & I think both the same as the outer check (has_permission("create", area_table))
I think both of these are redundant

modules/s3db/cap.py
```diff
@@ -3034,91 +3023,98 @@ def clone(self, r, **attr):
...
+                                tag_row_clone["area_id"] = new_area_id
+
+                                new_tag_id = tag_table.insert(**tag_row_clone)
+                                # Post-process create
+                                tag_row_clone["id"] = new_tag_id
+                                audit("create", "cap", "area_tag",
+                                      record=new_tag_id)
+                                set_record_owner(tag_table, new_tag_id)
+                                onaccept(tag_table, tag_row_clone)
+        if has_permission("create", resource_table):
```
flavour: Again, I don't think we need yet another check: Real world users have all or None

modules/templates/000_config.py
```diff
@@ -216,7 +216,8 @@
 # Change for different authority and organisations
 # See http://alerting.worldweather.org/ for prefix and oid
 #settings.cap.identifier_prefix = "PAGASA-DOST"
-#settings.cap.identifier_oid = "2.49.0.1.608.0"
+# root oid, the oid for the organisation include this base with an additional suffix
```
flavour: # Country root oid. The oid for the organisation includes this base with an additional suffix

modules/templates/SAMBRO/config.py
```diff
@@ -227,6 +227,31 @@ def customise_org_organisation_resource(r, tablename):
...
+        def custom_prep(r):
+            # Call standard prep
+            if callable(standard_prep):
+                result = standard_prep(r)
+            else:
+                result = True
+
+            table = current.s3db.org_organisation_tag
+            table.tag.default = "cap_oid"
+            table.value.default = settings.get_cap_identifier_oid()
```
flavour: @ToDo: Add validation to ensure user doesn't just accept this root as this would not be acceptable

flavour: I'm actually a little uncomfortable with this. WIll ALL orgs defined in the system be Alerting Orgs?
I would suspect that some (many?) might not - they might just be responding Orgs

biplovbhandari:
@flavour please merge

biplovbhandari:
@flavour please merge

modules/s3db/cap.py
```diff
@@ -1857,45 +1863,50 @@ def cap_rheader(r):
...
+                            area_row = db(area_table.alert_id == alert_id).\
+                                                select(area_table.id,
+                                                       limitby=(0, 1)).first()
+                            if area_row:
+                                # For Alert Editor
+                                if has_permission("update", "cap_alert",
+                                                  record_id=alert_id):
+                                    # Get the user ids for the role alert_approver
+                                    agtable = db.auth_group
+                                    group_rows = db(agtable.role == "Alert Approver").\
```
flavour: This should just be one, not multiple?
So limitby=(0,1)).first() & no loop within the rows later

modules/s3db/cap.py
```diff
@@ -1857,45 +1863,50 @@ def cap_rheader(r):
...
-                                action_btn = A(T("Review Alert"),
-                                               _href = URL(args = [record.id,
-                                                                   "review"
-                                                                   ],
-                                                           ),
-                                               _class = "action-btn",
-                                               )
 
                         if record.approved_by is not None:
                             if current.auth.s3_has_permission("create", "cap_alert"):
```
flavour: Can reuse the has_permission brought into scope earlier?

modules/s3db/cap.py
```diff
@@ -3034,91 +3018,99 @@ def clone(self, r, **attr):
...
+                    location_query = (location_table.area_id == area_id) &\
+                                     location_accessible
+                    location_rows = db(location_query).\
+                                        select(*location_fields)
+                    if location_rows:
+                        for location_row in location_rows:
+                            location_row_clone = location_row.as_dict()
+                            location_row_clone.update(alert_id = new_alert_id,
+                                                      area_id = new_area_id)
+                            onvalidation(s3db.gis_location,
```
flavour: gtable can be brought into scope outside outer loop

modules/s3db/cap.py
```diff
@@ -3034,91 +3018,99 @@ def clone(self, r, **attr):
...
+            if resource_rows:
+                resource_table_insert = resource_table.insert
+                for resource_row in resource_rows:
+                    resource_row_clone = resource_row.as_dict()
+                    resource_row_clone["alert_id"] = new_alert_id
+                    rid = resource_table_insert(**resource_row_clone)
+                    resource_row_clone["id"] = rid
+                    # Post-process create
+                    audit("create", "cap", "resource", record=rid)
+                    s3db.update_super(resource_table,
```
flavour: resource_table is not an instance of a super entity, is it?
So this isn't required

biplovbhandari: @flavour But it has reference to super entity https://github.com/flavour/eden/blob/master/modules/s3db/cap.py#L1004. This case requires update_super, yes?

flavour: A reference wouldn't need this, no.
However this is partially an instance of a super entity (missing this: https://github.com/flavour/eden/blob/master/modules/s3db/doc.py#L75)
This is for the document component, which is being used...although oddly it has an image field directly embedded instead of using the image component...I guess due to widget limitations? (This is commented in the crud_form, but could also be usefully added as a comment in the model)
So then yes, it should haave update_super here, so bring into scope outside outer loop.
Also fix the missing entry in doc.py

biplovbhandari:
@flavour Please merge

modules/s3db/cap.py
```diff
@@ -1856,21 +1864,24 @@ def cap_rheader(r):
...
+                                # For Alert Editor
+                                if has_permission("update", "cap_alert",
+                                                  record_id=alert_id):
+                                    # Get the user ids for the role alert_approver
+                                    agtable = db.auth_group
+                                    group_row = db(agtable.role == "Alert Approver").\
+                                                   select(agtable.id,
+                                                          limitby=(0, 1)).first()
+                                    if group_row:
+                                        group_members = auth.s3_group_members
```
flavour: Used only once now we removed the loop? So no need to bring into scope

modules/s3db/cap.py
```diff
@@ -3034,91 +3018,99 @@ def clone(self, r, **attr):
...
+                tag_fields = [tag_table[f] for f in tag_table.fields
+                              if f not in unwanted_fields]
+
+                for area_row in area_rows:
+                    area_id = area_row.id
+                    area_row_clone = area_row.as_dict()
+                    del area_row_clone["id"]
+                    area_row_clone["alert_id"] = new_alert_id
+
+                    new_area_id = area_table.insert(**area_row_clone)
```
flavour: area_table.insert can be brought into scope outside loop

modules/s3db/cap.py
```diff
@@ -3034,91 +3018,99 @@ def clone(self, r, **attr):
...
+                    set_record_owner(area_table, new_area_id)
+                    onaccept(area_table, area_row_clone)
+
+                    # Copy the area_location
+                    location_query = (location_table.area_id == area_id) &\
+                                     location_accessible
+                    location_rows = db(location_query).\
+                                        select(*location_fields)
+                    if location_rows:
+                        gtable = s3db.gis_location
```
flavour: This is still inside the outer loop

modules/s3db/cap.py
```diff
@@ -3034,91 +3018,99 @@ def clone(self, r, **attr):
...
+                            set_record_owner(location_table,
+                                             new_location_id)
+                            onaccept(location_table,
+                                     location_row_clone)
+
+                    # Copy the area_tag
+                    tag_query = (tag_table.area_id == area_id) &\
+                                tag_accessible
+                    tag_rows = db(tag_query).select(*tag_fields)
+                    if tag_rows:
```
flavour: I don't see the value in the if here...can go straight to the loop (even if loop is empty)

biplovbhandari:
@flavour hope this is okey now

PR #1229, opened 2016-02-24

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -1282,7 +1282,7 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
         configure(tablename,
                   #create_next = URL(f="area", args=["[id]", "location"]),
                   crud_form = crud_form,
-                  deduplicate = S3Duplicate(primary=("name", "alert_id")),
+                  deduplicate = self.cap_area_duplicate,
                   onaccept = update_alert_id(tablename),
                   onvalidation = self.cap_area_onvalidation,
                   )
@@ -1330,7 +1330,7 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
                                                     show_postcode = False,
                                                     ),
                         ),
-                     )
+                     *s3_meta_fields())
 
         # CRUD Strings
         crud_strings[tablename] = Storage(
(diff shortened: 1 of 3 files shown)


modules/s3db/cap.py
```diff
@@ -1348,7 +1348,7 @@ def model(self):
 
         configure(tablename,
                   # @ToDo: see duplicate for alert_id & location_id
-                  deduplicate = S3Duplicate(primary=("area_id", "location_id")),
+                  deduplicate = self.cap_area_location_duplicate,
```
nursix: Seems daft to me - can just "is_template" to primary, no? That way you can still deduplicate templates.

modules/s3db/cap.py
```diff
@@ -1672,6 +1672,51 @@ def cap_area_onvalidation(form):
...
     # -------------------------------------------------------------------------
     @staticmethod
+    def cap_area_duplicate(item):
+
+        db = current.db
+        table = item.table
+        data = item.data
+        name = data.get("name", None)
+        if name is not None:
+            is_template = data.get("is_template", False)
```
nursix: What if is_template is True? Shouldn't a template be deduplicated against an existing template with the same name?

modules/s3db/cap.py
nursix: To me: S3Duplicate(primary=("area_id", "location_id", "is_template"))

That way, if the item is not a template, it won't match a template - but if it is a template, then it will only match templates. Makes sense?

biplovbhandari: @nursix "is_template" field is in master table cap_area. This is component table cap_area_location.

nursix: Yeah, I picked the wrong line here. See below.

My point is that you're preventing deduplication of template<=>template here too, which doesn't seem to make sense. 

Shouldn't that be so that templates match templates, and non-templates match non-templates (i.e. is_template is equal between item and row), or do you really want to never deduplicate templates?

biplovbhandari: @nursix I didnt looked the template area because some demo data that were collected from previous training have some data with same name but different location_id.

nursix: Meaning: you should prevent deduplication of template<=>non-template as well as non-template<=>template. But template<=>template seems just as desirable as non-template<=>non-template. How else would you be able to update templates via import?

You change here prevents deduplication completely if the import item is a template - and that seems to throw the baby out with the bath water. If that's deliberate, then fine of course.

biplovbhandari: @nursix Thanks for heads up! I checked data and those were differing in some parts (as those were hand drawn). Will add this check too.

PR #1230, opened 2016-02-24

biplovbhandari:


modules/s3db/cap.py
```diff
@@ -1672,6 +1671,30 @@ def cap_area_onvalidation(form):
...
+        name = data.get("name")
+        if name is not None:
+            is_template = data.get("is_template", False)
+            if is_template:
+                query = (table.name == name) & \
+                        (table.is_template == True) & \
+                        (table.event_type_id == data.get("event_type_id"))
+            else:
+                query = (table.name == name) & \
+                        (table.is_template != True) & \
```
nursix: Just out of curiosity: is "is_template" actually relevant, or does alert_id==None just equally qualify the record as a template?

I'm just wondering that because is_template!=True does not catch is_template==None, which should of course never happen due to the default, but since there is no notnull contraint for that field, it could be broken.

There is no "unbreakable code" requirement here, so you don't need to fix anything. It's more for my understanding when reviewing this.

biplovbhandari: alert_id is for sync

controllers/cap.py
```diff
@@ -614,6 +614,18 @@ def custom_widget_fn_3(r, **attr):
 
             elif r.component_name == "area":
                 atable = r.component.table
+                if db(atable.info_id != None).select().first() is not None:
```
flavour: Put the atable.id, limitby=(0,1) inside the select...otherwise we are pulling back all fields and searching through all records, rather than stopping on first match

controllers/cap.py
```diff
@@ -635,19 +647,29 @@ def custom_widget_fn_3(r, **attr):
                     s3db.configure("cap_area",
                                    editable = False,
                                    deletable = False,
```
flavour: prefer alpha-sort: deletable before editable

controllers/cap.py
```diff
@@ -635,19 +647,29 @@ def custom_widget_fn_3(r, **attr):
...
-                                   insertable=False,
+                                   insertable = False,
                                    )
 
             elif r.component_name == "resource":
-                atable = r.component.table
+                rtable = r.component.table
+                if db(rtable.info_id != None).select().first() is not None:
```
flavour: As above

controllers/cap.py
```diff
@@ -614,6 +614,18 @@ def custom_widget_fn_3(r, **attr):
 
             elif r.component_name == "area":
                 atable = r.component.table
+                if db(atable.info_id != None).select().first() is not None:
+                    # Show info_id in list-fields as we use this for import
```
flavour: This doesn't make sense to me?
"If the 1st cap_area record has an info_id then show this column in list_fields otherwise don't?"
This doesn't seem related to the areas which are part of this alert or anything.
It's a little expensive to do a DB query test here...why not just include it always if you want it sometimes?

controllers/cap.py
```diff
@@ -635,19 +647,29 @@ def custom_widget_fn_3(r, **attr):
...
                                    deletable = False,
-                                   insertable=False,
+                                   insertable = False,
                                    )
 
             elif r.component_name == "resource":
-                atable = r.component.table
+                rtable = r.component.table
+                if db(rtable.info_id != None).select().first() is not None:
+                    # Show info_id in list-fields as we use this for import
```
flavour: As above

biplovbhandari:
@flavour Please merge

modules/s3db/cap.py
```diff
@@ -1672,6 +1672,30 @@ def cap_area_onvalidation(form):
...
+        db = current.db
+        table = item.table
+        data = item.data
+        name = data.get("name")
+        if name is not None:
+            event_type_id = data.get("event_type_id", None)
+            if event_type_id is not None:
+                query = (table.name == name) & \
+                        (table.is_template == True) & \
+                        (table.event_type_id == event_type_id)
```
flavour: if you are using event_type_id to say 'this is a template' then why bother with the additional is_template check?

modules/s3db/cap.py
```diff
@@ -1672,6 +1672,30 @@ def cap_area_onvalidation(form):
...
+            else:
+                query = (table.name == name) & \
+                        (table.is_template != True) & \
+                        (table.alert_id == data.get("alert_id", None))
```
flavour: If template !=True then we absolutely should have an alert_id right?
We don't ever want to match on alert_id is None?
In which case the is_templayte is redundant & we should check we have an alert_id before bothering with the query?

biplovbhandari: @flavour I think we need this for xml import where we have info_id and not alert_id

biplovbhandari:
@flavour Please merge

modules/s3db/cap.py
flavour: hmmm, a little dangerous to rely on multiple foreign XML imports to all use unique area names.
If we don't  have an alert_id in the data, then we assume this is an XML import...and thus we have an info_id? So we can use the info_id to lookup the alert_id?

biplovbhandari:
@flavour Please merge

controllers/cap.py
```diff
@@ -614,6 +614,16 @@ def custom_widget_fn_3(r, **attr):
 
             elif r.component_name == "area":
                 atable = r.component.table
+                row = db(atable.info_id != None).select(atable.id,
+                                                        limitby=(0, 1)).first()
+                if row is not None:
```
flavour: This still doesn't make sense to me. Please see my previous comments

controllers/cap.py
```diff
@@ -633,21 +643,30 @@ def custom_widget_fn_3(r, **attr):
...
+                                   editable = False,
+                                   insertable = False,
                                    )
 
             elif r.component_name == "resource":
-                atable = r.component.table
+                rtable = r.component.table
+                row = db(rtable.info_id != None).select(rtable.id,
+                                                        limitby=(0, 1)).first()
+                if row is not None:
```
flavour: As above

controllers/cap.py
```diff
@@ -969,6 +990,21 @@ def prep(r):
         # Area create from this controller is template
         artable.is_template.default = True
 
+        if r.representation == "json":
+            # @ToDo: fix JSON representation's ability to use component list_fields
```
flavour: Perhaps better to move this comment to the field which doesn't work...what happens currently? It is ignored? Perhaps better to comment out the problem field(s)?

modules/s3db/cap.py
```diff
@@ -1587,14 +1587,15 @@ def cap_info_onaccept(form):
...
                             URL(c="cap", f="alert", args=[alert_id]))
-            idata = {"priority"  : form_vars.get("priority", None),
-                     "urgency"   : form_vars.get("urgency", None),
-                     "severity"  : form_vars.get("severity", None),
-                     "certainty" : form_vars.get("certainty", None),
-                     "effective" : form_vars.get("effective", None),
-                     "onset"     : form_vars.get("onset", None),
-                     "expires"   : form_vars.get("expires", None),
-                     "web"       : web,
+            idata = {"priority"       : form_vars.get("priority", None),
```
flavour: Since used so many times, better to move this into scope: fvget = form_vars.get

modules/s3db/cap.py
```diff
@@ -1672,6 +1673,37 @@ def cap_area_onvalidation(form):
 
     # -------------------------------------------------------------------------
     @staticmethod
+    def cap_area_duplicate(item):
+
+        db = current.db
+        table = item.table
```
flavour: Why bring these 2 into scope outside of a branch which may end up not using them?

modules/s3db/cap.py
```diff
@@ -1672,6 +1673,37 @@ def cap_area_onvalidation(form):
...
+        data = item.data
+        name = data.get("name")
+
+        if name is not None:
+            event_type_id = data.get("event_type_id", None)
+            if event_type_id is not None:
```
flavour: A comment explaining this would be useful:
e.g. 
'# This must be a template

modules/s3db/cap.py
```diff
@@ -1672,6 +1673,37 @@ def cap_area_onvalidation(form):
...
+                query = (table.name == name) & \
+                        (table.event_type_id == event_type_id)
+            else:
+                alert_id = data.get("alert_id", None)
+                info_id = data.get("info_id", None)
+                query_ = (table.name == name)
+                if alert_id is not None:
+                    query = query_ & (table.alert_id == alert_id)
+                elif info_id is not None:
```
flavour: Comment: CAP XML Import

modules/s3db/cap.py
```diff
@@ -1672,6 +1673,37 @@ def cap_area_onvalidation(form):
...
+        if name is not None:
+            event_type_id = data.get("event_type_id", None)
+            if event_type_id is not None:
+                query = (table.name == name) & \
+                        (table.event_type_id == event_type_id)
+            else:
+                alert_id = data.get("alert_id", None)
+                info_id = data.get("info_id", None)
+                query_ = (table.name == name)
+                if alert_id is not None:
```
flavour: Comment: Real Alert, not Termplate

modules/s3db/cap.py
```diff
@@ -1672,6 +1673,37 @@ def cap_area_onvalidation(form):
...
+                    query = query_ & (table.alert_id == alert_id)
+                elif info_id is not None:
+                    query = query_ & (table.info_id == info_id)
+                else:
+                    return
```
flavour: Comment: Nothing we can use

modules/s3db/cap.py
```diff
@@ -1795,14 +1827,14 @@ def cap_rheader(r):
...
             s3db = current.s3db
             tablename = r.tablename
             if tablename == "cap_alert":
                 alert_id = record.id
                 itable = s3db.cap_info
-                row = current.db(itable.alert_id == alert_id).\
-                                        select(itable.id,
-                                               limitby=(0, 1)).first()
+                row = db(itable.alert_id == alert_id).select(itable.id,
+                                                             limitby=(0, 1)).first()
```
flavour: Why do this query outside a branch & hence it might not be used?

flavour: (Same for the 2 bring into scopes)

modules/templates/SAMBRO/config.py
```diff
@@ -43,6 +43,14 @@ def config(settings):
     # GeoNames username
     settings.gis.geonames_username = "eden_test"
 
+    # Restrict the Location Selector to just certain countries
+    # NB This can also be over-ridden for specific contexts later
+    # e.g. Activities filtered to those of parent Project
+    #settings.gis.countries = ("MM",)
+
+    # Default timezone for users
+    #settings.L10n.utc_offset = "+0700"
```
flavour: I don't think these make sense in SAMBRO/config.py since we have multiple sites sharing the same template.
Instead create separate templates for each of the SAMBRO deployments which have these deployment-specific things in.
We have cascading templates, so in 000_config.py can set: settings.base.template = ("locations.MM", "SAMBRO"))
Can also add a "SAMBRO.MM" at the end if there are SAMBRO-specific things to change, but generic ones like the timezone should be in core locations.

modules/s3db/cap.py
biplovbhandari: used in else also

PR #1231, opened 2016-02-26

biplovbhandari:
@flavour I hope this is okey to merge now..


modules/s3/s3notify.py
```diff
@@ -404,7 +404,7 @@ def get_template(path, filenames):
                 path = join("views", "msg")
                 template = get_template(path, filenames)
             if template is None:
-                template = StringIO(T("New updates are available."))
+                template = StringIO("New updates are available.")
```
flavour: So this is deliberately not localised?

modules/s3db/cap.py
```diff
@@ -1795,14 +1831,14 @@ def cap_rheader(r):
...
+            db = current.db
             s3db = current.s3db
             tablename = r.tablename
             if tablename == "cap_alert":
                 alert_id = record.id
                 itable = s3db.cap_info
-                row = current.db(itable.alert_id == alert_id).\
-                                        select(itable.id,
-                                               limitby=(0, 1)).first()
+                row = db(itable.alert_id == alert_id).select(itable.id,
```
flavour: This query is still outside a branch - so it might be run for no reason (still useful, but less-critical for the 2x bring into scope above, but a DAL query is always relatively expesnive so more important to optimise)

modules/s3db/cap.py
```diff
@@ -1937,14 +1972,18 @@ def cap_rheader(r):
...
                         # Check to see if 'Predefined Areas' tab need to be added
-                        artable = s3db.cap_area
-                        query = (artable.is_template == True) & \
-                                (artable.deleted == False)
-
-                        template_area_rows = current.db(query)._select(artable.id,
-                                                                       limitby=(0, 1))
-                        if template_area_rows:
-                            tabs.insert(2, (T("Predefined Areas"), "assign"))
+                        itable = s3db.cap_info
```
flavour: itable was already brought into scope earlier

modules/s3db/cap.py
```diff
@@ -1937,14 +1972,18 @@ def cap_rheader(r):
                        current.auth.s3_has_permission("update", "cap_alert",
                                                       record_id=alert_id):
                         # Check to see if 'Predefined Areas' tab need to be added
```
flavour: I prefer:
''Show predefined areas tab if we have some defined"

modules/s3db/cap.py
biplovbhandari: @flavour that is because there is else branch that is also using it

flavour: ok, yes, looked at whole file - big function!

modules/s3db/cap.py
```diff
@@ -1795,14 +1831,14 @@ def cap_rheader(r):
...
             if tablename == "cap_alert":
                 alert_id = record.id
                 itable = s3db.cap_info
-                row = current.db(itable.alert_id == alert_id).\
-                                        select(itable.id,
-                                               limitby=(0, 1)).first()
+                row = db(itable.alert_id == alert_id).select(itable.id,
+                                                             limitby=(0, 1)).first()
                 if record.is_template:
                     if not (row and row.id):
```
flavour: When is row True but not row.id?
I don't believe such a case exists? (I see this twice & it seems an unnecessary check to me)

PR #1232, opened 2016-02-29

biplovbhandari:
@flavour Please MERGE


modules/s3db/cap.py
```diff
@@ -706,6 +711,7 @@ def model(self):
             )
 
         configure(tablename,
+                  create_onaccept = self.cap_warning_priority_onaccept,
                   deduplicate = S3Duplicate(primary=("event_type_id", "name")),
```
nursix: This seems out of place - no event_type_id in this table, and no deduplicator will be reached if you have a unique key among the fields.

modules/s3db/cap.py
```diff
@@ -632,8 +633,10 @@ def model(self):
                                          _title="%s|%s" % (T("Event Code"),
                                                            T("Code (key) for the event like for eg. (2001, Typhoon), (2002, Flood)"))),
                            ),
-                     Field("name", notnull=True, length=64,
+                     Field("name", notnull=True, length=64, unique=True,
```
nursix: unique=True must me supplemented with an IS_NOT_ONE_OF validator, otherwise violations of the constraints during imports can break the entire import.

Also note that adding a unique-constraint will require a manual DB migration (can't just add the constraint)

modules/s3db/cap.py
```diff
@@ -632,8 +633,10 @@ def model(self):
...
                            label = T("Name"),
+                           requires = IS_MATCH('^[^"]+$',
```
nursix: This is not enough - validator must also catch the notnull/unique constraints (PostgreSQL will stop the entire transaction on violations, not just the current query, so this would break imports).

modules/s3db/cap.py
biplovbhandari: @nursix I would love to document these tips by you... Do we have a particular wiki page for this?

nursix: Nothing Sahana-specific here, just common knowledge/logic:

Adding a unique-constraint to a legacy database will fail if there are non-unique values in the field. That is why it does not auto-migrate - and you will have to sort this out by hand.

PostgreSQL doesn't do nested transactions, so if one command fails, no further commands can be added to the same transaction. Hence you must validate data against all constraints /before/ sending the command.

I don't think that documenting that on the Sahana wiki would add a lot of value - developers usually know that.

nursix: ...and if they don't know it, they wouldn't search the wiki for it.

The problem is that people always need a reason to search the wiki for information - and for these two things there is no such reason unless they are aware of these things. But if you already are aware, then there's no reason to search either ;) as the question is already the answer.

biplovbhandari:
@flavour Please merge

modules/s3db/cap.py
```diff
@@ -632,8 +633,11 @@ def model(self):
                                          _title="%s|%s" % (T("Event Code"),
                                                            T("Code (key) for the event like for eg. (2001, Typhoon), (2002, Flood)"))),
                            ),
-                     Field("name", notnull=True, length=64,
+                     Field("name", notnull=True, length=64, unique=True,
                            label = T("Name"),
+                           requires = [IS_NOT_ONE_OF(db, "%s.name" % tablename),
+                                       IS_MATCH('^[^"]+$',
+                                               error_message=current.T('Cannot be empty and Must not include "'))],
```
flavour: No need for current here: T is in scope

modules/s3db/cap.py
```diff
@@ -632,8 +633,11 @@ def model(self):
...
                            comment = DIV(_class="tooltip",
                                          _title="%s|%s" % (T("Name"),
                                                            T("The actual name for the warning priority, for eg. Typhoons in Philippines have five priority name (PSWS# 1, PSWS# 2, PSWS# 3, PSWS# 4 and PSWS# 5)"))),
```
flavour: five priority names

modules/s3db/cap.py
```diff
@@ -1708,6 +1751,38 @@ def cap_resource_onvalidation(form):
...
+    def cap_warning_priority_onaccept(form):
+
+        form_vars = form.vars
+        color_code = form_vars.color_code
+        if color_code:
+            db = current.db
+            s3db = current.s3db
+            stable = s3db.gis_style
+            etable = s3db.gis_layer_entity
+            rows = db(etable.instance_type == "gis_layer_feature"). \
```
flavour: This assumes that all feature Layers are CAP ones.
This is only true within the SAMBRO template, not all possible Sahana deployments which enable CAP

flavour: I suggest querying the gis_layer_feature table instead & filtering for c==cap
This covers both private & public Alerts
Could consider filtering to f==public or alert but I think that we can assume that no other cap resources are going to be put on the map

modules/templates/SAMBRO/config.py
```diff
@@ -254,29 +255,29 @@ def onapprove(record):
...
-                    try:
-                        current.msg.send_tweet(text=twitter_text)
-                    except tweepy.error.TweepError, e:
-                        current.log.debug("Sending tweets failed: %s" % e)
+                        try:
+                            current.msg.send_tweet(text=twitter_text)
+                        except tweepy.error.TweepError, e:
+                            current.log.debug("Sending tweets failed: %s" % e)
+                        except UnicodeDecodeError, e:
+                            current.log.debug("Unicode Decode Error in tweet string: %s" % e)
```
flavour: Should this not crash the system?
I am very wary of hiding this error here...

biplovbhandari:
@flavour Done please merge

PR #1234, opened 2016-03-02
SAMBRO: allow self subscription for editors

biplovbhandari:


modules/s3db/pr.py
```diff
@@ -6673,6 +6674,24 @@ def pr_get_pe_id(entity, record_id=None):
...
+def pr_get_group_pe_id(group_ids):
+    """
+        Get the PE-ID for list of group_ids
+
+        @param group_ids: the list of group_ids
+
+        @return: dictionary of pe_id and group_id
+    """
+
+    if len(group_ids) > 0:
```
flavour: I would generally say:
if group_ids:

This is both simpler and catches the case where the input is None.

(Can be fixed post-merge)

modules/templates/SAMBRO/config.py
```diff
@@ -319,14 +319,59 @@ def customise_pr_subscription_controller(**attr):
         from gluon.html import URL, A
         from s3 import S3CRUD
         s3 = current.response.s3
+        s3db = current.s3db
         auth = current.auth
+        request = current.request
```
flavour: Only used once? No need for interim variable

modules/templates/SAMBRO/config.py
```diff
@@ -336,19 +381,46 @@ def custom_postp(r, output):
...
+                                                            )
+                                          )
+
                 if "form" in output:
-                    add_btn = A(T("Create Subscription"),
-                                _class="action-btn",
-                                _href=URL(c="default", f="index", args=["subscriptions"])
-                                )
+                    # Modify Add Button
+                    if request.get_vars["option"] == "manage_recipient" and \
```
flavour: if request.get_vars["option"] == "manage_recipient" is being checked twice (actually 3 times since once outside the postp) So better to check once & then cache the result.
NB To pass into the postp requires a trick:
def custom_postp(r, output, manage_recipient=manage_recipient)

modules/templates/SAMBRO/controllers.py
```diff
@@ -626,240 +630,321 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
+                    # People
+                    for pe_id in user_pe_ids.values():
+                        user_id = [key for key, value in user_pe_ids.iteritems()
+                                   if value == pe_id][0]
+                        query = base_query & \
+                                (stable.pe_id == pe_id)
+                        rows = db(query).select(stable.id,
+                                                stable.filter_id,
+                                                ftable.query)
+                        if len(rows) > 0:
```
flavour: if rows:
simpler

modules/templates/SAMBRO/config.py
```diff
@@ -464,4 +536,82 @@ def custom_postp(r, output):
...
+            # Self Subscription
+            query = (stable.deleted != True) & \
+                    (stable.owned_by_group == None) & \
+                    (stable.owned_by_user == auth.user.id)
+
+        left = ftable.on(ftable.id == stable.filter_id)
+        rows = db(query).select(stable.filter_id,
+                                ftable.query,
+                                left=left)
+        if len(rows) > 0:
```
flavour: if rows:
simpler

PR #1236, opened 2016-03-05
SAMBRO: Tweaks

biplovbhandari:
@flavour please merge


controllers/sync.py
```diff
@@ -245,4 +245,12 @@ def prep(r):
                                 )
     return output
 
+# -----------------------------------------------------------------------------
+def task():
```
nursix: What is this needed for?

Ideally, there should be no independent controller here - or at least it must be restricted to exactly what it is needed for.

modules/s3layouts.py
```diff
@@ -517,10 +517,15 @@ def layout(item):
...
+        if item.method == "update":
+            _id = "%s_update" % item.function
+        else:
+            _id = "%s_add" % item.function
+
         popup_link = A(label,
                        _href=item.url(format="popup"),
                        _class="s3_add_resource_link",
-                       _id="%s_add" % item.function,
+                       _id=_id,
```
nursix: Wouldn't this be better expressed as class? Node ids can easily clash.

modules/templates/SAMBRO/config.py
```diff
@@ -430,6 +460,71 @@ def custom_postp(r, output):
...
+
+            @param resource: the S3Resource
+            @param data: the data returned from S3Resource.select
+            @param meta_data: the meta data for the notification
+            @param format: the contents format ("text" or "html")
+        """
+
+        if format == "text" or \
+           format == "html":
+            # Both text or html
```
nursix: You did not get my point when I explained this: the reason to pre-render the records in Python is because it is easier to debug than a view template. 

Yet you're doing it exactly the other way around, moving the rendering of the rows into the view template.

What's the reason for that?

modules/s3db/pr.py
```diff
@@ -6684,7 +6684,7 @@ def pr_get_group_pe_id(group_ids):
         @return: dictionary of pe_id and group_id
     """
 
-    if len(group_ids) > 0:
+    if group_ids > 0:
```
nursix: Now that doesn't make sense to me.

All of `if group_ids` or `if len(group_ids)` or `if len(group_ids) > 0` have a point, but comparing a list against 0 seems wrong.

modules/templates/SAMBRO/controllers.py
```diff
@@ -535,18 +537,21 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
+                                                     vars = dict(prefix = "sync"),
                                                      title = T("Add FTP Directory and Authentication Details"),
                                                      tooltip = T("Click on the link to begin creating your FTP repository and setting authentication details"),
                                                      )
-
+                    field = s3db.sync_task.repository_id
                 rows.append(("sync_task_repository_id__row",
                              "",
-                             "",
+                             multiselect(field, "",
```
nursix: "field" not defined if permission check fails

modules/templates/SAMBRO/controllers.py
```diff
@@ -711,7 +720,7 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
                         rows = db(query).select(stable.id,
                                                 stable.filter_id,
                                                 ftable.query)
-                        if len(rows) > 0:
+                        if rows:
```
nursix: This entire nested loop over queries can be done in a much better way - this is complicated, fragile and inefficient. 

And why is it repeated, for Groups/Persons separately? DRY

modules/s3db/cap.py
```diff
@@ -2831,8 +2859,12 @@ def apply_method(self, r, **attr):
                     area_id = int(area_id.strip())
                     add_area_from_template(area_id, alert_id)
                     added += 1
-            current.session.confirmation = T("%(number)s assigned") % \
-                                                            {"number": added}
+            if added == 1:
+                current.session.confirmation = T("%(number)s area assigned") % \
```
nursix: Bit pointless to construct with %(number)s here - it'll be exactly 1 according to the condition.

modules/s3db/cap.py
```diff
@@ -1982,23 +1989,44 @@ def cap_rheader(r):
...
+                                                       limitby=(0, 1)).first()
+                                if row.organisation_id == row_.organisation_id:
+                                    msg_type_buttons = TAG[""](
+                                            TR(TD(A(T("Update Alert"),
+                                                    _id = record.id,
+                                                    _data = "Update",
+                                                    _class = "action-btn cap-clone-update",
+                                                    ))),
+                                            TR(TD(A(T("Cancel Alert"),
+                                                    _id = record.id,
```
nursix: This is not valid - all anchors have the same node ID. Node IDs must be unique in the document.

modules/s3db/cap.py
```diff
@@ -1982,23 +1989,44 @@ def cap_rheader(r):
...
-                                                _data = "Error",
-                                                _class = "action-btn cap-clone-update",
-                                                ))),
-                                        )
+                                utable = db.auth_user
+                                row = db(utable.id == record.created_by).\
+                                        select(utable.organisation_id,
+                                               limitby=(0, 1)).first()
+                                row_ = db(utable.id == current.auth.user.id).\
+                                                select(utable.organisation_id,
```
nursix: Inconsistent indentation - makes it hard to read. Why have 2 queries here? Should be just one.

modules/templates/SAMBRO/config.py
```diff
@@ -374,14 +404,14 @@ def custom_prep(r):
 
         # Custom postp
         standard_postp = s3.postp
-        def custom_postp(r, output):
+        def custom_postp(r, output, option=option):
```
nursix: No point passing in "option" - is in scope anyway.

modules/templates/SAMBRO/controllers.py
```diff
@@ -523,9 +519,15 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
                                                      f = "repository",
                                                      m = "update",
                                                      args = [user_repository_id],
+                                                     vars = dict(prefix = "sync"),
```
nursix: {"prefix": "sync"}

Shouldn't use the dict constructor if all parameters are constants anyway (less efficient)

modules/templates/SAMBRO/views/msg/notify_email_html.html
```diff
@@ -0,0 +1,94 @@
+<html>
+<body>
+{{T = current.T}}
+{{from s3 import s3_str}}
+{{if new is not None:}}{{new_title=T("The following Alert has been added") if new == 1 else T("The following %(number)s Alerts have been added") % dict(number=new)}}
+{{=XML(new_title)}}
+<br/><br/>
+{{for record in new_records:}}{{new_alert_record=dict(record)}}
+{{=XML("%(line)s" % dict(line="--------------------------------------------------------------"))}}
```
nursix: This is a really strange string construction - what's the point here?

modules/templates/SAMBRO/controllers.py
nursix: Why have a local variable at all if single-use?

modules/s3db/cap.py
```diff
@@ -1982,23 +1989,44 @@ def cap_rheader(r):
...
+                                                    _class = "action-btn cap-clone-update",
+                                                    ))),
+                                            TR(TD(A(T("All Clear"),
+                                                    _id = record.id,
+                                                    _data = "AllClear",
+                                                    _class = "action-btn cap-clone-update",
+                                                    ))),
+                                            )
+                                else:
+                                    arow = db(r.table.id == r.id).select(r.table.msg_type,
```
nursix: r.record.msg_type? Do you really need an additional query here?

PR #1257, opened 2016-03-07
SAMBRO: HTML Emails/Relay Alerts

biplovbhandari:
@flavour Please merge.....


controllers/sync.py
```diff
@@ -245,4 +245,20 @@ def prep(r):
                                 )
     return output
 
+# -----------------------------------------------------------------------------
+def task():
+    """ RESTful CRUD controller for options.s3json lookups """
+
+    if auth.permission.format != "s3json":
```
nursix: "" is not a valid response for a popup link.

Like this:

    s3.prep = lambda r: r.representation == "s3json" and r.method == "options"

modules/s3db/pr.py
```diff
@@ -6729,7 +6729,7 @@ def pr_get_group_pe_id(group_ids):
         @return: dictionary of pe_id and group_id
     """
 
-    if len(group_ids) > 0:
+    if group_ids:
```
nursix: Ok, though I actually think that this function isn't needed. The case can be better solved with a join.

nursix: In fact - where is this used?

nursix: As much as I can see you're iterating through the subscriptions and compare the pe_id of the subscriber to the group_ids - but why not select just those subscriptions belonging to the respective groups/users in the first place? This can be solved via a join, and then you do not need either of the bulk lookup functions.

Besides, an intelligent join will remove the deep nesting in the function, making it more readable and less fragile.

There is really no common case for this function, and no plausible one can be constructed. I'd recommend to either move this into SAMBRO, or (much better) solve this with an intelligent join that doesn't require loop+compare.

modules/s3db/cap.py
```diff
@@ -1982,23 +1989,41 @@ def cap_rheader(r):
...
-                                        TR(TD(A(T("Cancel Alert"),
-                                                _id = record.id,
-                                                _data = "Cancel",
-                                                _class = "action-btn cap-clone-update",
-                                                ))),
-                                        TR(TD(A(T("Error Alert"),
-                                                _id = record.id,
-                                                _data = "Error",
+                                relay_alert = A(T("Relay Alert"),
+                                                _data = "%s/%s" % (record.msg_type, r.id),
```
flavour: So Relay can now be used by all users, even within same Org, right?
I'll assume a deliberate change...

modules/s3db/cap.py
```diff
@@ -1982,23 +1989,41 @@ def cap_rheader(r):
...
+                                                )
+                                if record.created_by:
+                                    utable = db.auth_user
+                                    row = db(utable.id == record.created_by).select(\
+                                                            utable.organisation_id,
+                                                            limitby=(0, 1)).first()
+                                    row_ = db(utable.id == current.auth.user.id).select(\
+                                                            utable.organisation_id,
+                                                            limitby=(0, 1)).first()
+                                    if row.organisation_id == row_.organisation_id:
```
flavour: So yes this is really where we should be using realms now, but I won't hold up the merge for it.
Users would need an ANY read & a write just for their realms...then can check for write access ('update') to show these buttons

modules/s3db/cap.py
biplovbhandari: @flavour No I just defined this here as this is used more than once

modules/templates/SAMBRO/config.py
```diff
@@ -430,6 +459,103 @@ def custom_postp(r, output):
...
+        created_on_selector = resource.prefix_selector("created_on")
+        created_on_colname = None
+        notify_on = meta_data["notify_on"]
+        last_check_time = meta_data["last_check_time"]
+        rows = data["rows"]
+        rfields = data["rfields"]
+        output = {}
+        new, upd = [], []
+        if format == "text":
+            # Standard text format
```
flavour: Nice to add the 'for SMS' here
(Won't hold up merge)

modules/s3db/cap.py
biplovbhandari: @flavour Thats why I asked you...This seems to solve the case for now...Shall we move to realm when it is needed as you mentioned in your email earlier?

modules/templates/SAMBRO/config.py
```diff
@@ -587,7 +712,7 @@ def pr_subscription_filter_row_options():
                         # Get the value for prefix
                         values = filter[1].split(",")
                         if prefix == "event_type_id__belongs":
-                            event_type_id = int(s3_str(values[0]))
+                            event_type_id = s3_str(values[0])
                             row_ = db(etable.id==event_type_id).select(\
```
flavour: Should have PEP8 spacing:
etable.id == event_type_id
(Won't hold up merge)

modules/templates/SAMBRO/controllers.py
```diff
@@ -643,12 +641,15 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
                 update_admin_subscription = self._update_admin_subscription
-                if group_ids is not None and \
-                   len(group_ids) > 0 and \
-                   len(user_ids) > 0:
-                    # Both Individuals & Groups selected
-                    user_pe_ids = bulk_user_pe_id(user_ids) # Current Selection
-                    group_pe_ids = pr_get_group_pe_id(group_ids)
+                if group_ids is None:
+                    group_ids = []
+                if (len(group_ids) > 0 or len(user_ids) > 0):
```
flavour: No need for the > 0 here...easier to read without.
(Won't hold up merge)

modules/templates/SAMBRO/controllers.py
```diff
@@ -643,12 +641,15 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
+                    user_pe_ids = None
+                    group_pe_ids = None
+                    if len(user_ids) > 0:
```
flavour: As above

modules/templates/SAMBRO/controllers.py
```diff
@@ -643,12 +641,15 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
+                        user_pe_ids = bulk_user_pe_id(user_ids) # Current Selection
+                    if len(group_ids) > 0:
```
flavour: As above

modules/templates/SAMBRO/controllers.py
```diff
@@ -662,241 +663,38 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
                                    Counter(priorities_s) == Counter(priorities):
                                     # Update
+                                    comments = json.loads(row.pr_subscription.comments)
                                     success_subscription = update_admin_subscription(\
                                             subscription,
                                             pe_id,
                                             ae_id,
-                                            user_id=int(user_id),
+                                            group_id=comments["pr_group.id"],
                                             filter_id=row.pr_subscription.filter_id,
```
flavour: Slight optimisation:
bring the pr_subscription element of row into scope as used twice
(Won't hold up merge)

modules/templates/SAMBRO/controllers.py
```diff
@@ -662,241 +663,38 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
-                                                ftable.query)
-                        if len(rows) > 0:
-                            for row in rows:
+                        elif user_pe_ids and (pe_id in user_pe_ids.values()):
                                 event_type_id_s, priorities_s = \
                                     get_event_type_priorities(row.pr_filter.query)
                                 if event_type_id_s == event_type_id and \
                                    Counter(priorities_s) == Counter(priorities):
                                     # Update
+                                    user_id = [key for key, value in user_pe_ids.iteritems()
```
flavour: If this is a list I'd prefer the variable name to reflect that:
user_ids

flavour: Ah no, it's a single due to the [0]

modules/templates/SAMBRO/controllers.py
```diff
@@ -905,46 +703,90 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
-                                    get_event_type_priorities(row.pr_filter.query)
-                                if event_type_id_s == event_type_id and \
-                                   Counter(priorities_s) == Counter(priorities):
-                                    # Update
-                                    success_subscription = update_admin_subscription(\
+                    # Groups
+                    if group_pe_ids:
+                        for pe_id in group_pe_ids.keys():
+                            group_id = [value for key, value in group_pe_ids.iteritems()
+                                        if key == pe_id][0]
```
flavour: This seems rather inefficient?
We iterate over group_pe_ids within an iteration of that same var!?
Why not do this for the outer loop?:
for pe_id, group_id in group_pe_ids.iteritems():

modules/templates/SAMBRO/controllers.py
```diff
@@ -905,46 +703,90 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
+                                       Counter(priorities_s) == Counter(priorities):
+                                        # Update
+                                        success_subscription = update_admin_subscription(\
                                             subscription,
                                             pe_id,
                                             ae_id,
-                                            user_id=int(user_id),
+                                            group_id=int(group_id),
                                             filter_id=row.pr_subscription.filter_id,
                                             subscription_id=row.pr_subscription.id,
```
flavour: Again the slight optimisation  of bringing row.pr_subscription into scope as used twice
(& this is within a loop so will matter more)

modules/templates/SAMBRO/controllers.py
```diff
@@ -905,46 +703,90 @@ def _manage_subscriptions(self, resources, filters, recipient_filters, group_fil
...
                                 # Create
                                 success_subscription = update_admin_subscription(\
                                                             subscription,
                                                             pe_id,
                                                             ae_id,
+                                                            group_id=int(group_id),
+                                                            )
+                    # People
+                    if user_pe_ids:
+                        for pe_id in user_pe_ids.values():
```
flavour: I see a common pattern here of:
    if x:
       for row in rows:
            # Lots of lines
    else:
        # Something small

I think I'd find it a lot easier to read with the negative:
    if not x:
        # something small
    else:
          # Add a comment referring back to the if
          for row in rows:
              # Lots of lines

modules/templates/SAMBRO/controllers.py
nursix: Very funny construction - does nothing else than `group_id = group_pe_ids[pe_id]` :D

flavour:
I'll merge to ease Nuwan's review (although really I would much prefer not to be pressured that way: the reviews should be able to happen independently of merges...by having the demo server run off your branch).
Please do go back & do the tweaks though...especially that inefficient loop-in-loop

modules/templates/SAMBRO/controllers.py
nursix: Sorry, what I actually wanted to say is that dict.iteritems() is deprecated, should be just dict.items():

    for pe_id, group_id in group_pe_ids.items():

nursix: It's not wrong to use iteritems, though - e.g. if you want to break out of the loop prematurely like you do here (so you can continue using it if you think you need to). However, this isn't any faster or more readable than:

    for pe_id in group_pe_ids:
        group_id = group_pe_ids[pe_id]

...which still allows you to break out of the loop before the end of the sequence.

nursix: (In fact, in Python 2.7 this variant is the faster than either of items() and iteritems(), so this is what I recommend)

PR #1258, opened 2016-03-15
SAMBRO: make country specific templates

biplovbhandari:
@flavour Please merge


modules/templates/SAMBRO/MV/config.py
```diff
@@ -0,0 +1,17 @@
...
+        Maldives specific template settings for CAP: Common Alerting Protocol
+    """
+
+    T = current.T
+
+    settings.base.system_name = T("Maldives Warning and Situational-Awareness System")
+
+    settings.L10n.languages = {"en-US": "English",
+                               "dv": "ދިވެހި"}
+
```
biplovbhandari: @flavour I added this here because the generic SAMBRO has other two language along this. The settings.L10n.languages["dv"] = "ދިވެހި" would just update the same and hence all four languages are visible for SAMBRO.MV

biplovbhandari:
@flavour Hope this is ok to merge

modules/templates/SAMBRO/MV/config.py
flavour: NB in settings.langauges, US is just 'en' (CAP languages is different)
I don't think that the default SAMBRO template should have any setting for languages any more...not the right place for it in an application template. Instead this should be the job of the Locations template(s) &/or any Deployment Template

modules/templates/SAMBRO/MM/config.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+        Myanmar specific template settings for CAP: Common Alerting Protocol
+    """
+
+    T = current.T
+
+    settings.base.system_name = T("Myanmar Warning and Situational-Awareness System")
+
+    settings.L10n.languages = {"en-US": "English",
+                               "my": "မြန်မာစာ"}
```
flavour: No - pls do as I suggested here

flavour: This is not needed for the Deployment template.
The location template should give exactly the same result if the app template doesn't interfere

modules/templates/SAMBRO/controllers.py
```diff
@@ -1500,6 +1502,29 @@ def __call__(self):
...
+        auth = current.auth
+
+        # Enforce non-interactive response
+        auth.permission.format = "json"
+
+        user = auth.user
+        if not user:
+            auth.permission.fail()
+        else:
+            response = {"o": current.deployment_settings.get_cap_expire_offset(),
```
flavour: See my comment in other commit:
Why have multiple separate controlelrs each to look up 1-2 settings?
far more efficient to do a single lookup for all settings.

modules/templates/SAMBRO/gis_layer_feature.csv
```diff
@@ -1,3 +1,3 @@
-Name,Description,Controller,Function,Marker,GPS Marker,Popup Format,Attributes,Filter,Style,Folder,Visible,Enabled,Style
-All Alerts,,cap,alert,marker_red,,"{headline} (T('Alert'))","cap_info.headline,cap_info.priority",~.is_template=False,,,False,False,"[{'prop':'priority','cat':'Urgent','fill':'FF0000','fillOpacity':0.4},{'prop':'priority','cat':'High','fill':'FFA500','fillOpacity':0.4},{'prop':'priority','cat':'Low','fill':'FFFF00','fillOpacity':0.4},{'prop':'priority','cat':'-','fill':'222222','fillOpacity':0.4}]"
-Public Alerts,,cap,public,marker_red,,"{headline} (T('Alert'))","cap_info.headline,cap_info.priority","~.is_template=False&~.scope=Public",,,False,False,"[{'prop':'priority','cat':'Urgent','fill':'FF0000','fillOpacity':0.4},{'prop':'priority','cat':'High','fill':'FFA500','fillOpacity':0.4},{'prop':'priority','cat':'Low','fill':'FFFF00','fillOpacity':0.4},{'prop':'priority','cat':'-','fill':'222222','fillOpacity':0.4}]"
+Name,Description,Controller,Function,Marker,GPS Marker,Popup Format,Attributes,Filter,Style,Opacity,Folder,Visible,Enabled,Style
```
flavour: You didn't answer why you added Opacity separately at the layer level when it is already in the style?

static/formats/s3csv/cap/info.xsl
```diff
@@ -19,7 +19,7 @@
          Description.................string.............CAP template info description
          Instruction.................string.............CAP template info instruction
          Contact.....................string.............CAP template info contact
-         Parameters..............list:key-value pair....CAP template info parameter
+         Parameters..............key-value pair.........CAP template info parameter
```
flavour: Still no indication of how these should be formatted

modules/templates/SAMBRO/gis_layer_feature.csv
biplovbhandari: @flavour I changed at the layer level but that didn't worked as I expected...so I changed it at the layer level

modules/templates/SAMBRO/MV/config.py
biplovbhandari: @flavour I have ```settings.base.template = ("SAMBRO", "locations.PH", "SAMBRO.PH")``` in models/000_config.py

The SAMBRO instance that we have here at AIT is not country specific ( I plan to use ```settings.base.template = "SAMBRO"``` ,ie all the four languages). In this case, do you think we should remove the language setting in the default SAMBRO template?

modules/templates/SAMBRO/gis_layer_feature.csv
flavour: Read what you wrote: it makes no sense

flavour: Perhaps say what you are tryign to achieve exactly?
The Features already have a transparency, correct? (it's in the style)
You want it to be more transparent?

biplovbhandari: @flavour sorry :( 
I meant it didn't worked as expected when I changed the opacity in the style

modules/templates/SAMBRO/MV/config.py
flavour: I do think that the SAMEBRO template shouldn't have languages in it.
Perhaps what you need is a SAMBRO.Demo template for this?
(or just use 000_config.py)

flavour: Generally the order to apply should be:
("locations.PH", "SAMBRO", "SAMBRO.PH")
Although hopefully the application should never over-ride location settings

modules/templates/SAMBRO/gis_layer_feature.csv
flavour: & what was expected? & what happened?

biplovbhandari: @flavour if you look at here: http://sambro.geoinfo.ait.ac.th/eden/cap/public/64/profile
The layer should be transparent. I changed the style opacity to 0.25 and this is the result. But if I change the layer opacity to 0.25 it works as expected.
See the mail for the login details

flavour: I don't need login details for public ;)
Page took ages to load as in debug mode...but OK, I see the issue:
This is the default style since none of the style elements are being matched (since we have no priority).
The simple option is to remove the fillOpacity from the style elements & just have it at the layer level  (either 0.4 or 0.25 whatever is desired).
The option with more control is to specify an element in the style list with a "fallback" attribute (the value is the title for this option, so 'No priority' or something here)

PR #1259, opened 2016-03-16
SAMBRO: make country specific templates

biplovbhandari:
@flavour Please merge


modules/templates/SAMBRO/tasks.cfg
```diff
@@ -21,7 +21,6 @@
 #  Markers
 gis,marker,gis_marker.csv,marker.xsl
 # Config
-gis,config,default/gis_config.csv,config.xsl
 gis,hierarchy,default/gis_hierarchy.csv,hierarchy.xsl
```
flavour: This should probably be removed from here too?
Perhaps the Demo SAMBRO template should have a gis_config to define the SITE_DEFAULT
On the Demo server you could set
settings.template += (Locations/MM, Locations/MV, Locations/PH, SAMBRO, SAMBRO/Demo)

PR #1260, opened 2016-03-16
SAMBRO: filter pvt alert and send to pvt group

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/config.py
+++ b/modules/templates/SAMBRO/config.py
@@ -233,6 +233,7 @@ def config(settings):
     # -------------------------------------------------------------------------
     def customise_cap_alert_resource(r, tablename):
 
+        db = current.db
         s3db = current.s3db
         def onapprove(record):
             # Normal onapprove
@@ -249,7 +250,6 @@ def config(settings):
                     current.log.debug("tweepy module needed for sending tweets")
                 else:
                     T = current.T
-                    db = current.db
                     alert_id = int(record["id"])
                     atable = s3db.cap_alert
                     itable = s3db.cap_info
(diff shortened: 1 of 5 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -280,6 +280,65 @@ def onapprove(record):
                         except tweepy.error.TweepError, e:
                             current.log.debug("Sending tweets failed: %s" % e)
 
+            # Send out private alerts to addresses
+            # @ToDo: Check for LEFT join when required
+            # this is ok for now since every Alert should have an Info & an Area
+            # @ToDo: Handle multi-lingual alerts when required
```
flavour: ok, will merge even with this major issue unresolved! ;)
We should probably have a list of all the places which need handling for this.
I thought we already had this usecase in our 1st deployment countries or is that not the case?

modules/templates/SAMBRO/config.py
```diff
@@ -280,6 +280,65 @@ def onapprove(record):
                         except tweepy.error.TweepError, e:
                             current.log.debug("Sending tweets failed: %s" % e)
 
+            # Send out private alerts to addresses
+            # @ToDo: Check for LEFT join when required
+            # this is ok for now since every Alert should have an Info & an Area
```
flavour: I don't think this needs a ToDo actually...was more for your info

modules/templates/SAMBRO/config.py
```diff
@@ -746,6 +810,29 @@ def get_html_email_content(row):
...
+        event_type_id = row["cap_info.event_type_id"]
+        priority_id = row["cap_info.priority"]
+
+        if not isinstance(event_type_id, lazyT):
+            event_type = itable.event_type_id.represent(event_type_id)
+        else:
+            event_type = event_type_id
+
+        if priority_id and \
+           priority_id != "-":
```
flavour: "-" comes from current.messages["NONE"] right?
better to have that in this check so that if that value changes then this code won't need updating

PR #1263, opened 2016-03-23
CAP: create location from geocode SAME + separate function for area and resource component

biplovbhandari:
@flavour Please MERGE..

### DIFF
--- a/modules/s3cfg.py
+++ b/modules/s3cfg.py
@@ -2438,6 +2438,14 @@ class S3Config(Storage):
 
         return self.cap.get("post_to_twitter", False)
 
+    def get_cap_same_code(self):
+        """
+            Name of the tag that will be used to lookup in the gis_location_tag
+            to extract location_id for the alert
+        """
+
+        return self.cap.get("same_code")
+
     # -------------------------------------------------------------------------
     # CMS: Content Management System
     #
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -1133,7 +1133,7 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
         configure(tablename,
                   crud_form = crud_form,
                   list_fields = list_fields,
-                  onaccept = update_alert_id(tablename),
+                  onaccept = self.cap_resource_onaccept,
                   onvalidation = self.cap_resource_onvalidation,
                   super_entity = "doc_entity",
                   )
@@ -1303,7 +1303,7 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
                   #create_next = URL(f="area", args=["[id]", "location"]),
                   crud_form = crud_form,
                   deduplicate = self.cap_area_duplicate,
-                  onaccept = update_alert_id(tablename),
+                  onaccept = self.cap_area_onaccept,
                   onvalidation = self.cap_area_onvalidation,
                   )
 
@@ -1368,7 +1368,7 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
 
         configure(tablename,
                   deduplicate = S3Duplicate(primary=("area_id", "location_id")),
-                  onaccept = update_alert_id(tablename),
+                  onaccept = self.cap_area_location_onaccept
                   )
 
         # ---------------------------------------------------------------------
@@ -1412,8 +1412,8 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
                      *s3_meta_fields())
 
         configure(tablename,
-                  onaccept = update_alert_id(tablename),
-        #         deduplicate = self.cap_area_tag_deduplicate,
+                  onaccept = self.cap_area_tag_onaccept,
+                  deduplicate = S3Duplicate(primary=("area_id", "tag", "value")),
                   )
 
         # ---------------------------------------------------------------------
@@ -1691,18 +1691,6 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
 
     # -------------------------------------------------------------------------
     @staticmethod
-    def cap_area_onvalidation(form):
-        """
-            Custom Form Validation
-        """
-
-        form_vars = form.vars
-        if form_vars.get("ceiling") and not form_vars.get("altitude"):
-            form.errors["altitude"] = \
-            current.T("'Altitude' field is mandatory if using 'Ceiling' field.")
-
-    # -------------------------------------------------------------------------
-    @staticmethod
     def cap_area_duplicate(item):
 
         data = item.data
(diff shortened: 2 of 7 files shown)


modules/s3db/cap.py
```diff
@@ -1737,6 +1725,129 @@ def cap_area_duplicate(item):
...
+                if trow:
+                    # Match
+                    ltable = db.cap_area_location
+                    ldata = {"area_id": area_id,
+                             "alert_id": alert_id or None,
+                             "location_id": trow.location_id,
+                             }
+                    lid = ltable.insert(**ldata)
+                    current.auth.s3_set_record_owner(ltable, lid)
+                    s3db.onaccept(ltable, dict(id=lid))
```
nursix: This doesn't look right - the onaccept expects an area_id not id. Maybe pass the entire ldata:

    ldata["id"] = lid
    s3db.onaccept(ltable, ldata)

?

modules/s3db/cap.py
```diff
@@ -1737,6 +1725,129 @@ def cap_area_duplicate(item):
...
+            alert_id = item.alert_id
+
+            if alert_id:
+                db(db.cap_area.id == form_vars.id).update(alert_id = alert_id)
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def cap_area_onvalidation(form):
+        """
+            Custom Form Validation
```
nursix: It's not custom if it is in the standard model. If it is meant to be custom, move it into the template. Otherwise remove the "Custom" here.

modules/s3db/cap.py
flavour: +1

modules/s3db/cap.py
```diff
@@ -1737,6 +1725,129 @@ def cap_area_duplicate(item):
...
+        """
+
+        form_vars = form.vars
+        if form_vars.get("ceiling") and not form_vars.get("altitude"):
+            form.errors["altitude"] = \
+            current.T("'Altitude' field is mandatory if using 'Ceiling' field.")
```
nursix: PEP8: increase indent if it's a continued line

modules/s3db/cap.py
```diff
@@ -1737,6 +1725,129 @@ def cap_area_duplicate(item):
...
+        alert_id = row.alert_id
+        if alert_id:
+            # This is not template area
+            # NB Template area are not linked with alert_id
+            db(db.cap_area_location.id == form_vars.id).update(alert_id = alert_id)
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def cap_area_tag_onaccept(form):
+
```
nursix: docstring missing

modules/s3db/cap.py
```diff
@@ -1737,6 +1725,129 @@ def cap_area_duplicate(item):
...
+        form_vars = form.vars
+        if form_vars.get("ceiling") and not form_vars.get("altitude"):
+            form.errors["altitude"] = \
+            current.T("'Altitude' field is mandatory if using 'Ceiling' field.")
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def cap_area_location_onaccept(form):
+
```
nursix: docstring missing

modules/s3db/cap.py
```diff
@@ -1737,6 +1725,129 @@ def cap_area_duplicate(item):
 
     # -------------------------------------------------------------------------
     @staticmethod
+    def cap_area_onaccept(form):
+
```
nursix: docstring missing

modules/s3db/cap.py
flavour: Although actually I would comment this anyway as it wouldn't be needed in this case
(can keep commented in case that changes)

modules/s3cfg.py
```diff
@@ -2438,6 +2438,14 @@ def get_cap_post_to_twitter(self):
 
         return self.cap.get("post_to_twitter", False)
 
+    def get_cap_same_code(self):
+        """
+            Name of the code that will be used to lookup in the gis_location_tag
```
nursix: Actually: name of the tag - right? Easier to understand.

modules/s3db/cap.py
```diff
@@ -1737,6 +1725,129 @@ def cap_area_duplicate(item):
...
+                    s3db.onaccept(ltable, dict(id=lid))
+
+        if alert_id:
+            # This is not template area
+            db(db.cap_area_tag.id == form_vars.id).update(alert_id = alert_id)
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def cap_resource_onaccept(form):
+
```
nursix: docstring

modules/s3db/cap.py
flavour: Personally I don't see that every onaccept needs a docstring - we know what onaccepts are so something like "Post DB processing for x table" is pretty tautological.
I prefer the "what it does" to be within the code itself as it can often do several things & even if just 1 it makes more sense to me to have within the code as easier to read.
What is the argument to have a docstring here?

modules/s3cfg.py
flavour: Well, we call this 'Key' normally...just that that's a reserved word in SQL so we need to use the fieldname 'tag' & hence the table gets that too.
Some UI usecases expose 'Tag' to the user too, although Key is more' common'
Here I think 'Name of the code' is fine

nursix: I was thinking "Name of the tag" as opposed to "Value of the Tag". Name of the code is a bit cryptic.

However, I just commented for reason of completeness - my actual point was about the onaccept.

modules/s3db/cap.py
nursix: http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/CodeConventions#CommentsandDocstrings

Of course, something like "post-DB processing" is just nonsense ;) however, I prefer having a bullet point list with the actions done onaccept rather than having to study the function (comments in the function may explain "how" but a summary of what would be nice in the docstring - a nice practice introduced by @flavour).

This function, in any case, has neither the one nor the other.

nursix: Good examples:
https://github.com/sahana/eden/blob/master/modules/s3db/hrm.py#L1026
https://github.com/sahana/eden/blob/master/modules/s3db/gis.py#L2210
https://github.com/sahana/eden/blob/master/modules/s3db/gis.py#L2742
...and many more.

modules/s3cfg.py
nursix: Unfortunately, "name of the code" doesn't quite give a hint that the "code" is stored as location tag :/ I can see it now within the context of this PR - but I (or another developer studying SAMBRO) may not be able to remember that in a year or two. At least there should be a hint that this has to do with gis_location_tags. Just a hint. Please?

biplovbhandari:
@flavour Please merge

modules/s3db/cap.py
```diff
@@ -1737,6 +1725,140 @@ def cap_area_duplicate(item):
...
+                    # Match
+                    ltable = db.cap_area_location
+                    ldata = {"area_id": area_id,
+                             "alert_id": alert_id or None,
+                             "location_id": trow.location_id,
+                             }
+                    lid = ltable.insert(**ldata)
+                    ldata["id"] = lid
+                    current.auth.s3_set_record_owner(ltable, lid)
+                    s3db.onaccept(ltable, ldata)
```
flavour: I would comment this line as does nothing here (but commented is a reminder to uncomment if the fn changes) Can be done post-merge

modules/s3db/cap.py
```diff
@@ -2719,16 +2758,16 @@ def apply_method(r, **attr):
...
+                        #TR(TD(B("%s: " % T("User"))),
+                        #   TD(INPUT(_type="text", _name="user",
+                        #            _id="user", _value="")),
+                        #   TD(),
+                        #   ),
+                        #TR(TD(B("%s: " % T("Password"))),
+                        #   TD(INPUT(_type="text", _name="password",
+                        #            _id="password", _value="")),
+                        #   TD(),
+                        #   ),
```
flavour: hmm, why remove this?
This is a useful feature as some feeds require a username/password.
If you wish to simplify the UI for a single usecase which doesn't require it then this should be done using a deployment_setting

PR #1265, opened 2016-03-26
SAMBRO: MM Alert Editors can create template

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/MM/menus.py
+++ b/modules/templates/SAMBRO/MM/menus.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+
+from gluon import *
+from s3 import *
+from s3layouts import *
+try:
+    from .layouts import *
+except ImportError:
+    pass
+import s3menus as default
+
+# =============================================================================
+class S3OptionsMenu(default.S3OptionsMenu):
+    """
+        Custom Controller Menus
+
+        The options menu (left-hand options menu) is individual for each
+        controller, so each controller has its own options menu function
+        in this class.
+
+        Each of these option menu functions can be customised separately,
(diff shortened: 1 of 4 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -31,7 +31,7 @@ def config(settings):
     settings.base.system_name_short = T("SAMBRO")
 
     # Pre-Populate
-    settings.base.prepopulate += ("SAMBRO", "SAMBRO/Demo", "default/users")
+    settings.base.prepopulate += ("SAMBRO", "SAMBRO/MM", "SAMBRO/Demo", "default/users")
```
flavour: probably not wanted for the core template!?

modules/templates/SAMBRO/MM/menus.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @staticmethod
+    def cap():
+        """ CAP menu """
+
+        s3_has_role = current.auth.s3_has_role
+        cap_editors = lambda i: s3_has_role("ALERT_EDITOR") or \
+                                s3_has_role("ALERT_APPROVER")
+
+        return M(c="cap")(
```
flavour: Whilst it is possible to do a full override like this, I wonder if this would add too much to maintenance?
For adding a single entry it is more DRY to do something like:
menu = super(S3OptionsMenu, self).cap()
menu[1] = M()

It may need some playing with...e.g. here super is the core s3menus version, not any SAMBRO version that exists, so that may need tweaking.

I'm not too objecting to a completely different menu here...just worth trying to be DRY if we're just adding 1 entry

PR #1266, opened 2016-03-29
SAMBRO: subscription fixes + SAMBRO: MM editors can create template + CAP:tweaks

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -609,7 +609,7 @@ def alert():
                     s3db.configure("cap_info",
                                    deletable = False,
                                    editable = False,
-                                   insertable=False,
+                                   insertable = False,
                                    )
                 if settings.get_cap_restrict_fields():
                     if r.record.msg_type in ("Update", "Cancel", "Error"):
@@ -734,6 +734,9 @@ def alert():
                     row_clone["alert_id"] = lastid
                     row_clone["template_info_id"] = row.id
                     row_clone["is_template"] = False
+                    row_clone["effective"] = request.utcnow
+                    row_clone["expires"] = s3db.cap_expiry_date()
+                    row_clone["sender_name"] = s3db.cap_sender_name()
                     itable.insert(**row_clone)
 
                 # Clone all cap_resource entries from the alert template
--- a/modules/s3/s3notify.py
+++ b/modules/s3/s3notify.py
@@ -284,7 +284,8 @@ class S3Notifications(object):
         # Authorization (subscriber must be logged in)
         auth = current.auth
         pe_id = subscription["pe_id"]
-        if not auth.s3_logged_in() or auth.user.pe_id != pe_id:
+
+        if not pe_id:
             r.unauthorised()
 
         # Fields to extract
@@ -419,6 +420,9 @@ class S3Notifications(object):
                 exc_info = sys.exc_info()[:2]
                 error = ("%s: %s" % (exc_info[0].__name__, exc_info[1]))
                 errors.append(error)
+                continue
+
+            if not message:
                 continue
 
             # Send the message
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -74,6 +74,8 @@ class S3CAPModel(S3Model):
              "cap_area_location",
              "cap_area_tag",
              "cap_info_category_opts",
+             "cap_expiry_date",
+             "cap_sender_name",
              "cap_template_represent",
              )
 
@@ -901,13 +903,14 @@ class S3CAPModel(S3Model):
                      s3_datetime("expires",
                                  label = T("Expires at"),
                                  past = 0,
-                                 default = self.get_expirydate,
+                                 default = self.cap_expirydate,
                                  comment = DIV(_class="tooltip",
                                                _title="%s|%s" % (T("The expiry time of the information of the alert message"),
                                                                  T("If this item is not provided, each recipient is free to enforce its own policy as to when the message is no longer in effect."))),
                                  ),
                      Field("sender_name",
                            label = T("Sender's name"),
+                           default = self.cap_sendername,
                            comment = DIV(_class="tooltip",
                                          _title="%s|%s" % (T("The text naming the originator of the alert message"),
                                                            T("The human-readable name of the agency or authority issuing this alert."))),
@@ -1425,6 +1428,8 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
                     cap_area_represent = area_represent,
                     cap_info_represent = info_represent,
                     cap_info_category_opts = cap_info_category_opts,
+                    cap_expiry_date = self.cap_expirydate,
+                    cap_sender_name = self.cap_sendername,
                     cap_template_represent = self.cap_template_represent,
                     )
 
@@ -1480,7 +1485,7 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
 
     # -------------------------------------------------------------------------
     @staticmethod
-    def get_expirydate():
+    def cap_expirydate():
         """
             Default Expiry date based on the expire offset
         """
@@ -1488,6 +1493,27 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
         return current.request.utcnow + \
                datetime.timedelta(days = current.deployment_settings.\
                                   get_cap_expire_offset())
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def cap_sendername():
+        """
+            Default Sender name for the alert
+            Sendername is the name of the organisation if user is associated
+            else None
+        """
+
+        db = current.db
+        utable = db.auth_user
+        otable = current.s3db.org_organisation
+        query = (utable.id == current.auth.user.id) & \
+                (utable.organisation_id == otable.id) & \
+                (otable.deleted != True)
+        row = db(query).select(otable.name,
+                               limitby=(0, 1)).first()
+        if row:
+            return row.name
+        return None
 
     # -------------------------------------------------------------------------
     @staticmethod
@@ -1670,6 +1696,8 @@ T("Upload an image file(bmp, gif, jpeg or png), max. 800x800 pixels!"))),
         alert_id = record["id"]
 
         # Update approved_on at the time the alert is approved
+        # @ToDo: update approved_on when approval is not required
+        # i.e. we allow editors to be self approver
         if alert_id:
             db = current.db
             approved_on = record["approved_on"]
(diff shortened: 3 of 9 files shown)


controllers/cap.py
```diff
@@ -734,6 +734,9 @@ def postp(r, output):
                     row_clone["alert_id"] = lastid
                     row_clone["template_info_id"] = row.id
                     row_clone["is_template"] = False
+                    row_clone["effective"] = request.utcnow
+                    row_clone["expires"] = s3db.cap_expiry_date,
```
flavour: trailing comma is probably not desired (do you want a tuple here?)

modules/s3db/cap.py
```diff
@@ -1491,6 +1496,27 @@ def get_expirydate():
 
     # -------------------------------------------------------------------------
     @staticmethod
+    def get_sendername():
```
flavour: Model methods should have a prefix to avoid clashes

controllers/cap.py
```diff
@@ -734,6 +734,9 @@ def postp(r, output):
...
+                    row_clone["sender_name"] = s3db.cap_sender_name
```
flavour: Should not this & the above one be run?
i.e. don't you want:
row_clone["expires"] = s3db.cap_expiry_date()
row_clone["sender_name"] = s3db.cap_sender_name()

modules/s3/s3notify.py
```diff
@@ -284,7 +284,12 @@ def send(cls, r, resource):
...
         auth = current.auth
         pe_id = subscription["pe_id"]
-        if not auth.s3_logged_in() or auth.user.pe_id != pe_id:
+        ptable = current.s3db.pr_pentity
+        query_ = (ptable.pe_id == pe_id) & (ptable.deleted != True)
+        row_ = current.db(query_).select(ptable.instance_type,
+                                         limitby=(0, 1)).first()
+        if row_.instance_type == "pr_person" and \
+           (not auth.s3_logged_in() or auth.user.pe_id != pe_id):
             r.unauthorised()
```
flavour: I don't think this is correct.
Like this we can bypass security just by providing a pe_id which isn't related to a person.
This needs more thought

modules/s3/s3notify.py
```diff
@@ -412,39 +417,42 @@ def get_template(path, filenames):
...
-            except:
-                exc_info = sys.exc_info()[:2]
-                error = ("%s: %s" % (exc_info[0].__name__, exc_info[1]))
-                sent = False
-
-            if sent:
-                # Successful if at least one notification went out
-                success = True
+            if output["new"] is None and output["upd"] is None:
+                message = "No new notifications"
```
flavour: Do we really want to send such a message if there are no new notifications?
Personally I'd prefer to receive nothing.
I can see that potentially some people might disagree & prefer to receive such a message daily whether or not there are any new notifications...so perhaps add a deployment_setting to allow this option & then T() the message that it sends.
Ideally the normal case, where we don't want to send empty notifications, would be caught earlier

nursix: No! Templates want to customize that as well. And it is not internationalized anyway, so plain and simply: no.

flavour: > caught earlier

i.e. Why submit a POST request needlessly?
This adds a whole request overhead cycle to the system...

modules/templates/SAMBRO/auth_roles.csv
```diff
@@ -1,8 +1,7 @@
 uid,role,description,controller,function,table,uacl,oacl,hidden,system,protected,Notes
-ANONYMOUS,Anonymous,,cap,public,,READ,,,,,Can only see scope=public alerts
+ANONYMOUS,Anonymous,,cap,alert,,READ,,,,,for subscriptions purpose
```
flavour: hmm, this seems to break the main security model!

modules/templates/SAMBRO/config.py
```diff
@@ -355,6 +356,14 @@ def customise_cap_alert_controller(**attr):
         # Custom prep
         standard_prep = s3.prep
         def custom_prep(r):
+            auth = current.auth
+            if not auth.user:
+                # For notifications for group
+                if r.representation == "msg":
+                    s3.filter = (FS("scope") != "Private")
+                else:
+                    auth.permission.fail()
```
flavour: ok, I see you put in a custom check here
But the cap/public also needs access - since this is used for all other representations.
Can the 'msg' representation not be set to use the public controller for anonymous requests?

modules/templates/SAMBRO/config.py
```diff
@@ -536,41 +545,35 @@ def custom_msg_render(resource, data, meta_data, format=None):
...
-                    else:
-                        if s3_utc(created_on) >= last_check_time:
-                            append_record = new.append
-
-                record = []
-                append_column = record.append
-                for colname, label in labels:
-                    append_column((label, row[colname]))
-                append_record(record)
+                row_ = db(atable.id == row["cap_alert.id"]).select(atable.approved_on,
```
flavour: Just checking:
approved_on gets set in both workflow options, right?
If 'approval is not needed' then the Editor auto-approves under the hood?

modules/s3/s3notify.py
biplovbhandari: This was added because for HTML emails empty emails ( no body only subject) were received

modules/s3/s3notify.py
nursix: Actually I think this is correct, though it can be simplified by removing the logged-in check completely.

If the lookup request comes in anonymously, then we apply ANONYMOUS permissions in the data lookup (resource.select below) - which allows us to subscribe other PE's than persons. 

Hence, the subscriber does not have to be logged-in - the only thing that is important at this point is that we do have a valid subscriber (pe_id!=None).

modules/s3/s3notify.py
nursix: @biplovbhandari: Doesn't make sense - send() exits with a JSON message when no rows were found, and doesn't send a message. 
https://github.com/sahana/eden/blob/master/modules/s3/s3notify.py#L302

nursix: @flavour: To see whether there are updates which are accessible by the subscriber (and hence to be notified), you need to impersonate the subscriber and go through the resource's controller (including custom prep/postp and other customizations of the template), which is what the POST request is for.

This can't be done during notify() - hence the POST request is always required.

nursix: ...but it's just a local request, so overheads are small - and notify isn't time-critical anyway.

nursix: @biplovbhandari : What you may want to do is to add:

    if not message:
        continue

before this block:

https://github.com/sahana/eden/blob/master/modules/s3/s3notify.py#L424

That way, if the message template returns empty, no message would be sent.

modules/templates/SAMBRO/config.py
biplovbhandari: @flavour both workflow options means? approved_on is set here: https://github.com/sahana/eden/blob/master/modules/s3db/cap.py#L1678

Not for imported but for system alert. Anyway we have to either Relay Alert for the imported alert.

modules/s3/s3notify.py
biplovbhandari: @nursix But for html emails, although the body is empty, the message is not empty as it has at least <html> and <body> tags, no?

nursix: Err - that is entirely up to your message template.

If that template adds tags even if there is no contents, then yes - but there is nothing that says the template must add tags. It could render empty if there is nothing to render.

At the same time, there is nothing that says the template must not add contents even if there are no updates. It could render an info like the one you want to hardcode here.

Entirely your choice - which to leave is exactly why I'm insisting on this to be passed to the template and not hardcoded here.

nursix: Here is where these tags get added:
https://github.com/sahana/eden/blob/master/modules/templates/SAMBRO/views/msg/notify_email_html.html

...so you can easily prevent that.

nursix: Btw, there is yet another problem here that forbids this change: the data extraction method is completely customizable (settings.get_msg_notify_renderer), and if it doesn't return a dict with "upd" and "new", then this check will just crash.

But the whole point about having this customizable is that you can implement any kind of custom rendering, there is nothing that says it must have a "upd" or "new" - it could also render complete HTML elements or text sections, and hence a different dict structure.

So, this change is simply a no-go - the only acceptable alternative is to check whether the message template renders empty and in such cases not sending the message.

nursix: ...and I did recommend before to move your complex rendering from the message template into a renderer function (so that it can be debugged - which is difficult with view templates), and have the template just be `{{=contents}}`.

That way, you can easily render empty by returning `contents=""` from the renderer ;)

nursix: And here you did:
https://github.com/sahana/eden/blob/master/modules/templates/SAMBRO/config.py#L528

Now, your renderer returns a dict structure with "upd" and "new" - but it doesn't have to. The renderer can be tailored to the template and vice versa - the dict structure is only between these two.

And to leave it at that, we can not and we must not inspect the renderer output in send() - but rather check whether the template produces any message at all from that (which seems legitimate as we don't really want to send completely empty messages).

biplovbhandari:
@flavour Please merge

modules/s3/s3notify.py
nursix: You haven't corrected this ;)

flavour: This hasn't yet been modified.
We don't need the DB call to lookup the instance type.
We just need pe_id != None

modules/templates/SAMBRO/auth_roles.csv
```diff
@@ -1,8 +1,8 @@
 uid,role,description,controller,function,table,uacl,oacl,hidden,system,protected,Notes
 ANONYMOUS,Anonymous,,cap,public,,READ,,,,,Can only see scope=public alerts
+ANONYMOUS,Anonymous,,cap,alert,,READ,,,,,for subscriptions purpose
```
flavour: Is there really no way to avoid this?

modules/templates/SAMBRO/config.py
flavour: Both workflow options are:
* Approval Required (i.e. 2 steps with 2 roles: Editor & Approver)
* Approval not Required (i.e. 1 step with 1 role: Editor)
or did this change? (so hard to remember all details)
Do we now always have the 2 steps?
Just that some deployments allow the Editors to be self-approvers (i.e. they are given Approver role) & some deployments have this be a separate role

modules/templates/SAMBRO/views/msg/notify_email_html.html
```diff
@@ -1,12 +1,7 @@
...
-{{pass}}
-</body>
-</html>
+{{if new is not None:}}
+<html>
+{{new_title=T("The following Alert has been added") if new == 1 else T("The following %(number)s Alerts have been added") % dict(number=new)}}
+{{=XML(new_title)}}
+{{=XML(new_records)}}
+</html>
+{{pass}}
```
flavour: Remember that lines with just python {{}} still give a blank line in the output.
More optimised output is obtained using:
{{if new is not None:}}<html>
</html>{{pass}}

modules/templates/SAMBRO/config.py
biplovbhandari: @flavour currently its 2 steps...I will leave comment to remind when we have option 2 also

PR #1267, opened 2016-04-01
SAMBRO: subscription fixes + SAMBRO: MM editors can create template  + CAP:tweaks

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/controllers.py
+++ b/modules/templates/SAMBRO/controllers.py
@@ -248,41 +248,44 @@ class subscriptions(S3CustomController):
                    #                 options = self._options("location_id"),
                    #                 _name = "location-filter",
                    #                 ),
-                   S3OptionsFilter("language",
-                                   label = T("Language"),
-                                   options = current.deployment_settings.get_cap_languages(),
-                                   represent = "%(name)s",
-                                   resource = "cap_info",
-                                   _name = "language-filter",
-                                   ),
                    ]
+        cap_languages = current.deployment_settings.get_cap_languages()
+        if len(cap_languages) > 1:
+            language_filters = S3OptionsFilter("language",
+                                               label = T("Language"),
+                                               options = cap_languages,
+                                               represent = "%(name)s",
+                                               resource = "cap_info",
+                                               _name = "language-filter",
+                                               )
+            filters.append(language_filters)
 
         if current.request.get_vars["option"] == "manage_recipient" and \
-           (has_role("ALERT_EDITOR") or has_role("ALERT_APPROVER")):
-                from s3 import S3Represent
-                recipient_filters = [S3OptionsFilter("id",
-                                           label = T("People"),
-                                           represent = S3Represent(lookup="auth_user",
-                                                fields = ["first_name", "last_name"],
-                                                field_sep = " ",
-                                                ),
-                                           widget = "multiselect",
-                                           resource = "auth_user",
-                                           _name = "person-filter",
-                                           ),
-                                     ]
-                group_filters = [S3OptionsFilter("id",
-                                           label = T("Groups"),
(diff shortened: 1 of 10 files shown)


modules/templates/SAMBRO/controllers.py
```diff
@@ -339,6 +339,11 @@ def _options(fieldname):
             # IDs converted inside widget's _options() function
             rows = current.db(query).select(ltable.id)
             options = [row.id for row in rows]
+        elif fieldname == "language":
+            options = current.deployment_settings.get_L10n_languages()
```
flavour: I'm not sure why you changed this? What ou had was correct: the dropdown should be the list of cap languages not UI languages.
The suggestions I had were:
* Hide filter completely if only 1x CAP language
* Default the filter to the user's language for user-controlled subscriptions if there is a match (here was wehere en<>en_US needed tackling)

modules/templates/SAMBRO/controllers.py
```diff
@@ -248,44 +248,65 @@ def __call__(self):
...
+                default_option_language = None
+                default_user_language = auth.user.language
+                if default_user_language:
+                    if default_user_language == "en":
+                        default_user_language = "en-US"
+                    if default_user_language in cap_languages:
+                        default_option_language = default_user_language
+                language_filters = S3OptionsFilter("language",
+                                                   label = T("Language"),
+                                                   default = default_option_language,
```
biplovbhandari: @flavour How do you default value in S3OptionsFilter? This does not seem to work

nursix: That is exactly how to do it - specify a "default" value. Are you sure this is executed, and the default is in the selectable options?

nursix: And what do you mean by "does not seem to work"? How are you testing it?

nursix: Ah ok - I see. This "filter form" doesn't actually filter anything - and if it does not filter, filter defaults can not be applied. So setting a filter default here is pointless.

For what you seem trying to do, you must send the default values in the get_vars when calling S3FilterForm.fields in manage_subscriptions (that is, when it isn't an update of an existing subscription).

nursix: That is here:
https://github.com/biplovbhandari/eden/blob/master/modules/templates/SAMBRO/controllers.py#L445

The filter widgets get their current values from the get_vars.

If you want to set a default selection for those widgets, you must pass those defaults through the get_vars (unless it's an update). 

That seems relatively simple to do because you are generating the get_vars here:
https://github.com/biplovbhandari/eden/blob/master/modules/templates/SAMBRO/controllers.py#L949
https://github.com/biplovbhandari/eden/blob/master/modules/templates/SAMBRO/controllers.py#L1008

Simply don't leave them empty if there is no subscription row, but set your defaults.

nursix: Please note that a filter default is not the same as pre-selecting an option in a CRUD form widget - because a filter default does actually mean to /filter/ the target resource by default. That does obviously only work if the filter widget is actually attached to a resource /and/ that resource is the target of the current request - otherwise it would claim to be filtering while it can not.

Since you're re-purposing the filter widgets into CRUD widgets here, the filter default option is indeed not the way to pre-select values. In this case you must treat the filter widgets like CRUD widgets, and hence pass in the default value as the current value when rendering the widget.

In normal CRUD widgets, this happens via the "value" parameter - in filter widgets, this is done through the "get_vars" parameter (which is passed on from the filter form to individual widgets).

biplovbhandari:
@flavour Please merge...

flavour:
Looks good now - nice enhancements :)

PR #1268, opened 2016-04-04
added internationalized labels

avni510:


biplovbhandari:
@avni510 did you forked/cloned the latest eden repository? This branch has conflicts with the master branch of eden. Besides these are not internationalisation.
See the wiki for git at http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/Git and for internationalisation at http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/Internationalisation

modules/s3db/disease.py
```diff
@@ -92,12 +92,12 @@ def model(self):
         tablename = "disease_disease"
         define_table(tablename,
                      self.super_link("doc_id", "doc_entity"),
-                     Field("name",
+                     Field("name", "text"
```
nursix: This modifies the field type rather than the label.

An internationalized label looks like this:

    Field("name",
          label = T("Name"), # <= internationalized label, a T-instance
         ),

The field type "text" is not correct here, btw - name if of type "string", which is the default type and therefore not specified explicitly.

Same for the other fields.

nursix:
If you look at https://github.com/avni510/eden/tree/disease_mod_translatable_labels, you can see that it is 125 commits behind sahana/eden. That means you need to rebase it on the latest sahana/eden before it can be merged.

It is recommendable that you always rebase before you push to GitHub - making that a habit prevents accumulation of merge conflicts, keeps you up-to-date with the latest development and saves us the extra round of comms to point you to that ;)

nursix:
Generally speaking, you always need to test whether your changes have the desired effect before you send a pull request.

In this case, the desired effect is that by changing the GUI language (using the language selector), the labels for these fields should appear translated in the corresponding forms.

Getting you to run the GUI in order to verify the effect of your changes was the actual idea behind this exercise ;) - the required changes are trivial, but they are not the point here.

avni510:
Ok thank you for the feedback. My apologies! And thank you for your patience. After editing the code based on suggestions ex: Field("description", "text", label = T("Description"))
and running it locally, I found that some of the text was not translated. Ex: description and short name but some fields were translated ex: Name and Acronym. 

Is this something that I can fix?

flavour:
This depends on which language you selected & whether there is a complete translation available.
If you can speak the language then you can add missing translations to the appropriate file in the languages folder.

PR #1270, opened 2016-04-04
CAP: add event type to alert segment and filter template based on event type + hide event type in info->dont remove because it is needed for priority

biplovbhandari:
@flavour Please merge


static/scripts/S3/S3.js
```diff
@@ -1094,7 +1094,7 @@ S3.openPopup = function(url, center) {
             // Append lookup key to the URL
             var q;
             if (value) {
-                q = lookupResource + '.' + lookupKey + '=' + value;
+                q = '~.' + lookupKey + '=' + value;
```
nursix: No, this is not correct. Must be lookupResource here, can't rely on ~

nursix: What is the problem that made you change this?

flavour: lookupResoruce is used for the controller, which isn't always the resource
e.g. cap/template (which is a filtered version of cap/alert)
What cases are lookupResource != ~ ?

nursix: The lookupResource is != ~ when the lookupPrefix specifies a parent (e.g. "pr/person/1/address").

I do remember that this was deliberately changed from "~" into "lookupResource" due to a bug report, so I don't think it can't just be changed back.

flavour: ok, so then we do need the lookupController option adding
(it can default to lookupResource if not specified)

nursix: Ahm, actually I don't think so. The change with lookupResource predates the introduction of the ~ shortcut, so it can't have been related to that.

Try using ~ - this should cover all current cases, and if a problem comes up we can look over this again.

nursix: We can still use lookupURL if we really wanted to add a parent (this is the existing option to override the URL construction with a custom URL, and I don't think we need another one)

flavour: So you're saying that this change can stay as-is, right?

controllers/cap.py
```diff
@@ -602,6 +602,9 @@ def custom_widget_fn_3(r, **attr):
                 if alert_id:
                     itable.web.default = settings.get_base_public_url()+\
                                          URL(c="cap", f="alert", args=alert_id)
+                    row = db(r.table.id == alert_id).select(r.table.event_type_id,
```
flavour: Avoid duplicate r.table: bring into scope

controllers/cap.py
```diff
@@ -943,6 +946,9 @@ def prep(r):
             if alert_id:
                 itable.web.default = settings.get_base_public_url()+\
                                      URL(c="cap", f="alert", args=alert_id)
+                row = db(r.table.id == alert_id).select(r.table.event_type_id,
```
flavour: Avoid duplicate r.table: bring into scope

biplovbhandari:
@flavour Please merge

PR #1271, opened 2016-04-05
added internationalized labels

avni510:
Added internationalized labels to the creating a new disease table in the disease module


flavour:
Basically correct now, however if you look at how the formatting is done in other models, you could adjust to match :)

PR #1272, opened 2016-04-06
CAP: add event type to alert segment and filter template based on event type + hide event type in info->dont remove because it is needed for priority  SAMBRO: bugfix

biplovbhandari:


modules/s3db/cap.py
```diff
@@ -286,6 +286,22 @@ def model(self):
...
+                                                                        T("Event Type is classification of event."))),
+                                        script = '''
+$.filterOptionsS3({
+    'trigger':'event_type_id',
+    'target':'template_id',
+    'lookupPrefix': 'cap',
+    'lookupResource':'template',
+    'fncRepresent': function(record,PrepResult){return record.template_title},
+    'optional': true,
+    'lookupURL': S3.Ap.concat('/cap/template.json?~.event_type_id=')
```
flavour: slightly less HTML to send back to browser if you substitute the app name server-side instead of client-side (can be cleaned up post merge)

PR #1273, opened 2016-04-09
SAMBRO: Google cloud messaging push notifications for mobile client

biplovbhandari:


modules/s3/s3msg.py
```diff
@@ -804,6 +805,63 @@ def dispatch_to_pe_id(pe_id,
...
+            Push the message relating to google cloud messaging server
+
+            @param title: The title for notification
+            @param message: The message to be sent to GCM server
+            @param api_key: The API key for GCM server
+            @param registration_ids: The list of id that will be notified
+            @param channel_id: The specific channel_id to use for GCM push
+        """
+
+        if not title and not uri and not message and not len(registration_ids):
```
flavour: Should these be ORs not ANDs?

modules/s3/s3msg.py
```diff
@@ -804,6 +805,63 @@ def dispatch_to_pe_id(pe_id,
...
+            return
+
+        try:
+            from gcm import GCM
+        except ImportError:
+            current.log.error("gcm module needed for google cloud push")
```
flavour: Is this small enough to include in modules/ along with the other 3rd-party modules there?

flavour: Seems it is to me

modules/s3/s3msg.py
```diff
@@ -804,6 +805,63 @@ def dispatch_to_pe_id(pe_id,
...
+            s3db = current.s3db
+            gcmtable = s3db.msg_gcm_channel
+            if channel_id:
+                query = (gcmtable.channel_id == channel_id)
+            else:
+                query = (gcmtable.enabled == True) & (gcmtable.deleted != True)
+
+            row = db(query).select(gcmtable.api_key,
+                                   limitby=(0, 1)).first()
+            if row:
```
flavour: No error if not?
I'd have thought a try/except (faster for the usual case where this is configured) with the except giving a current.log.error

modules/s3db/msg.py
```diff
@@ -897,6 +899,80 @@ def model(self):
...
+
+    names = ("msg_gcm_channel",
+             )
+
+    def model(self):
+
+        T = current.T
+
+        set_method = self.set_method
+        super_link = self.super_link
```
flavour: Only used once so no need to bring into scope

modules/templates/SAMBRO/config.py
```diff
@@ -240,62 +240,86 @@ def customise_cap_alert_resource(r, tablename):
...
+                     (table.deleted != True) & \
+                     (itable.alert_id == table.id) & \
+                     (itable.deleted != True)
+            rows_ = db(query_).select(table.status,
+                                      itable.headline,
+                                      itable.sender_name,
+                                      itable.web)
+
             # Normal onapprove
             s3db.cap_alert_approve(record)
```
flavour: Why move this under the queries? Those aren't sued for this & seems clearer to me to start with this...as it was...

modules/templates/SAMBRO/config.py
```diff
@@ -240,62 +240,86 @@ def customise_cap_alert_resource(r, tablename):
...
+                # Google Cloud Messaging
+                stable = s3db.pr_subscription
+                ctable = s3db.pr_contact
+    
+                query = (stable.pe_id == ctable.pe_id) & \
+                        (ctable.contact_method == "GCM") & \
+                        (ctable.value != None) & \
+                        (ctable.deleted != True) & \
+                        (stable.deleted != True) & \
+                        (stable.method.like("%GCM%"))
```
flavour: Why do you need the LIKE here?
It should be exactly "GCM", no?

PR #1283, opened 2016-05-07
Update jquery.calendars.afghan-ps.js

mohibullahiqbal111:
Hi Fran
i just made change in names of months and days of the week but i really try to delete the English names but i cant so please delete the English names also

in the calendar we are not using short names so please use full names

thank you


nursix:
Hi - you shouldn't change the order of the day names, this creates an inconsistency with the minified names, and accidentally changes the first day of the week.

mohibullahiqbal111:
Hi - so i will edit again and send you new pull request and also there are some language wise problem also i will edit day by day 

thank you for your kindly reply

PR #1285, opened 2016-05-17
SAMBRO: enhancements

biplovbhandari:
@flavour Please merge


controllers/cap.py
```diff
@@ -208,9 +230,34 @@ def prep(r):
             s3.formats["cap"] = r.url() # .have added by JS
 
         if r.interactive:
-
             if not r.component:
-                if r.method == "profile":
+                if r.get_vars["~.approved_by!"] == "None":
```
flavour: This is a little confusing...it appears to be a REST operator, but I don't think it is.
(The REST operator would be ~.approved_by__ne=None)
Is this doing a filter or just configuring these options?

flavour:
All seems fine other than the query above :)

controllers/cap.py
biplovbhandari: @flavour This is for filter to show the approved alerts `eden/cap/alert?~.approved_by!=None` and I am configuring list_fields according to this URL (Menu item)?

flavour: I see the menu item, but I don't see the filtering.
If you want to filter then you should use the REST operator:
approved_by__ne=None

approved_by!=None may look like it would do this, but I don't believe that it does.

biplovbhandari: @flavour not sure if this works correctly.

I have 22 approved alerts in my local db.
- `eden/cap/alert?~.approved_by__ne=None` gives me 10 records
image

- `eden/cap/alert?~.approved_by!=None` gives 22 records
image

Is there something I am missing here?

PR #1286, opened 2016-05-19
SAMBRO: alert history + SAMBRO: allow templates to be deleted + SAMBRO: send email about new priority options + email subject fix + SAMBRO: tweaks

biplovbhandari:


nursix:
Shouldn't use inspect.isfunction, but just callable(subject) - it ain't wrong to pass any callable object, doesn't have to be a function (and just using callable saves one import).

controllers/cap.py
```diff
@@ -1211,4 +1222,110 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+
+                if notify_form.accepts(request.post_vars,
+                                       session,
+                                       formname = "notify_form",
+                                       keepvalues = False,
+                                       hideerror = False):
+                    # Process Notify Form
+                    form_vars = notify_form.vars
+                    send_by_pe_id = msg.send_by_pe_id
+                    if len(pe_ids) == 1:
```
flavour: I don't think you need this special case here?
Just need:
for pe_id in pe_ids:
This is DRYer and should perform a little better

controllers/cap.py
```diff
@@ -1211,4 +1222,110 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                        for wprow in wprows:
+                            db(wptable.id == wprow.id).update(last_checked = request.utcnow)
+                    else:
+                        for pe_id in pe_ids:
+                            send_by_pe_id(pe_id=pe_id,
+                                          subject=form_vars.subject,
+                                          message=form_vars.message,
+                                          )
+                        for wprow in wprows:
+                            db(wptable.id == wprow.id).update(last_checked = request.utcnow)
```
flavour: Minor tweak: can bring utcnow into scope outside loop

controllers/cap.py
```diff
@@ -1211,4 +1222,110 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
...
+                                      subject=form_vars.subject,
+                                      message=form_vars.message,
+                                      )
+                        for wprow in wprows:
+                            db(wptable.id == wprow.id).update(last_checked = request.utcnow)
+                    else:
+                        for pe_id in pe_ids:
+                            send_by_pe_id(pe_id=pe_id,
+                                          subject=form_vars.subject,
+                                          message=form_vars.message,
```
flavour: Minor tweak: can bring subject/message into scope outside loop

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+        area_id = S3ReusableField("area_id", "reference %s" % tablename,
+                                  label = T("Area"),
+                                  ondelete = "CASCADE",
+                                  represent = area_represent,
+                                  requires = IS_ONE_OF(db, "cap_area.id",
+                                                       area_represent),
+                                  )
+
+        # ToDo: Use a widget tailored to entering <polygon> and <circle>.
+        # Want to be able to enter them by drawing on the map.
```
flavour: is this not the case already?

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+        # ToDo: Provide a mechanism for pre-loading geocodes that are not tied
+        # to individual areas.
+        # ToDo: Allow sharing the key-value pairs. Cf. Ruby on Rails tagging
+        # systems such as acts-as-taggable-on, which has a single table of tags
+        # used by all classes. Each tag record has the class and field that the
+        # tag belongs to, as well as the tag string. We'd want tag and value,
+        # but the idea is the same: There would be a table with tag / value
+        # pairs, and individual cap_area, event_event, org_whatever records
+        # would link to records in the tag table. So we actually would not have
+        # duplicate tag value records as we do now.
```
flavour: It depends on the usecase...the current usecases for tagging are very specific to each resource & hence make sense to be restricted to each resource...a global tag table would be poor there.
A generic tagging table would meet other usecases (although I'm surprised this comes up for CAP!?).

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+            ]
+
+        configure(tablename,
+                  deletable = False,
+                  editable = False,
+                  filter_widgets = filter_widgets,
+                  insertable = False,
+                  list_fields = list_fields,
+                  list_orderby = "cap_info_history.expires desc",
+                  orderby = "cap_info_history.expires desc",
```
flavour: I think list_orderby falls back to orderby, so if they are the same, then no need to specify both

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+                  )
+
+        # Components
+        add_components(tablename,
+                       cap_area_history = {"name": "area_history",
```
flavour: default name, so no need to specify

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+                                           "joinby": "alert_history_id",
+                                           },
+                       cap_area_location_history = {"name": "location_history",
+                                                    "joinby": "alert_history_id",
+                                                    },
+                       cap_area_tag_history = {"name": "tag_history",
+                                               "joinby": "alert_history_id",
+                                               },
+                       cap_info_history = {"name": "info_history",
```
flavour: default name, so no need to specify

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+                                           "joinby": "alert_history_id",
+                                           },
+                       cap_resource_history = {"name": "resource_history",
```
flavour: default name, so no need to specify

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+                                   ondelete = "CASCADE",
+                                   represent = alert_history_represent,
+                                   requires = IS_EMPTY_OR(
+                                                IS_ONE_OF(db, "cap_alert_history.id",
+                                                          alert_history_represent)),
+                                   )
+
+        # ---------------------------------------------------------------------
+        # CAP Info History Table
+        # @ToDo: i18n: Need label=T("")
```
flavour: I see label = T() ?

modules/s3db/cap.py
```diff
@@ -1291,19 +1304,619 @@ def model(self):
...
+                     s3_comments(),
+                     *s3_meta_fields())
+
+        configure(tablename,
+                  onaccept = self.cap_area_tag_onaccept,
+                  deduplicate = S3Duplicate(primary=("area_id", "tag", "value")),
+                  )
+
+        # ---------------------------------------------------------------------
+        # Alert History Table
```
flavour: I would create a new class for all the History tables
- there are few (if any!) usecases which require loading both tables together...so keep performance trim by loading only what is required: Working tables OR Historic tables

modules/templates/SAMBRO/config.py
```diff
@@ -383,6 +380,30 @@ def custom_prep(r):
...
+        standard_prep = s3.prep
+        def custom_prep(r):
+            # Call standard prep
+            if callable(standard_prep):
+                result = standard_prep(r)
+            else:
+                result = True
+
+            # On-delete option
+            current.s3db.event_type_id.attr.ondelete = "SET NULL"
```
flavour: This seems like the wrong place to do this?
No need to wrap in prep & in fact better to do within customise_resource?

modules/templates/SAMBRO/config.py
```diff
@@ -526,6 +547,30 @@ def custom_postp(r, output):
...
+        standard_prep = s3.prep
+        def custom_prep(r):
+            # Call standard prep
+            if callable(standard_prep):
+                result = standard_prep(r)
+            else:
+                result = True
+
+            # On-delete option
+            current.s3db.pr_person_id.attr.ondelete = "SET NULL"
```
flavour: Same as above

modules/templates/SAMBRO/config.py
flavour: In fact even better is to simply modify this in the usecases in the CAP module
So where you have:
self.event_type_id()
Replace with:
self.event_type_id(ondelete = "SET NULL")
This puts it directly into the core model

PR #1288, opened 2016-05-26
SAMBRO: enhancements + SAMBRO: more logic

biplovbhandari:
@flavour Please merge


controllers/cap.py
```diff
@@ -210,7 +232,56 @@ def prep(r):
...
+                        if r.get_vars["status"] == "incomplete":
+                            # Show incomplete alerts, ie. without info and area segment
+                            s3.filter = (s3base.FS("info.id") == None)
+                            s3.filter = (s3base.FS("area.id") == None)
+                            url = URL(c="cap", f="alert", args=["[id]"])
+                            s3base.S3CRUD.action_buttons(r, update_url=url, read_url=url)
+                        elif not r.get_vars:
+                            # Filter those alerts having at least info and area segment
+                            s3.filter = (s3base.FS("info.id") != None)
+                            s3.filter = (s3base.FS("area.id") != None)
```
flavour: s3.filter & s3base.FS used 4 times? Better to bring into scope

controllers/cap.py
```diff
@@ -210,7 +232,56 @@ def prep(r):
...
+                                                                       limitby=(0, 1)).first()
+                        if not irow or not arow:
+                            # This is incomplete alert
+                            session.warning = T("This Alert is incomplete! You can complete it now.")
+                            redirect(URL(c="cap", f="alert", args=[alert_id]))
+                    else:
+                        if r.get_vars["status"] == "incomplete":
+                            # Show incomplete alerts, ie. without info and area segment
+                            s3.filter = (s3base.FS("info.id") == None)
+                            s3.filter = (s3base.FS("area.id") == None)
```
nursix: Ahm - question: two consecutive assignments to the same variable - what's the point here? The second would override the first. Should perhaps better use resource.add_filter (since you're already in the prep, there is no need for s3.filter).

Besides: FS is global, no need for s3base.

controllers/cap.py
nursix: Same: second assignment overriding the first - what's the point? Should use add_filter, I suppose.

modules/s3db/cap.py
```diff
@@ -366,6 +367,7 @@ def model(self):
                      s3_datetime("sent",
                                  default = "now",
                                  writable = False,
+                                 represent = date_representation,
```
nursix: Why are you hardcoding a non-standard date representation? The current date format is a global setting which can be customized in the template (or in the controller if needs be), yet this one case doesn't apply the setting.

modules/s3db/cap.py
```diff
@@ -788,6 +792,7 @@ def model(self):
                            ),
                      Field("category", "list:string", # 1 or more allowed
                            label = T("Category"),
+                           required = True,
```
nursix: required = True or IS_NOT_EMPTY? That's two different things: 

- required=True means the field must be specified in *every* write command to the DB (i.e. both create and update), but it can be set to None (!)
- IS_NOT_EMPTY means that it can not be set to None - but it doesn't have to be specified in every write command (since there's a default for create, and update can just keep the value without having to read it first)

I highly suspect that you want IS_NOT_EMPTY rather than required=True.

PR #1289, opened 2016-05-30
SAMBRO: model for cap info parameter to separate parameters going to mobile + SAMBRO: cleanup demo data + Admin subscription only to admin  SAMBRO: allow template to be deleted + SAMBRO: subscriptions fix + SAMBRO: tweaks

biplovbhandari:
@flavour Please merge


controllers/cap.py
```diff
@@ -263,14 +267,14 @@ def prep(r):
                     else:
                         if r.get_vars["status"] == "incomplete":
                             # Show incomplete alerts, ie. without info and area segment
-                            s3.filter = (s3base.FS("info.id") == None)
-                            s3.filter = (s3base.FS("area.id") == None)
+                            s3.filter = (FS("info.id") == None) | (FS("area.id") == None)
```
flavour: Are you sure you want | here not &?

controllers/cap.py
```diff
@@ -263,14 +267,14 @@ def prep(r):
...
+                            response.s3.crud_strings["cap_alert"].title_list = T("Incomplete Alerts")
```
flavour: No need for 'response' in the controller context: s3 already in-scope (as you use on the previous line)

controllers/cap.py
```diff
@@ -955,6 +988,32 @@ def postp(r, output):
     return output
 
 # -----------------------------------------------------------------------------
+def info_parameter():
+    """
+        RESTful CRUD controller
```
flavour: Seems odd to have this controller for something which would pretty much always get used in component context?
Anyway I won't block on it

flavour: But explaining the usecase in the docstring would be useful

PR #1290, opened 2016-06-01
SAMBRO: CAP History Model + SAMBRO: fixes

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -17,6 +17,17 @@ def index():
     s3_redirect_default(URL(f="alert"))
 
 # -----------------------------------------------------------------------------
+def alert_history():
+    """
+        RESTful CRUD controller
+    """
+
+    output = s3_rest_controller("cap", "alert_history",
+                                rheader = s3db.cap_history_rheader,
+                                )
+    return output
+
+# -----------------------------------------------------------------------------
 def info_prep(r):
     """
         Preprocessor for CAP Info segments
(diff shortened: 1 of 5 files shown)


controllers/cap.py
```diff
@@ -1391,4 +1403,106 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
 
         return cap_area_options
 
+# -----------------------------------------------------------------------------
+def compose():
+
+    if settings.has_module("msg") and \
+       settings.get_cap_show_notification_menu() and \
+       auth.permission.has_permission("update", c="cap", f="alert"):
+        if request.get_vars.option == "notify_subscribers":
```
flavour: Can just use get_vars in the controller context :)

controllers/cap.py
```diff
@@ -1391,4 +1403,106 @@ def cap_AreaRowOptionsBuilder(alert_id, caller=None):
 
         return cap_area_options
 
+# -----------------------------------------------------------------------------
+def compose():
```
flavour: Did you try using msg.compose() ?

views/cap/compose.html
```diff
@@ -0,0 +1,2 @@
+{{extend "layout.html"}}
+{{include "cap/_compose.html"}}
```
flavour: I don't think you gain value from separating these files within a single theme
- the point of the separation is to allow multiple themes

flavour: I'm not sure why CAP needs a separate Compose function & View?
What is different about the CAP one?

Sure there are some permission checks etc as a wrapper to the compose form, but the form itself should be standard & hence best DRY?

biplovbhandari:
@flavour corrected! Please merge

PR #1291, opened 2016-06-02
SAMBRO: gcm for mobile client + SAMBRO: fixes

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -125,11 +125,11 @@ def alert():
         #s3db.configure(tablename,
         #               filter_widgets = filter_widgets,
         #               )
-
-        if r.representation == "html":
+        representation = r.representation
+        if representation == "html"  or representation == "aadata":
             r.table.msg_type.represent = None
 
-        elif r.representation == "dl":
+        elif representation == "dl":
             # DataList: match list_layout
             list_fields = ["msg_type",
                            "info.headline",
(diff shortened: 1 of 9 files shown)


controllers/cap.py
```diff
@@ -125,11 +125,11 @@ def prep(r):
         #s3db.configure(tablename,
         #               filter_widgets = filter_widgets,
         #               )
-
-        if r.representation == "html":
+        representation = r.representation
+        if representation == "html"  or representation == "aadata":
```
flavour: better:
if representation in ("html", "aadata"):

modules/gcm.py
```diff
@@ -0,0 +1,566 @@
+import requests
```
flavour: Useful to note this dependency in optional_requirements.txt

modules/s3db/msg.py
```diff
@@ -899,6 +901,79 @@ def model(self):
...
+
+        # ---------------------------------------------------------------------
+        return {}
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def msg_gcm_channel_onaccept(form):
+
+        if form.vars.enabled:
+            # Ensure only a single account enabled
```
flavour: Seems an odd limitation?
Might we not want to ahve different Orgs having diff accounts for diff billing/reporting?

models/tasks.py
```diff
@@ -83,6 +83,18 @@ def cap_ftp_sync(user_id=None):
 
     tasks["cap_ftp_sync"] = cap_ftp_sync
 
+    # -------------------------------------------------------------------------
+    def cap_gcm(title, uri, message, registration_ids, user_id=None):
```
flavour: Why is this cap_gcm & not msg_gcm?
This seems to be msg_ to me...

flavour:
Merging - can answer/cleanup afterwards

PR #1292, opened 2016-06-05
SAMBRO: show profile action button + tweaks

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -100,14 +100,10 @@ def alert():
         rows = db(itable.expires >= request.utcnow).select(itable.id,
                                                            orderby=itable.id)
         if rows:
-            unexpired_ids = ",".join([str(row.id) for row in rows])
+            unexpired_ids = default_filter = ",".join([str(row.id) for row in rows])
         else:
             unexpired_ids = "*"
-
-        if unexpired_ids == "*":
             default_filter = None
-        else:
-            default_filter = unexpired_ids
 
         filter_widgets = s3db.get_config(tablename, "filter_widgets")
         filter_widgets.insert(0, S3OptionsFilter("info.id",
@@ -126,7 +122,7 @@ def alert():
         #               filter_widgets = filter_widgets,
         #               )
         representation = r.representation
-        if representation == "html"  or representation == "aadata":
+        if representation in ("html", "aadata"):
             r.table.msg_type.represent = None
 
         elif representation == "dl":
@@ -433,9 +429,10 @@ def alert():
                                                  _class="cap-label upper"
                                             ),
                                             SPAN(info.headline,
-                                                 _class="cap-value"
+                                                 _class="cap-value cap-bold cap-large"
                                                  ),
                                             ))
+                            fn_1_append(BR())
                         if info.description:
                             fn_1_append(DIV(SPAN("%s :: " % T("Description"),
                                                  _class="cap-label upper"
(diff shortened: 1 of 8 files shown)


modules/s3/s3resource.py
```diff
@@ -5252,6 +5252,8 @@ def __init__(self,
                         # Limited master query without count/getids
                         # (=rows is the subset, only need page IDs)
                         page = self.getids(rows, pkey)
+                        if not ids:
+                            ids = self.getids(rows, pkey)
```
nursix: Nope, this is wrong - no ids required at this point, and even if they were required, this is the wrong way to determine them.

If you need all ids, you must specify getids=True in the select.

nursix: What's your point with this change? Maybe should fix the select instead.

optional_requirements.txt
```diff
@@ -11,6 +11,8 @@ reportlab>=2.5
 pyserial>=2.6
 # Warning: S3Msg unresolved dependency tweepy required for non-Tropo Twitter support
 tweepy>=1.9
+# Warning: S3Msg unresolved dependency requests required for Google Cloud Messaging support
```
nursix: This seems wrong too (and doesn't pass tests) - request library is already an optional requirement (for posting to facebook), and specified as >=2.3.0 there.

modules/s3/s3resource.py
biplovbhandari: @nursix this was for /eden/msg/compose which does not get the person auto complete suggestion because ids are not available for pr/search

nursix: It's the wrong approach, however.

nursix: Besides: what is pr/search?

If you meant pr/person_seach, then this is using pr_search_ac, and pr_search_ac is not using ids anywhere.

So can you please specify the problem more precisely?

Dominic

> 6 juni 2016 kl. 02:48 skrev Biplov Bhandari <notifications@github.com>:
> 
> @nursix this was for /eden/msg/compose which does not get the person auto complete suggestion because ids are not available for pr/search

nursix: I guess (!) you mean pe_search_ac ;)

Well, I'm afraid that is a wrong implementation in the pe_search_ac function. It uses data.ids, but it doesn't request ids. And even if it did, it would be a wrong assumption that ids[i] belongs to rows[i] - because by definition, data.ids holds all record ids even outside of the requested page.

In pe_search_ac, the record ID must be added to the requested fields, and then read from the row - using data.ids is wrong in this way (bug).

I can try fixing it, if you give me a little time - and back out this change in S3ResourceData.

Dominic

biplovbhandari: @nursix I meant for /eden/msg/compose which does not get the person auto complete suggestion because ids are not available for pr/pentity/search_ac.json
I have back out the change in S3Resource

biplovbhandari:
@flavour Please merge

modules/s3/s3resource.py
nursix: Yup, I figured that it was about pe_search_ac rather than pr/search.

The use of S3ResourceData.ids in that function is not valid - and actually unnecessary. 

The point with using data.ids was that the "pe_id" field was represented as name, and could thus not be used as option value - but the solution to that is as simple as to use the raw data, like here:
https://github.com/nursix/eden/commit/deb9373b1f12a279c54053f1f660413f724a4bcf

With this, msg/compose should work as expected - testing+feedback appreciated, of course.

biplovbhandari:
@flavour please merge

PR #1293, opened 2016-06-12
Update prs.py

mohibullahiqbal111:
Hi 
I just solve some language wise problem in to the system


flavour:
Great, thanks :)
2 minor enhancements I will do after the merge:
Remove comment from 'Profile' now that it has a translation
Remove extra space at the right of one of the strings

PR #1294, opened 2016-06-14
SAMBRO: show profile action button + tweaks + SAMBRO:Alert hub  SAMBRO: Implement facebook post

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -90,27 +90,28 @@ def alert():
 
     def prep(r):
         from s3 import S3OptionsFilter
+        table = r.table
         itable = s3db.cap_info
+
         rows = db(itable.expires < request.utcnow).select(itable.id,
                                                           orderby=itable.id)
-        if rows:
-            expired_ids = ",".join([str(row.id) for row in rows])
-        else:
-            expired_ids = "*"
+        expired_ids = ",".join([str(row.id) for row in rows]) or None
+
         rows = db(itable.expires >= request.utcnow).select(itable.id,
                                                            orderby=itable.id)
-        if rows:
-            unexpired_ids = ",".join([str(row.id) for row in rows])
-        else:
-            unexpired_ids = "*"
+        unexpired_ids = default_filter = ",".join([str(row.id) for row in rows]) or None
 
-        if unexpired_ids == "*":
-            default_filter = None
-        else:
-            default_filter = unexpired_ids
+        rows = db(table.external == True).select(table.id,
+                                                 orderby=table.id)
+        external_alerts = ",".join([str(row.id) for row in rows]) or None
+
+        rows = db(table.external != True).select(table.id,
+                                                 orderby=table.id)
+        internal_alerts = default_alert = ",".join([str(row.id) for row in rows]) or None
 
         filter_widgets = s3db.get_config(tablename, "filter_widgets")
-        filter_widgets.insert(0, S3OptionsFilter("info.id",
+        filter_widgets_insert = filter_widgets.insert
+        filter_widgets_insert(0, S3OptionsFilter("info.id",
                                                  label = T("Expiration"),
                                                  options = OrderedDict(
                                                         [(expired_ids, T("Expired")),
@@ -121,13 +122,24 @@ def alert():
                                                  multiple = False,
                                                  default = default_filter,
                                                  ))
+        filter_widgets_insert(1, S3OptionsFilter("id",
+                                                 label = T("Source"),
+                                                 options = OrderedDict(
+                                                        [(internal_alerts, T("Internal")),
+                                                         (external_alerts, T("External")),
+                                                         ("*", T("All")),
+                                                         ]),
+                                                 cols = 3,
+                                                 multiple = False,
+                                                 default = default_alert,
+                                                 ))
         # No need to put them back - the edit happens in-place
         #s3db.configure(tablename,
         #               filter_widgets = filter_widgets,
         #               )
         representation = r.representation
-        if representation == "html"  or representation == "aadata":
-            r.table.msg_type.represent = None
+        if representation in ("html", "aadata"):
+            table.msg_type.represent = None
 
         elif representation == "dl":
             # DataList: match list_layout
@@ -189,7 +201,6 @@ def alert():
         #                   )
 
         if r.id:
-            atable = r.table
 
             if r.record.is_template:
                 redirect(URL(c="cap", f="template",
@@ -199,29 +210,25 @@ def alert():
             # Don't show event_type_id and template once created
             # If the user used the wrong event type,
             # then they should "Cancel" that alert and start a new one
-            atable.event_type_id.readable = False
-            atable.event_type_id.writable = False
-            atable.template_id.readable = False
-            atable.template_id.writable = False
+            table.event_type_id.readable = False
+            table.event_type_id.writable = False
+            table.template_id.readable = False
+            table.template_id.writable = False
 
             if r.record.approved_by is not None:
                 # Once approved, don't allow to edit
                 # Don't allow to delete
                 s3db.configure(tablename,
-                               editable = False,
                                deletable = False,
+                               editable = False,                               
                                insertable = False,
                                )
             if r.record.reference is not None:
-                # Don't show template_id for Updated/Cancelled/Error/Relay Alert
-                atable.template_id.readable = False
-                atable.template_id.writable = False
-                atable.msg_type.writable = False
+                table.msg_type.writable = False
 
             if settings.get_cap_restrict_fields():
                 if r.record.msg_type in ("Update", "Cancel", "Error"):
                     # Use case for change in msg_type
-                    atable = r.table
                     for f in ("template_id",
                               "sender",
                               "status",
@@ -235,9 +242,9 @@ def alert():
                               "reference",
                               "incidents",
                               ):
-                        atable[f].writable = False
+                        table[f].writable = False
         else:
-            r.resource.add_filter(r.table.is_template == False)
+            r.resource.add_filter(table.is_template == False)
             s3.formats["cap"] = r.url() # .have added by JS
 
         if r.interactive:
@@ -261,11 +268,50 @@ def alert():
                                    insertable = False,
                                    list_fields = list_fields,
                                    )
+                elif r.get_vars["~.external"] == "True":
+                    query = (table.id == itable.alert_id) & \
+                            (table.external == True) & \
+                            (table.is_template != True) & \
+                            (table.deleted != True)
+                    rows = db(query).select(itable.sender_name,
+                                            groupby=itable.sender_name)
+                    sender_options = {}
+                    from s3 import s3_str
+                    for row in rows:
+                        sender_name = row.sender_name
+                        sender_options[sender_name] = s3_str(T(sender_name))
+
+                    rows = db(query).select(itable.event_type_id,
+                                            groupby=itable.event_type_id)
+                    event_type_options = {}
+                    for row in rows:
+                        event_type_id = row.event_type_id
+                        event_type_options[event_type_id] = itable.event_type_id.represent(event_type_id)
+                    filter_widgets = [
+                        S3OptionsFilter("info.sender_name",
+                                        label = T("Sender"),
+                                        options = sender_options,
+                                        ),
+                        S3OptionsFilter("info.event_type_id",
+                                        options = event_type_options,
+                                        ),
+                        ]
+                    s3.crud_strings["cap_alert"].title_list = T("Alerts Hub")
+                    s3base.S3CRUD.action_buttons(r, deletable=False)
+                    profile_button = {"url": URL(args=["[id]", "profile"]),
+                                      "_class": "action-btn",
+                                      "_target": "_blank",
+                                      "label": str(T("View Profile"))
+                                      }
+                    s3.actions.insert(1, profile_button)
+
+                    s3db.configure(tablename,
+                                   filter_widgets = filter_widgets,
+                                   )
 
                 if r.method == "review":
                     alert_id = r.id
                     if alert_id:
-                        itable = s3db.cap_info
                         artable = s3db.cap_area
                         irow = db(itable.alert_id == alert_id).select(itable.id,
                                                                       limitby=(0, 1)).first()
@@ -328,7 +374,6 @@ def alert():
 
                     # Info
                     # @ToDo: handle multiple languages
-                    itable = s3db.cap_info
                     info = db(itable.alert_id == alert_id).select(itable.id,
                                                                   itable.language,
                                                                   itable.category,
@@ -423,8 +468,6 @@ def alert():
                                       bbox = bbox,
                                       )
 
-                    table = r.table
-
                     def custom_widget_fn_1(r, **attr):
                         fn_1 = DIV()
                         fn_1_append = fn_1.append
@@ -433,9 +476,10 @@ def alert():
                                                  _class="cap-label upper"
                                             ),
                                             SPAN(info.headline,
-                                                 _class="cap-value"
+                                                 _class="cap-value cap-headline"
                                                  ),
                                             ))
+                            fn_1_append(BR())
                         if info.description:
                             fn_1_append(DIV(SPAN("%s :: " % T("Description"),
                                                  _class="cap-label upper"
@@ -444,6 +488,7 @@ def alert():
                                                  _class="cap-value"
                                                  )
                                             ))
+                            fn_1_append(BR())
                         if info.response_type:
                             fn_1_append(DIV(SPAN("%s :: " % T("Response Type"),
                                                  _class="cap-label upper"
@@ -452,6 +497,7 @@ def alert():
                                                  _class="cap-strong"
                                                  ),
                                             ))
+                            fn_1_append(BR())
                         if info.instruction:
                             fn_1_append(DIV(SPAN("%s :: " % T("Instructions"),
                                                  _class="cap-label upper"
@@ -535,13 +581,6 @@ def alert():
                                             _class="cap-value"
                                             ),
                                        ),
-                                   DIV(SPAN("%s :: " % T("Contact Info"),
-                                            _class="cap-label"
-                                            ),
-                                       SPAN(info.contact,
-                                            _class="cap-value"
-                                            ),
-                                       ),
                                    )
                         fn_2_insert = fn_2.insert
                         if info.category:
@@ -568,6 +607,14 @@ def alert():
                                                      _class="cap-value"
                                                      )
                                                 ))
+                        if info.contact:
+                            fn_2_insert(13, DIV(SPAN("%s :: " % T("Contact Info"),
+                                                     _class="cap-label"
+                                                     ),
+                                                SPAN(info.contact,
+                                                     _class="cap-value"
+                                                     ),
+                                                ))
                         if len(parameters):
                             parameter_text = []
                             for parameter in parameters:
@@ -581,7 +628,6 @@ def alert():
                                                  _class="cap-value"
                                                  )
                                             ))
-                            
                         return fn_2
 
                     custom_widget_2 = dict(type = "custom",
@@ -630,36 +676,40 @@ def alert():
                                             _class="cap-value"
                                             ),
                                        ),
-                                   DIV(SPAN("%s :: " % T("Note"),
-                                            _class="cap-label"
-                                            ),
-                                       SPAN(record.note,
-                                            _class="cap-value"
-                                            ),
-                                       ),
-                                   DIV(SPAN("%s :: " % T("Reference ID"),
-                                            _class="cap-label"
-                                            ),
-                                       SPAN(record.reference,
-                                            _class="cap-value"
-                                            ),
-                                       ),
-                                   DIV(SPAN("%s :: " % T("Incident IDs"),
-                                            _class="cap-label"
-                                            ),
-                                       SPAN(table.incidents.represent(record.incidents),
-                                            _class="cap-value"
-                                            ),
-                                       ),
                                    )
+                        fn_3_insert = fn_3.insert
                         if record.codes:
-                            fn_3.insert(7, DIV(SPAN("%s :: " % T("Handling Code"),
+                            fn_3_insert(7, DIV(SPAN("%s :: " % T("Handling Code"),
                                                     _class="cap-label"
                                                     ),
                                                SPAN(table.codes.represent(record.codes),
                                                     _class="cap-value"
                                                     ),
                                                ))
+                        if record.note:
+                            fn_3_insert(8, DIV(SPAN("%s :: " % T("Note"),
+                                                    _class="cap-label"
+                                                    ),
+                                               SPAN(record.note,
+                                                    _class="cap-value"
+                                                    ),
+                                               ))
+                        if record.reference:
+                            fn_3_insert(9, DIV(SPAN("%s :: " % T("Reference ID"),
+                                                    _class="cap-label"
+                                                    ),
+                                               SPAN(record.reference,
+                                                    _class="cap-value"
+                                                    ),
+                                               ))
+                        if record.incidents:
+                            fn_3_insert(10, DIV(SPAN("%s :: " % T("Incident IDs"),
+                                                     _class="cap-label"
+                                                     ),
+                                                SPAN(table.incidents.represent(record.incidents),
+                                                     _class="cap-value"
+                                                     ),
+                                                ))
                         return fn_3
 
                     custom_widget_3 = dict(type = "custom",
@@ -699,19 +749,18 @@ def alert():
                                               ),)
 
             elif r.component_name == "info":
-                itable = r.component.table
                 # Do not show this as overwritten in onaccept
                 itable.web.readable = False
                 itable.web.writable = False
 
                 alert_id = request.args(0)
+
                 # Check for prepopulate
                 if alert_id:
-                    atable = r.table
                     itable.web.default = settings.get_base_public_url()+\
                                          URL(c="cap", f="alert", args=alert_id)
-                    row = db(atable.id == alert_id).select(atable.event_type_id,
-                                                           limitby=(0, 1)).first()
+                    row = db(table.id == alert_id).select(table.event_type_id,
+                                                          limitby=(0, 1)).first()
                     itable.event_type_id.default = row.event_type_id
 
                 if r.record.approved_by is not None:
@@ -787,7 +836,6 @@ def alert():
                 tid = post_vars["template_id"]
                 if tid:
                     # Read template and copy locked fields to post_vars
-                    table = db.cap_alert
                     template = db(table.id == tid).select(table.template_settings,
                                                           limitby=(0, 1)).first()
                     try:
@@ -895,6 +943,15 @@ def alert():
                 r.next = URL(c="cap", f="alert", args=[lastid, "info", "create"])
 
         if r.interactive:
+            if not r.component and not r.method:
+                s3_action_buttons(r, read_url=URL(args="[id]"))
+                profile_button = {"url": URL(args=["[id]", "profile"]),
+                                  "_class": "action-btn",
+                                  "_target": "_blank",
+                                  "label": str(T("View Profile"))
+                                  }
+                s3.actions.insert(1, profile_button)
+
             if get_vars.get("_next"):
                 r.next = get_vars.get("_next")
 
--- a/modules/s3cfg.py
+++ b/modules/s3cfg.py
@@ -2441,6 +2441,13 @@ class S3Config(Storage):
 
         return self.cap.get("same_code")
 
+    def get_cap_post_to_facebook(self):
+        """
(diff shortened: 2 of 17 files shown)


controllers/cap.py
```diff
@@ -199,29 +210,28 @@ def prep(r):
...
+                               editable = False,                               
                                insertable = False,
                                )
             if r.record.reference is not None:
                 # Don't show template_id for Updated/Cancelled/Error/Relay Alert
-                atable.template_id.readable = False
-                atable.template_id.writable = False
-                atable.msg_type.writable = False
+                table.template_id.readable = False
+                table.template_id.writable = False
```
flavour: This seems to always repeat what was done in lines 215-216

controllers/cap.py
```diff
@@ -433,9 +479,10 @@ def custom_widget_fn_1(r, **attr):
                                                  _class="cap-label upper"
                                             ),
                                             SPAN(info.headline,
-                                                 _class="cap-value"
+                                                 _class="cap-value headline"
```
flavour: 'headline' alone risks clashes...the namespaced version of cap-headline is safer

modules/s3cfg.py
```diff
@@ -2441,6 +2441,20 @@ def get_cap_same_code(self):
 
         return self.cap.get("same_code")
 
+    def get_cap_info_language_represent(self):
+        """
+            Languages for the CAP Info language representation
```
flavour: I thought the representation would always use the complete list, no?

modules/s3cfg.py
```diff
@@ -2441,6 +2441,20 @@ def get_cap_same_code(self):
...
+        """
+
+        return self.cap.get("info_language_represent")
+
+    def get_cap_post_to_facebook(self):
+        """
+            Whether to post the alerts in facebook
+        """
+
+        return self.cap.get("post_to_twitter", False)
```
flavour: Looks like a copy/paste bug ;)

modules/s3db/cap.py
```diff
@@ -787,11 +794,10 @@ def model(self):
...
                            label = T("Language"),
-                           represent = lambda opt: languages.get(opt,
-                                                                 UNKNOWN_OPT),
-                           requires = IS_EMPTY_OR(
-                                        IS_IN_SET(languages)
-                                        ),
+                           represent = S3Represent(options = settings.get_cap_info_language_represent()),
+                           #requires = IS_EMPTY_OR(
+                           #                       IS_IN_SET(languages)
+                           #                       ),
```
flavour: I still think the users are going to want to have a fixed list in a dropdown to select from rather than having to enter data in freetext, which is error-prone.
An Alert Hub configuration can have this set to the complete list if-required...and perhaps the requires can be modified between the 2 usecases in a prep in config.py

modules/templates/SAMBRO/config.py
```diff
@@ -108,6 +111,43 @@ def config(settings):
     # Translate the cap_area name
     settings.L10n.translate_cap_area = True
 
+    # CAP Info Language representation options
+    # English name taken from here: http://www.loc.gov/standards/iso639-2/php/code_list.php
+    represent_languages = OrderedDict([
+        ("ar", "Arabic"),
```
flavour: I think these should be translatable, so:
("ar", T("Arabic")),
etc

I also think this should be a complete list which isn't configurable & can be used in other cases to CAP (this representation of language codes has much wider applicability).
modules/s3/s3translate.py would seem to be the best place to put this
Then in CAP, can from s3.s3translate import languages

modules/templates/SAMBRO/config.py
```diff
@@ -314,62 +444,46 @@ def onapprove(record):
...
 
-            # Send out private alerts to addresses
-            # @ToDo: Check for LEFT join when required
-            # this is ok for now since every Alert should have an Info & an Area
-            # @ToDo: Handle multi-lingual alerts when required
-            if record["scope"] == "Private":
-                atable = s3db.cap_area
+                # Facebook Post
+                if settings.get_cap_post_to_facebook():
+                    # @ToDo: Handle the multi-message nicely?
```
flavour: Copy/paste fragment ;)
(I don't think FB messages are size limited...)

modules/templates/SAMBRO/config.py
```diff
@@ -314,62 +444,46 @@ def onapprove(record):
...
+                    # @ToDo: Send resource url with post
+                    for row_ in rows_:
+                        content = get_facebook_content(row_),
+                        try:
+                            current.msg.post_to_facebook(text=content)
+                        except Exception, e:
+                            current.log.debug("Posting Alert failed on facebook: %s" % e)
```
flavour: "Posting Alert to Facebook failed:

modules/templates/SAMBRO/config.py
```diff
@@ -314,62 +444,46 @@ def onapprove(record):
...
-                for row_ in rows:
-                    send_by_pe_id(row_.pe_id, subject, email_content)
-                    try:
-                        send_by_pe_id(row_.pe_id, subject, sms_content, contact_method="SMS")
-                    except ValueError:
-                        current.log.error("No SMS Handler defined!")
+
+                for row in rows_:
+                    priority = itable.priority.represent(row.cap_info.priority)
+                    if priority == "-":
```
flavour: Should this be current.messages["NONE"] in case that changes?

modules/templates/SAMBRO/controllers.py
```diff
@@ -56,7 +56,7 @@ def __call__(self):
                               "id"        : "search_results",
                               "layer_id"  : layer_id,
                               # @ToDo: Make the filter update timestamp on refresh in the client side
-                              "filter"    : "~.info.expires__gt=%s" % request.utcnow,
+                              "filter"    : "~.info.expires__gt=%s&~.external!=True" % request.utcnow,
```
flavour: ~.external__ne=True

modules/templates/SAMBRO/menus.py
```diff
@@ -258,4 +291,33 @@ def pr():
...
 
+    # -------------------------------------------------------------------------
+    @classmethod
+    def settings_messaging(cls):
+        """ Messaging settings menu items:
+
+            These items are used in multiple menus, but each item instance can
+            always only belong to one parent, so we need to re-instantiate
+            with the same parameters, and therefore this is defined as a
+            function here.
```
flavour: Would be nice to include a note as to why this is subclassed - what is different vs the std one?
(I know this is the separate RSS channels, since I hunted, but making this more explicit is nice)

modules/templates/SAMBRO/config.py
nursix: Used like this:

    represent = IS_ISO639_2_LANGUAGE_CODE.represent_local

IMHO that does exactly the same, doesn't it?

nursix: Example: https://github.com/sahana/eden/blob/master/modules/s3db/dvr.py#L551

nursix: Can still specify settings.get_cap_languages() as subset ("select" parameter). The good part is that it would limit the selection, but still be able to represent codes which are not in the subset.

I don't think there is a need for either the representation function nor a custom IS_IN_SET construct here - but maybe I'm missing something here.

flavour: No this is perfect - I knew we should have a global one instead of a lcoal one...forgot we already had one & so no need to make it :)

biplovbhandari:
@flavour corrected..Please merge

modules/s3db/cap.py
```diff
@@ -787,8 +798,7 @@ def model(self):
                      Field("language",
                            default = "en-US",
                            label = T("Language"),
-                           represent = lambda opt: languages.get(opt,
-                                                                 UNKNOWN_OPT),
+                           represent = S3Represent(options = dict(represent_languages)),
                            requires = IS_EMPTY_OR(
```
nursix: Shouldn't this be:

    requires = IS_EMPTY_OR(IS_ISO639_2_LANGUAGE_CODE(select=represent_languages, translate=True))

?

nursix: (I also thought that you wanted a configurable subset as settings.cap.languages?)

modules/s3db/cap.py
```diff
@@ -1649,12 +1660,23 @@ def cap_alert_create_onaccept(form):
...
             if user:
-                current.db(current.s3db.cap_alert.id == form_vars.id).update(
-                                                        approved_by = user.id)
+                db(query).update(approved_by = user.id)
+
+        # mci==0 -> original copies from system
+        if (alert and alert.mci != 0) or \
+            "import_feed" in request.args or \
+            (request.function == "rss_channel" and request.get_vars.type == "cap"):
+            current.db(current.s3db.cap_alert.id == form_vars.id).update(external = True)
```
nursix: If you have an external-flag, then you don't need to set it onaccept (nor to check the MCI).

Just let it default to True, and in the interactive prep, set the default to False.

modules/s3db/cap.py
```diff
@@ -499,6 +503,12 @@ def model(self):
                                  readable = False,
                                  writable = False,
                                  ),
+                     # To separate the external CAP Alert
+                     Field("external", "boolean",
+                           default = False,
```
nursix: If you let this default to True instead, and in the interactive prep change the default to False, then you do not need to update it onaccept, or by mci, because it would always only be False for interactively (=locally) created alerts, while all imported alert would have external=True by default.

modules/s3db/cap.py
nursix: Btw, I don't get the IS_EMPTY_OR - why would you allow the language to be set to empty deliberately in the GUI?

According to the standard, a null value shall be understood as en-US, which already is your default. Hence, if empty and en-US are equivalent, then IMHO allowing empty makes no sense. It's either en-US, or it is another language - but it is never "no language".

modules/s3db/cap.py
```diff
@@ -1809,6 +1831,14 @@ def cap_info_onvalidation(form):
                 form.errors["category"] = \
                     current.T("Atleast one category is required.")
 
+        alert_id = form.vars.get("alert_id", None)
+        atable = current.s3db.cap_alert
+        alert = current.db(atable.id == alert_id).select(atable.mci,
+                                                         limitby=(0, 1)).first()
+        if alert and alert.mci != 0:
+            # Not internal alerts
+            current.s3db.cap_info.language.requires = IS_IN_SET(dict(represent_languages))
```
nursix: This shouldn't be necessary - should disallow empty for all cases.

modules/s3db/cap.py
```diff
@@ -3841,21 +3873,32 @@ def apply_method(r, **attr):
...
+                                                    ignore_errors = form_vars_get("ignore_errors", None))
                 except:
                     response.error = str(sys.exc_info()[1])
                 else:
-                    import_count = resource.import_count
-                    if import_count:
-                        response.confirmation = "%s %s" % \
-                            (import_count,
-                             T("Alerts successfully imported."))
+                    reply = eval(response_)
```
nursix: json.loads, not eval

nursix: Btw, you can check for resource.error - which may be easier than parsing the response locally.

modules/s3db/cap.py
```diff
@@ -3841,21 +3873,32 @@ def apply_method(r, **attr):
...
+                    if reply["statuscode"] == "200":
+                        if "created" in reply:
```
nursix: len(resource.import_created)

modules/s3db/cap.py
```diff
@@ -3841,21 +3873,32 @@ def apply_method(r, **attr):
...
+                            alert_id = reply["created"][0]
+                        elif "updated" in reply:
```
nursix: len(resource.import_updated)

modules/s3db/cap.py
```diff
@@ -3841,21 +3873,32 @@ def apply_method(r, **attr):
...
+                            alert_id = reply["updated"][0]
```
nursix: resource.import_updated[0]

modules/s3db/cap.py
```diff
@@ -3841,21 +3873,32 @@ def apply_method(r, **attr):
...
                 else:
-                    import_count = resource.import_count
-                    if import_count:
-                        response.confirmation = "%s %s" % \
-                            (import_count,
-                             T("Alerts successfully imported."))
+                    reply = eval(response_)
+                    if reply["statuscode"] == "200":
+                        if "created" in reply:
+                            alert_id = reply["created"][0]
```
nursix: resource.import_created[0]

modules/s3db/cap.py
nursix: I don't quite see a need to parse the JSON message - since it's local context, you can just pick the information from the resource.

modules/templates/SAMBRO/config.py
```diff
@@ -130,21 +133,22 @@ def customise_msg_rss_channel_resource(r, tablename):
...
-
         s3db = current.s3db
         def onaccept(form):
             # Normal onaccept
             s3db.msg_channel_onaccept(form)
-            _id = form.vars.id
             db = current.db
             table = db.msg_rss_channel
+            form_vars = form.vars
+            _id = form_vars.get("id", None)
```
nursix: Ideally, "record_id" not "_id". And according to PEP8, trailing underscore rather than leading underscore (to avoid confusion with semi-private variables)

modules/templates/SAMBRO/config.py
```diff
@@ -166,6 +170,73 @@ def onaccept(form):
...
 
     # -------------------------------------------------------------------------
+    def customise_msg_rss_channel_controller(**attr):
+
+        s3 = current.response.s3
+        table = current.s3db.msg_rss_channel        
+        type = current.request.get_vars.get("type", None)
+        if type == "cap":
+            # CAP RSS Channel
+            s3.filter = (table.type == "cap")
```
nursix: Better to use an S3 filter expression: FS("type") == "cap" (because the framework can not introspect web2py queries). That may not be required here, but better to make that a general habit: if you can FS, then better do FS.

modules/templates/SAMBRO/config.py
```diff
@@ -166,6 +170,73 @@ def onaccept(form):
...
+                if type == "cap":
+                    # CAP RSS Channel
+                    update_url = URL(c="msg", f="rss_channel",
+                                     args=["[id]", "update"],
+                                     vars={"type": "cap"})
+                    delete_url= URL(c="msg", f="rss_channel",
+                                    args=["[id]", "delete"],
+                                    vars={"type": "cap"})
+                    action_buttons = s3.actions
+                    for action_button in action_buttons:
```
nursix: Better to override-replace than to update (relying on the _class attribute is prone to regression):

    s3.actions = S3CRUD.action_buttons(...)

This would also check permissions and hide buttons if action is not permitted.

modules/templates/SAMBRO/config.py
```diff
@@ -166,6 +170,73 @@ def onaccept(form):
...
+                        if "_class" in action_button:
+                            if action_button["_class"] == "action-btn edit":
+                                action_button["url"] = update_url
+                                continue
```
nursix: Superfluous continue - where else would it go?

modules/templates/SAMBRO/config.py
```diff
@@ -166,6 +170,73 @@ def onaccept(form):
...
+                            elif action_button["_class"] == "delete-btn":
+                                action_button["url"] = delete_url
+                                continue
+
+                    if "form" in output:
+                        # Modify Add Button
+                        add_btn = A(T("Add CAP Feed"),
```
nursix: Not checking permissions?

modules/templates/SAMBRO/config.py
```diff
@@ -310,66 +404,50 @@ def onapprove(record):
...
+                                              priority)
+                    email_content = "%s%s%s" % ("<html>", XML(get_html_email_content(row)), "</html>")
+                    sms_content = get_sms_content(row)
+    
+                    for row_ in rows:
+                        send_by_pe_id(row_.pe_id, subject, email_content)
+                        try:
+                            send_by_pe_id(row_.pe_id, subject, sms_content, contact_method="SMS")
+                        except ValueError:
+                            current.log.error("No SMS Handler defined!")
```
nursix: If that is the assumption, then you should break out of the loop here, right?

nursix: ...out of both loops actually.

modules/templates/SAMBRO/config.py
```diff
@@ -1019,4 +1097,101 @@ def get_sms_content(row):
...
+
+        if event_type_id and event_type_id != "-":
+            if not isinstance(event_type_id, lazyT):
+                event_type = itable.event_type_id.represent(event_type_id)
+            else:
+                event_type = event_type_id
+        else:
+            event_type = T("None")
+
+        if priority_id and \
```
nursix: Why this line break? Short enough to have that in a single line.

modules/templates/SAMBRO/config.py
```diff
@@ -1019,4 +1097,101 @@ def get_sms_content(row):
...
+        else:
+            priority = T("Alert")
+
+        category = itable.category.represent(row["cap_info.category"])
+
+        if not response_type: 
+            response_type = T("Unknown")
+
+        subject = \
+        T("%(Scope)s %(Status)s Alert") % \
```
nursix: Indentation (PEP8)! Even after a line break.

modules/templates/SAMBRO/config.py
```diff
@@ -1019,4 +1097,101 @@ def get_sms_content(row):
...
+            dict(Scope = s3_str(row["cap_alert.scope"]),
```
nursix: dict members should follow regular name rules, i.e. all lowercase and underscore for word separation. And better to use the dict literal than the constructor:

    {"scope": s3_str(row["cap_alert.scope"],
     "status": s3_str(row["cap_alert.status"]),
    }

modules/templates/SAMBRO/config.py
```diff
@@ -1019,4 +1097,101 @@ def get_sms_content(row):
...
+            body8 = T("For more details visit %(URL)s or contact %(Contact)s") % \
+                    dict(URL = s3_str(row["cap_info.web"]),
+                         Contact = s3_str(row["cap_info.contact"]))
+        else:
+            body8 = T("For more details visit %(URL)s") % \
+                    dict(URL = s3_str(row["cap_info.web"]))
+        body9 = T("VIEW ALERT ON THE WEB: %(profile)s") % \
+                dict(profile = "%s/%s" % ((s3_str(row["cap_info.web"])), "profile"))
+        facebook_content = \
+"%(subject)s\n%(headline)s\n%(id)s\n\n%(body1)s\n\n%(body2)s\n\n%(body3)s\n\n%(body4)s\n\n%(body5)s\n\n%(body7)s\n\n%(body8)s\n\n%(body9)s" % \
```
nursix: What a string template! Have you tried making bodies an array, and use "\n\n".join(bodies)?

modules/templates/SAMBRO/config.py
```diff
@@ -1019,4 +1097,101 @@ def get_sms_content(row):
...
+        if not response_type: 
+            response_type = T("Unknown")
+
+        subject = \
+        T("%(Scope)s %(Status)s Alert") % \
+            dict(Scope = s3_str(row["cap_alert.scope"]),
+                 Status = s3_str(row["cap_alert.status"]))
+        headline = T((s3_str(row["cap_info.headline"])))
+        id = T("ID: %(Identifier)s") % dict(Identifier = s3_str(row["cap_alert.identifier"]))
+        body1 = \
```
nursix: If it is an array, then it should be an array:

    body = []
    body.append(...)
    body.append(...)
    body.append(...)

    body = "\n\n".join(body)

nursix: (meaning: indices in variable names indicate that this should actually be a list, but the programmer didn't know how to make a list in Python. Does that apply for you?)

static/scripts/S3/s3.gis.js
```diff
@@ -3619,6 +3619,10 @@ OpenLayers.ProxyHost = S3.Ap.concat('/gis/proxy?url=');
                         defaults[key] = '';
                     }
                     _.defaults(attributes, defaults);*/
+                    // Since this is single feature case, feature should have single id
+	            if (attributes.id.constructor === Array) {
+		        attributes.id = attributes.id[0]
```
nursix: Missing semicolon

static/scripts/S3/s3.gis.js
```diff
@@ -3730,6 +3734,10 @@ OpenLayers.ProxyHost = S3.Ap.concat('/gis/proxy?url=');
                         defaults[key] = '';
                     }
                     _.defaults(attributes, defaults);*/
+                    // Since this is single feature case, feature should have single id
+		    if (feature.attributes.id.constructor === Array) {
+		        feature.attributes.id = feature.attributes.id[0]
```
nursix: Missing semicolon

controllers/cap.py
```diff
@@ -630,36 +676,40 @@ def custom_widget_fn_3(r, **attr):
...
-                                       ),
-                                   DIV(SPAN("%s :: " % T("Incident IDs"),
-                                            _class="cap-label"
-                                            ),
-                                       SPAN(table.incidents.represent(record.incidents),
-                                            _class="cap-value"
-                                            ),
-                                       ),
                                    )
+                        fn_3_insert = fn_3.insert
```
nursix: What is the reason for indices in the variable names here. Those "fn" don't seem generic, they have a specific purpose, like: details_widget, info_widget, qualifier_widget - so why not indicate that clearly in the names. Should those functions be defined inline in the controller, or should those be decorated functions in the model?

controllers/cap.py
```diff
@@ -568,6 +607,14 @@ def custom_widget_fn_2(r, **attr):
                                                      _class="cap-value"
                                                      )
                                                 ))
+                        if info.contact:
+                            fn_2_insert(13, DIV(SPAN("%s :: " % T("Contact Info"),
```
nursix: Is there a reason for (or even a point with) the training blank in the text node of the SPAN? HTML would ignore it anyway. And is the double colon just personal style or does that have a reason (it looks odd tbh).

controllers/cap.py
```diff
@@ -630,36 +676,40 @@ def custom_widget_fn_3(r, **attr):
...
-                            fn_3.insert(7, DIV(SPAN("%s :: " % T("Handling Code"),
+                            fn_3_insert(7, DIV(SPAN("%s :: " % T("Handling Code"),
                                                     _class="cap-label"
                                                     ),
                                                SPAN(table.codes.represent(record.codes),
                                                     _class="cap-value"
                                                     ),
                                                ))
+                        if record.note:
+                            fn_3_insert(8, DIV(SPAN("%s :: " % T("Note"),
```
nursix: Inserting into a DIV isn't quite efficient - and a bit fragile in refactoring. It may be better to create the sub-DIVs as a sequence, and define the outer DIV with that sequence of components at the end of the function. Then you can append to the sequence, or even use "" for the empty components and avoid append/insert altogether.

There is also a lot (!) you could DRY here - this repeated DIV(SPAN(something, _class="cap-label"), SPAN(something_else, _class="cap-value")) is really unnecessary and makes the code hard to read.

biplovbhandari: @nursix I created a div as:

nursix: @biplovbhandari : serious question?

Try Python:

    def form_row(label, value, lclass="cap-label", vclass="cap-value"):

        return DIV(SPAN(label, _class=lclass),
                   SPAN(value, _class=vclass),
                   )

nursix: @biplovbhandari : How about the following variant:

                    widget = s3db.cap_AlertProfileWidget
                    component = widget.component

                    @widget(None)
                    def alert_widget(r, **attr):
                        return (
                          component("Headline", info.headline,
                                    uppercase = True,
                                    hide_empty = True,
                                    ),
                          component("Description",
                                    info.description,
                                    uppercase = True,
                                    hide_empty = True,
                                    ),
                          component("Response Type",
                                    " ".join(info.response_type),
                                    uppercase = True,
                                    hide_empty = True,
                                    ),
                          component("Instructions",
                                    info.instruction,
                                    uppercase = True,
                                    hide_empty = True,
                                    ),
                        )

                    @widget("Information",
                            label = "Event",
                            value = itable.event_type_id.represent(info.event_type_id),
                            )
                    def info_widget(r, **attr):
                        
                        if len(parameters):
                            params = ", ".join("%s: %s" % (p.name, p.value)
                                               for p in parameters)
                        else:
                            params = None

                        return (
                            component("Language",
                                      info.language,
                                      represent = itable.language.represent,
                                      ),
                            component("Category",
                                      info.category,
                                      represent = itable.category.represent,
                                      hide_empty = True,
                                      ),
                            component("Urgency",
                                      info.urgency,
                                      represent = itable.urgency.represent,
                                      ),
                            component("Severity",
                                      info.severity,
                                      represent = itable.severity.represent,
                                      ),
                            component("Certainty",
                                      info.certainty,
                                      represent = itable.certainty.represent,
                                      ),
                            component("Audience",
                                      info.audience,
                                      hide_empty=True,
                                      ),
                            component("Effective Date",
                                      info.effective,
                                      represent = itable.effective.represent,
                                      ),
                            component("Onset Date",
                                      info.onset,
                                      represent = itable.onset.represent,
                                      ),
                            component("Expiry Date",
                                      info.expires,
                                      represent = itable.expires.represent,
                                      ),
                            component("Information URL",
                                      info.web,
                                      ),
                            component("Sender",
                                      info.sender_name,
                                      hide_empty=True,
                                      ),
                            component("Contact Info",
                                      info.contact,
                                      ),
                            component("Parameters",
                                      parameters,
                                      hide_empty=True,
                                      ),
                        )

                    @widget("Alert Qualifiers")
                    def qualifiers_widget(r, **attr):

                        return (
                            component("Sender ID",
                                      record.sender,
                                      ),
                            component("Sent Date/Time",
                                      record.sent,
                                      represent = table.sent.represent,
                                      ),
                            component("Message Status",
                                      record.status,
                                      represent = table.status.represent,
                                      ),
                            component("Message Type",
                                      record.msg_type,
                                      ),
                            component("Handling Code",
                                      record.codes,
                                      represent = table.codes.represent,
                                      hide_empty = True,
                                      ),
                            component("Scope",
                                      record.scope,
                                      ),
                            component("Note",
                                      record.note,
                                      ),
                            component("Reference ID",
                                      record.reference,
                                      ),
                            component("Incident IDs",
                                      record.incidents,
                                      represent = table.incidents.represent,
                                      ),
                        )

                    s3db.configure(tablename,
                                   profile_header = profile_header,
                                   profile_layers = (layer,),
                                   profile_widgets = ({"type": "custom",
                                                       "fn": alert_widget,
                                                       },
                                                      map_widget,
                                                      {"type": "custom",
                                                       "fn": info_widget,
                                                       },
                                                      {"type": "custom",
                                                       "fn": qualifiers_widget,
                                                       },
                                                      ),
                                   )

Does this look ok?

Here is the DRY part of it (in modules/s3db/cap.py then):
class cap_AlertProfileWidget(object):
    """ Custom profile widget builder """

    def __init__(self, title, label=None, value=None):

        self.title = title
        self.label = label
        self.value = value

    def __call__(self, f):
        """ Widget builder """

        def widget(r, **attr):
            components = f(r, **attr)
            components = [c for c in components if c is not None]
            title = self.title
            if title:
                label = self.label
                value = self.value
                if label and value:
                    title_ = DIV(SPAN("%s " % T(title),
                                      _class="cap-value upper"
                                      ),
                                 SPAN("%s: " % T(label),
                                      _class="cap-label upper"
                                      ),
                                 SPAN(value,
                                      _class="cap-strong"
                                      ),
                                 )
                else:
                    title_ = DIV(SPAN(T(title)))
                
                header = TAG[""](title_,
                                 DIV(_class="underline"),
                                 )
                output = DIV(header, *components)
            else:
                output = DIV(components)
            return output

        return widget

    @staticmethod
    def component(label, value,
                  represent = None,
                  uppercase = False,
                  strong = False,
                  hide_empty = False):
        """ Component builder """
                      
        if not value and hide_empty:
            return None
        else:
            if represent:
                value = represent(value)
            label_class = "cap-label"
            if uppercase:
                label_class = "%s upper" % label_class
            value_class = "cap-strong" if strong else "cap-value"
            output = DIV(SPAN("%s: " % T(label), _class=label_class),
                            SPAN(value, _class=value_class),
                            )
            return output

Clean separation of logic and layout, readable code.

nursix: Yeah, of course - just quickly hacked together. Lacks proper docstrings, and some details may need to be adjusted, but it illustrates the idea.

With this widget configuration pattern, you can quickly add/extend/refactor the profile widgets - and even make this customisable (both contents and layout) - and the component order doesn't break when you add, move or remove components (much more robust than insert).

And I think, there's also potential for generalisation of the pattern so it can be re-used in other profiles - but ok, that might go too far right now ;)

PR #1295, opened 2016-06-15
SAMBRO: show profile action button, admin menu fix,apply the filter in the datalist for ajax, allow the imported alerts to be relayed, Alert hub, Implement facebook post, Profile page refactor + tweaks

biplovbhandari:
@flavour Please MERGE

### DIFF
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -39,6 +39,7 @@ __all__ = ("S3CAPModel",
            "cap_AssignArea",
            "cap_AreaRepresent",
            "cap_CloneAlert",
+           "cap_AlertProfileWidget",
            #"cap_gis_location_xml_post_parse",
            #"cap_gis_location_xml_post_render",
            )
@@ -279,6 +280,10 @@ cap_info_certainty_opts = OrderedDict([
     ("Unknown", T("Certainty unknown")),
 ])
 
+# CAP info language complete list
+represent_languages = IS_ISO639_2_LANGUAGE_CODE.language_codes()
+represent_languages.append(("en-US", "English"))
+
 # =============================================================================
 class S3CAPModel(S3Model):
     """
@@ -499,6 +504,12 @@ $.filterOptionsS3({
                                  readable = False,
                                  writable = False,
                                  ),
+                     # To separate the external CAP Alert
+                     Field("external", "boolean",
+                           default = True,
+                           readable = False,
+                           writable = False,
+                           ),
                      *s3_meta_fields())
 
         list_fields = ["info.event_type_id",
@@ -572,6 +583,7 @@ $.filterOptionsS3({
                   context = {"location": "location.location_id",
                              },
                   create_onaccept = self.cap_alert_create_onaccept,
+                  deduplicate = S3Duplicate(primary=("identifier")),
                   filter_widgets = filter_widgets,
                   list_fields = list_fields,
                   list_layout = cap_alert_list_layout,
@@ -787,11 +799,8 @@ $.filterOptionsS3({
                      Field("language",
                            default = "en-US",
                            label = T("Language"),
-                           represent = lambda opt: languages.get(opt,
-                                                                 UNKNOWN_OPT),
-                           requires = IS_EMPTY_OR(
-                                        IS_IN_SET(languages)
-                                        ),
+                           represent = S3Represent(options = dict(represent_languages)),
+                           requires = IS_IN_SET(languages),
                            comment = DIV(_class="tooltip",
                                          _title="%s|%s" % (T("Denotes the language of the information"),
                                                            T("Code Values: Natural language identifier per [RFC 3066]. If not present, an implicit default value of 'en-US' will be assumed. Edit settings.cap.languages in 000_config.py to add more languages. See <a href=\"%s\">here</a> for a full list.") % "http://www.i18nguy.com/unicode/language-identifiers.html")),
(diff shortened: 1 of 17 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -1019,4 +1119,120 @@ def get_sms_content(row):
...
+
+        if row["cap_info.contact"]:
+            body7 = T("For more details visit %(url)s or contact %(contact)s") % \
+                {"url": s3_str(row["cap_info.web"]),
+                 "contact": s3_str(row["cap_info.contact"]),
+                 }
+        else:
+            body7 = T("For more details visit %(url)s") % \
+                {"url": s3_str(row["cap_info.web"])}
+        body_append(body7)
```
nursix: You can't let go of the index in the variable name, can you? 

body1, body2, body3...why not just "item" and re-use that, or why have a local variable at all?

Could just make this a list (rather than appending):

nursix: (Example to illustrate the pattern - don't use verbatim):

        body = [
            T("%(scope)s %(status)s Alert") % \
                {"scope": s3_str(row["cap_alert.scope"]),
                 "status": s3_str(row["cap_alert.status"]),
                 },
            T((s3_str(row["cap_info.headline"]))),
            T("ID: %(identifier)s") % \
                {"identifier": s3_str(row["cap_alert.identifier"]),
                 },
            T("""%(priority)s message %(message_type)s in effect for %(area_description)s""") % \
                {"priority": s3_str(priority),
                 "message_type": s3_str(row["cap_alert.msg_type"]),
                 "area_description": s3_str(row["cap_area.name"]),
                 },
            T("This %(severity)s %(event_type)s is %(urgency)s and is %(certainty)s") % \
                {"severity": s3_str(row["cap_info.severity"]),
                 "event_type": s3_str(event_type),
                 "urgency": s3_str(row["cap_info.urgency"]),
                 "certainty": s3_str(row["cap_info.certainty"]),
                 },

            T("""Message %(identifier)s: %(event_type)s (%(category)s) issued by %(sender_name)s sent at %(date)s from %(source)s""") % \
                {"identifier": s3_str(row["cap_alert.identifier"]),
                 "event_type": s3_str(event_type),
                 "category": s3_str(category),
                 "sender_name": s3_str(row["cap_info.sender_name"]),
                 "date": s3_str(row["cap_alert.sent"]),
                 "source": s3_str(row["cap_alert.source"]),
                 },
            # and so forth...
        ]

        # str.join not working with T's
        body = "\n\n".join(s3_str(item) for item in body)

biplovbhandari:
@flavour please merge!

modules/s3db/cap.py
```diff
@@ -499,6 +504,12 @@ def model(self):
                                  readable = False,
                                  writable = False,
                                  ),
+                     # To separate the external CAP Alert
+                     Field("external", "boolean",
+                           default = True,
```
flavour: Nice to add a comment explaining why this default & poitning to the setting of True in interactive
(It looks odd seeing it here naked)

modules/s3db/cap.py
```diff
@@ -787,11 +799,8 @@ def model(self):
                      Field("language",
                            default = "en-US",
                            label = T("Language"),
-                           represent = lambda opt: languages.get(opt,
-                                                                 UNKNOWN_OPT),
-                           requires = IS_EMPTY_OR(
-                                        IS_IN_SET(languages)
-                                        ),
+                           represent = S3Represent(options = dict(represent_languages)),
+                           requires = IS_IN_SET(languages),
```
flavour: This isn't i18n like this (i.e. Strings will always be English, not translated)
I see that you can't use the validator as-is since you need the extra option for en-US.
I would propose adding an option to the validator to pass in extra options & then use that as-is

modules/s3db/cap.py
```diff
@@ -787,11 +799,8 @@ def model(self):
...
                            comment = DIV(_class="tooltip",
```
flavour: This comment doesn't match the core model definition.
If the requires is modified to a subset for interactive records in the controller then this label should be added there too (which also improves performance a fraction ;) )

modules/s3db/cap.py
```diff
@@ -1809,6 +1823,14 @@ def cap_info_onvalidation(form):
                 form.errors["category"] = \
                     current.T("Atleast one category is required.")
```
flavour: At least (missing space)

modules/s3db/cap.py
```diff
@@ -2388,11 +2410,8 @@ def model(self):
...
                      Field("language",
                            label = T("Language"),
-                           represent = lambda opt: languages.get(opt,
-                                                                 UNKNOWN_OPT),
-                           requires = IS_EMPTY_OR(
-                                        IS_IN_SET(languages)
-                                        ),
+                           represent = S3Represent(options = dict(represent_languages)),
+                           requires = IS_IN_SET(languages),
                            comment = DIV(_class="tooltip",
```
flavour: As-above

modules/s3db/cap.py
```diff
@@ -4515,4 +4551,92 @@ def represent_row(self, row):
...
+        def widget(r, **attr):
+            components = f(r, **attr)
+            components = [c for c in components if c is not None]
+            title = self.title
+            if title:
+                label = self.label
+                value = self.value
+                if label and value:
+                    title_ = DIV(SPAN("%s " % T(title),
+                                      _class="cap-value upper"
```
flavour: semantically this should be cap-title

flavour:
I will merge even with the little cleanups left to do

PR #1296, opened 2016-06-18
SAMBRO:fix +  Allow Basic Authentication for CAP RSS 

biplovbhandari:
@flavour Please merge


modules/s3cfg.py
```diff
@@ -2127,6 +2127,12 @@ def get_msg_basestation_code_unique(self):
         """
         return self.msg.get("basestation_code_unique", False)
 
+    def get_msg_rss_use_links(self):
+        """
+            Whether to use links of entry element if link fail
```
flavour: No, this isn't what I was suggesting to make configurable...it was the following of links at all.
This is what seems usecase-specific.
That is what should either be a deployment_setting (but then I'm not sure what the consistent pattern should be for handling such returned data?) or else left to the parser (seems more correct to me)

controllers/cap.py
```diff
@@ -632,6 +641,13 @@ def qualifiers_widget(r, **attr):
                                               ),)
 
             elif r.component_name == "info":
+                # Filter the langauge options
+                itable.language.requires = IS_ISO639_2_LANGUAGE_CODE(\
+                                            zero=None,
+                                            select=settings.get_cap_languages())
```
flavour: Currently this is the same as the default model setting, just without translate=True
I'd have thought you defintiely wanted translate=True here.
Maybe the model should remove the 'select' parameter? (Otherwise the comment is valid there too)

modules/s3db/msg.py
```diff
@@ -1326,6 +1326,20 @@ def model(self):
                            readable = False,
                            writable = False,
                            ),
+                     Field("username",
+                           label = T("Username"),
+                           readable = False,
```
flavour: I don't think the username needs to be readable False?

modules/s3db/msg.py
```diff
@@ -1326,6 +1326,20 @@ def model(self):
...
+                           comment = DIV(_class="tooltip",
+                                         _title="%s|%s" % (T("Username"),
+                                                           T("Optional username for HTTP Basic Authentication."))),
+                           ),
+                     Field("password", "password",
+                           label = T("Password"),
+                           readable = False,
```
flavour: Seems that we could use some stdisation of this...other settings are inconsistent here.
I think readable=False is probably correct and also enhanced with:
widget = S3PasswordWidget()

biplovbhandari:
@flavour corrected! Please merge

modules/s3cfg.py
```diff
@@ -2456,6 +2456,12 @@ def get_cap_post_to_facebook(self):
 
         return self.cap.get("post_to_facebook", False)
 
+    def get_cap_rss_use_links(self):
+        """
+            Whether to use links of entry element if link fail
+        """
+        return self.cap.get("rss_use_links", False)
```
flavour: This does nothing any more?

modules/templates/SAMBRO/parser.py
```diff
@@ -306,20 +306,124 @@ def parse_rss_2_cap(message):
...
-            resource = s3db.resource("cap_alert")
-            stylesheet = os.path.join(current.request.folder, "static", "formats", "cap", "import.xsl")
-            success = resource.import_xml(File, stylesheet=stylesheet)
+                        # Import via XSLT
+                        resource = s3db.resource("cap_alert")
+                        stylesheet = os.path.join(current.request.folder, "static", "formats", "cap", "import.xsl")
+                        success = resource.import_xml(File, stylesheet=stylesheet)
+                        return
+                else:
+                    if current.deployment_settings.get_cap_rss_use_links():
```
flavour: Ah, well used here but is there any point?
When would a SAMBRO deployment not want this option?
The Parser is already template-specific

biplovbhandari:
@flavour How can I not apply the gis configuration for the imported alert area because for country instance it will give validation error? Or apply default one?
Also can you please minify the js files? Many thanks!

flavour:
Minifying now

If you need an instance to support Alert Areas outside of the country then don't have any restricted boundary in the default config for that instance.

biplovbhandari:
@flavour What is the proper way to change the length of the column event_event_type.name (https://github.com/sahana/eden/blob/master/modules/s3db/event.py#L106)? This is required for the imported alerts as some event name are longer than that. If I try to change that value to other value, it generates a null value error (error ticket: http://pastebin.com/QJkz4pz0)

PR #1299, opened 2016-06-21
SAMBRO: tweaks

biplovbhandari:
@flavour Please merge

### DIFF
--- a/static/formats/cap/export.xsl
+++ b/static/formats/cap/export.xsl
@@ -412,7 +412,7 @@
                 </xsl:variable>
                 <circle>
                     <!-- Convert radius from m to km -->
-                    <xsl:value-of select="concat($lat, ',', $lon, ' ', $radius * 1000)"/>
+                    <xsl:value-of select="concat($lat, ',', $lon, ' ', $radius * 0.001)"/>
                 </circle>
             </xsl:when>
             <xsl:when test="starts-with($wkt,'&#34;POLYGON')">
--- a/static/formats/cap/import.xsl
+++ b/static/formats/cap/import.xsl
@@ -94,11 +94,12 @@
                     <xsl:value-of select="cap:restriction" />
                 </data>
             </xsl:if>
-            <xsl:if test="cap:addresses!=''">
-                <data field="addresses"> <!-- further in python code -->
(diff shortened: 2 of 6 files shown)


static/formats/cap/export.xsl
```diff
@@ -411,8 +411,7 @@
                     <xsl:value-of select="data[@field='lon']" />
                 </xsl:variable>
                 <circle>
-                    <!-- Convert radius from m to km -->
```
flavour: Why was this removed?
has the CAP std changed?
Internally Sahana GIS uses m, my understanding was that CAP used km, hence the need to multiply by 1000

static/formats/cap/import.xsl
```diff
@@ -457,8 +462,7 @@
                     </data>
                     
                     <data field="radius">
-                        <!-- Radius comes in as km, so convert to m -->
```
flavour: As-above

static/formats/rss/modules/cap.xsl
```diff
@@ -8,7 +8,7 @@
     <!-- cap_info -->
     <!-- @ToDo: Handle multiple info -->
     <xsl:template match="resource[@name='cap_alert']" mode="contents">
-        <xsl:if test="./data[@field='is_template' and @value='false']">
+        <xsl:if test="./data[@field='is_template' and @value='false'] and ./data[@field='external' and @value='false']">
```
flavour: Should this be OR not AND?

static/formats/cap/import.xsl
```diff
@@ -436,7 +441,7 @@
         <resource name="cap_area_location">
             <reference field="location_id" resource="gis_location">
                 <resource name="gis_location">
-                    <data field="gis_feature_type" value="1">Point</data>
+                    <data field="gis_feature_type" value="3">Polygon</data>
```
flavour: Why change to a Polygon here?
As I see it we are importing Lat/Lon (& maybe Radius) hence these are Points.
Currently we are not importing Polygons
Sure we should probably enhance to import Polygons where-defined, but just changing this to Polygon doesn't achieve that...

biplovbhandari:
@flavour Done...Please merge

PR #1303, opened 2016-06-26
SAMBRO: notify all approver even if logged in as approver

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -1265,8 +1265,9 @@ def notify_approver():
                 for user_id in user_ids:
                     pe_append(user_pe_id(int(user_id)))
                 subject = "%s: Alert Approval Required" % settings.get_system_name_short()
+                review_url = URL(c="cap", f="alert", args=[alert_id, "review"])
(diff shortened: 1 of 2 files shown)


flavour:
This appears wrong, but must have come from somewhere.
Why would we notify all approvers whenever anyone reviews an alert?
This has the potential for many copies of a 'please review x alert' message.
I've not investigated where it should happen instead, but think it really should have been an earlier stage which sent the notifications...there should never be a need to send notifications when this button is pressed?

biplovbhandari:
@flavour This is the requirement from Myanmar. Even though the Approver can approve the alert without consent with others, they want to know that someone is approving the alert.

flavour:
I am not arguing the requirement.
I am arguing the proposed workflow to meet this.
(1) It doesn't seem to actually do what Myanmar want:
- The solution as-proposed send them a notification to reviw an alert when an alert is reviewed
- What they seem to want is a notification saying that an Alert has been approved (onapprove hook)
(2) There is a danger that users will be flooded with multiple copies of a 'Please review x alert' message

PR #1310, opened 2016-07-04
SAMBRO: handle multiple info in profile + SAMBRO:tweaks

biplovbhandari:
@flavour Please merge..

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -388,38 +388,41 @@ def alert():
                     alert_id = record.id
 
                     # Info
-                    # @ToDo: handle multiple languages
-                    info = db(itable.alert_id == alert_id).select(itable.id,
-                                                                  itable.language,
-                                                                  itable.category,
-                                                                  itable.event_type_id,
-                                                                  itable.response_type,
-                                                                  itable.urgency,
-                                                                  itable.severity,
-                                                                  itable.certainty,
-                                                                  itable.audience,
-                                                                  itable.effective,
-                                                                  itable.onset,
-                                                                  itable.expires,
-                                                                  itable.sender_name,
-                                                                  itable.headline,
-                                                                  itable.description,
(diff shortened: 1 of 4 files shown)


controllers/cap.py
```diff
@@ -388,38 +388,44 @@ def prep(r):
...
+                            query = (ptable.info_id == info["cap_info.id"])
+                        parameters = db(query).select(ptable.name,
+                                                      ptable.value)
 
                     # Area
-                    # @ToDo: handle multiple areas
-                    atable = s3db.cap_area
-                    area = db(atable.alert_id == alert_id).select(atable.name,
-                                                                  limitby=(0, 1)).first()
+                    data = resource.select(["area.name"], raw_data=True)
```
flavour: Why do a separate select here? Why not just add this field to the 1st one?

controllers/cap.py
```diff
@@ -662,7 +667,7 @@ def qualifiers_widget(r, **attr):
                 # Check for prepopulate
                 if alert_id:
                     itable.web.default = settings.get_base_public_url()+\
-                                         URL(c="cap", f="alert", args=alert_id)
+                                         URL(c="cap", f="public", args=alert_id)
```
flavour: Is this true for all Alerts or do we need to distinguish?

modules/s3db/cap.py
```diff
@@ -1759,7 +1760,7 @@ def cap_info_onaccept(form):
             #    set_.update(parameter = fstring)
 
             web = "%s%s" % (current.deployment_settings.get_base_public_url(),
-                            URL(c="cap", f="alert", args=[alert_id]))
+                            URL(c="cap", f="public", args=[alert_id]))
```
flavour: Same question

biplovbhandari:
@flavour Corrected please merge

PR #1311, opened 2016-07-05
SAMBRO: edit approval message + SAMBRO: fix translation problem + SAMBRO: tweaks

biplovbhandari:
@flavour Please merge

### DIFF
--- a/controllers/cap.py
+++ b/controllers/cap.py
@@ -814,6 +814,9 @@ def alert():
                 rows = db(itable.alert_id == alert.template_id).select(*fields)
                 for row in rows:
                     row_clone = row.as_dict()
+                    audience = row_clone["audience"]
+                    if not audience or audience == messages["None"]:
+                        audience = None
                     del row_clone["id"]
                     row_clone["alert_id"] = lastid
                     row_clone["template_info_id"] = row.id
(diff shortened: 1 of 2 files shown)


modules/s3db/cap.py
```diff
@@ -59,7 +59,6 @@
 from gluon.tools import fetch
 from ..s3 import *
 
-T = current.T
```
flavour: Good, but you'll need to do a lot more, no?
All the cap_incident_type_opts justunder this refer to T which will no longer be available...these will need moving into a class...can be a small class with just these to minimise need for loading table models unnecessarily.

controllers/cap.py
```diff
@@ -1287,7 +1299,16 @@ def notify_approver():
                 subject = "%s: Alert Approval Required" % settings.get_system_name_short()
                 url = "%s%s" % (settings.get_base_public_url(),
                                 URL(c="cap", f="alert", args=[alert_id, "review"]))
-                message = "You are requested to take action on this alert:\n\n%s" % url
+                message = """
+Hello Editor,
+%(first_name)s %(last_name)s has created the alert message.
```
flavour: Much nicer...although I would suggest using s3_fullname if possible...this supports Middle Names too (& we can adjust word order based on locale in future)

modules/s3db/cap.py
nursix: Or, you simply remove the T's in these option dicts, and instead add translate=True to the respective S3Represent instances.

flavour: Also used in S3OptionsFilter options

nursix: Ah...doesn't really hurt to add a translate-option to the S3OptionsFilter with fixed options dicts as well, does it? Maybe I should do that...

PR #1312, opened 2016-07-05
SAMBRO: edit approval message + SAMBRO: fix translation problem + SAMBRO: tweaks

biplovbhandari:
@flavour corrected...Please merge

### DIFF
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -27,7 +27,16 @@
     OTHER DEALINGS IN THE SOFTWARE.
 """
 
-__all__ = ("S3CAPModel",
+__all__ = ("get_cap_incident_type_opts",
+           "get_cap_alert_status_code_opts",
+           "get_cap_alert_msgType_code_opts",
+           "get_cap_alert_scope_code_opts",
+           "get_cap_info_category_opts",
+           "get_cap_info_responseType_opts",
+           "get_cap_info_urgency_opts",
+           "get_cap_info_severity_opts",
+           "get_cap_info_certainty_opts",
+           "S3CAPModel",
            "S3CAPHistoryModel",
            "S3CAPAreaNameModel",
            "cap_alert_is_template",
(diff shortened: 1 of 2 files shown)


modules/s3db/cap.py
```diff
@@ -807,10 +869,10 @@ def model(self):
                      Field("category", "list:string", # 1 or more allowed
                            label = T("Category"),
                            required = IS_NOT_EMPTY(),
-                           represent = S3Represent(options = cap_info_category_opts,
+                           represent = S3Represent(options = get_cap_info_category_opts(),
                                                    multiple = True,
                                                    ),
-                           requires = IS_IN_SET(cap_info_category_opts,
+                           requires = IS_IN_SET(get_cap_info_category_opts(),
```
nursix: Calling a function twice that delivers an invariant result? Probably better to call it once and assign to a local variable.

modules/s3db/cap.py
```diff
@@ -842,10 +904,10 @@ def model(self):
                      ),
                      Field("response_type", "list:string", # 0 or more allowed
                            label = T("Response Type"),
-                           represent = S3Represent(options = cap_info_responseType_opts,
+                           represent = S3Represent(options = get_cap_info_responseType_opts(),
```
nursix: NB slight violation of naming convention (camel case) - maybe take the opportunity to fix that?

modules/s3db/cap.py
```diff
@@ -2470,9 +2543,10 @@ def model(self):
                            ),
                      Field("severity",
                            label = T("Severity"),
-                           represent = lambda opt: \
-                            cap_info_severity_opts.get(opt, UNKNOWN_OPT),
-                           requires = IS_IN_SET(cap_info_severity_opts),
+                           represent = S3Represent(options = get_cap_info_severity_opts(),
+                                                   translate = True,
```
nursix: Now that translate=True doesn't make sense as you have T's in the options dict. Either T there or translate here.

modules/s3db/cap.py
```diff
@@ -4637,10 +4713,18 @@ def component(label, value,
             return None
         else:
             if isinstance(value, list):
-                value = ", ".join(value)
-
-            if represent:
-                value = represent(value)
+                for i, value_ in enumerate(value):
+                    if not value_:
+                        value.remove(value_)
```
nursix: Changing the iteratable inside the loop => this can crash. Safer to build a new list, or skip here and add an iterator into the join().

Should also update the docstring with the represent-option.

modules/s3db/cap.py
```diff
@@ -4637,10 +4713,18 @@ def component(label, value,
...
+                    elif represent:
+                        value[i] = represent(value_)
```
nursix: 1:1 replacement is ok, just not remove.

modules/s3db/cap.py
```diff
@@ -27,7 +27,16 @@
     OTHER DEALINGS IN THE SOFTWARE.
 """
 
-__all__ = ("S3CAPModel",
+__all__ = ("get_cap_incident_type_opts",
```
flavour: I was imagining a single fn rather than all these => called 1/model => less overheads

flavour: Although with the new enhancement Dominic added to S3OptionsFilter, then no need for fns...can do the T() inside the Represent & Filter instead...so can define plain strings at the module level

modules/s3db/cap.py
```diff
@@ -2470,9 +2543,10 @@ def model(self):
                            ),
                      Field("severity",
                            label = T("Severity"),
-                           represent = lambda opt: \
-                            cap_info_severity_opts.get(opt, UNKNOWN_OPT),
-                           requires = IS_IN_SET(cap_info_severity_opts),
+                           represent = S3Represent(options = get_cap_info_severity_opts(),
+                                                   translate = True,
+                                                   ),
+                           requires = IS_IN_SET(get_cap_info_severity_opts()),
```
flavour: Yikes, you even call the fn twice ;)

modules/s3db/cap.py
flavour: The case which doesn't have a translate=True option is requires=IS_IN_SET
But can do the T() part before that inside the model (then may as well re-use for the S3Represent/S3OptionsFIlter)

PR #1314, opened 2016-07-07
SAMBRO: message acknowledgement link for reponders + SAMBRO: use update policy for notification + SAMBRO: tweaks

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/config.py
+++ b/modules/templates/SAMBRO/config.py
@@ -7,6 +7,7 @@ from collections import OrderedDict
 from gluon import current
 from gluon.html import *
 from gluon.storage import Storage
+from gluon.languages import lazyT
 
 from s3 import FS, s3_str
 
@@ -89,6 +90,9 @@ def config(settings):
 
     # ALlow RSS to use links of entry if link fails
     settings.cap.rss_use_links = True
+
+    # SAMBRO supports ack workflow
+    settings.cap.use_ack = True
 
     # -------------------------------------------------------------------------
     # L10n (Localization) settings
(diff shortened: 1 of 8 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -398,84 +387,90 @@ def onapprove(record):
...
-                        send_by_pe_id(row_.pe_id, subject, email_content)
-                        try:
-                            send_by_pe_id(row_.pe_id, subject, sms_content, contact_method="SMS")
-                        except ValueError:
-                            current.log.error("No SMS Handler defined!")
+                    query = query_ & (gtable.id.belongs(addresses))
+                rows = db(query).select(ptable.pe_id)
+                subject = get_email_subject(arow, non_system=True)
+                cap_use_ack = settings.get_cap_use_ack()
+                html_content = get_html_email_content(arow, non_system=True)
```
flavour: No need to run this if cap_use_ack is False

modules/templates/SAMBRO/config.py
```diff
@@ -398,84 +387,90 @@ def onapprove(record):
...
+                        ack_id = create_ack(alert_id, get_user_id(pe_id=row.pe_id))
+                        email_content = "%s%s%s" % ("<html>",
+                                                    XML(get_html_email_content(arow,
+                                                                    ack_id=ack_id,
+                                                                    non_system=True)),
+                                                    "</html>")
+                        sms_content = get_sms_content(arow, ack_id=ack_id, non_system=True)
+                    else:
+                        email_content = "%s%s%s" % ("<html>", XML(html_content), "</html>")
+                        sms_content = get_sms_content(arow, non_system=True)
```
flavour: if cap_use_ack is False then do this once outside the loop

modules/templates/SAMBRO/config.py
```diff
@@ -959,144 +943,159 @@ def pr_subscription_filter_row_options():
...
 
     # -------------------------------------------------------------------------
-    def get_html_email_content(row):
+    def get_html_email_content(row, ack_id=None, non_system=False):
         """
             prepare the content for html email
+
+            @param row: the row from which the email will be constructed
+            @param ack_id: cap_alert_ack.id for including the acknowledgement link
+            @param non_system: boolean to distinguish the notification email
```
flavour: distinguish from?

flavour: In general better to avoid double-negatives like non_system=False, system=True would seem more intuitive...but I don't understand the 2 use-cases yet

biplovbhandari: from the first responders email. because row coming from .msg is already represented and formatted if it is a list.

PR #1317, opened 2016-07-25
GIS: key vars for loading google maps api + SAMBRO: show new created alert via datalist ajax + SAMBRO: enhancements and bugfix

biplovbhandari:
@flavour Please Merge


biplovbhandari:
@flavour Please merge

static/formats/rss/base.xsl
```diff
@@ -95,13 +74,11 @@
     <xsl:template name="channel_header">
 
         <xsl:param name="resource_url"/>
-        <title>
-            <xsl:value-of select="$title"/>
-        </title>
+        <title>RSS-Feed from Sahana-Eden</title>
```
nursix: Ahm, no - not hardcoding this here.

static/formats/rss/base.xsl
```diff
@@ -95,13 +74,11 @@
...
         <link>
             <xsl:value-of select="$resource_url"/>
         </link>
-        <description></description>
+        <description>RSS-Feed from Sahana-Eden</description>
```
nursix: Same - can't hardcode this here (base is generic)

static/formats/rss/base.xsl
```diff
@@ -181,7 +158,7 @@
             </xsl:choose>
         </xsl:variable>
         <xsl:variable name="time"
-                      select="substring-before(substring-after($iso_datetime, 'T'), 'Z')"/>
+                      select="substring-after($iso_datetime, 'T')"/>
```
nursix: Should retain a fallback for the case that the iso_datetime has a Z (which can happen if so configured)

static/formats/rss/base.xsl
```diff
@@ -54,35 +54,14 @@
...
-        <xsl:param name="resource_url"/>
-        <xsl:variable name="resource_name">
-            <xsl:call-template name="resource_name"/>
-        </xsl:variable>
-        <item>
-            <link>
-                <xsl:value-of select="concat($resource_url, '?', $resource_name, '.uid=', @uuid)"/>
-            </link>
-            <pubDate>
-                <xsl:if test="@modified_on">
```
nursix: Nope - don't remove that, there was a point to it. 
If you need an exception, then implement an exception based on resource_name - but don't change all other cases.

Also - not really a good idea to un-DRY this, the <item> wrapper should be retained in base.

static/formats/rss/base.xsl
```diff
@@ -54,35 +54,14 @@
                         </xsl:otherwise>
                     </xsl:choose>
                 </xsl:variable>
-                <xsl:value-of select="$target"/>
-                <xsl:apply-templates select=".//resource[@name=$target]" mode="item">
-                    <xsl:sort select="@modified_on" order="descending"/>
```
nursix: Required for all other resource types - pls don't remove this. If you need an exception, implement a xsl:choose.

PR #1323, opened 2016-09-04
SAMBRO: fix CAP RSS Feed import

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/s3/s3msg.py
+++ b/modules/s3/s3msg.py
@@ -1932,23 +1932,10 @@ class S3Msg(object):
 
     # -------------------------------------------------------------------------
     @staticmethod
-    def poll_rss(channel_id):
+    def feed_parser(channel):
         """
-            Fetches all new messages from a subscribed RSS Feed
+           Helper method to parse the (non)-authenticated feeds
         """
-
-        db = current.db
-        s3db = current.s3db
-        table = s3db.msg_rss_channel
-        query = (table.channel_id == channel_id)
-        channel = db(query).select(table.date,
-                                   table.etag,
-                                   table.url,
-                                   table.username,
-                                   table.password,
-                                   limitby=(0, 1)).first()
-        if not channel:
-            return "No Such RSS Channel: %s" % channel_id
 
         # http://pythonhosted.org/feedparser
         import feedparser
@@ -1978,6 +1965,30 @@ class S3Msg(object):
                                         status=d.bozo_exception.message,
                                         period=(300, 3600))
             return
+
+        return d
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def poll_rss(channel_id):
+        """
+            Fetches all new messages from a subscribed RSS Feed
+        """
+
+        db = current.db
+        s3db = current.s3db
+        table = s3db.msg_rss_channel
+        query = (table.channel_id == channel_id)
+        channel = db(query).select(table.date,
+                                   table.etag,
+                                   table.url,
+                                   table.username,
+                                   table.password,
+                                   limitby=(0, 1)).first()
+        if not channel:
+            return "No Such RSS Channel: %s" % channel_id
+
+        d = S3Msg.feed_parser(channel)
 
         # Update ETag/Last-polled
         now = current.request.utcnow
--- a/modules/templates/SAMBRO/parser.py
+++ b/modules/templates/SAMBRO/parser.py
@@ -241,19 +241,30 @@ class S3Parser(object):
         db = current.db
         s3db = current.s3db
         table = s3db.msg_rss
-        record = db(table.message_id == message.message_id).select(table.channel_id,
-                                                                   table.title,
-                                                                   table.from_address,
-                                                                   table.body,
-                                                                   table.date,
-                                                                   table.location_id,
-                                                                   table.author,
-                                                                   limitby=(0, 1)
-                                                                   ).first()
+        message_id = message.message_id
+        record = db(table.message_id == message_id).select(table.channel_id,
+                                                           table.title,
+                                                           table.from_address,
+                                                           table.body,
+                                                           table.date,
+                                                           table.location_id,
+                                                           table.author,
+                                                           limitby=(0, 1)
+                                                           ).first()
         if not record:
             return
 
-        channel_id = record.channel_id
+        pstable = s3db.msg_parsing_status
+        # not adding (pstable.channel_id == record.channel_id) to query
+        # because two channels (http://host.domain/eden/cap/public.rss and
+        # (http://host.domain/eden/cap/alert.rss) may contain common url
+        # eg. http://host.domain/eden/cap/public/xx.cap
+        pquery = (pstable.message_id == message_id)
+        prows = db(pquery).select(pstable.id, pstable.is_parsed)
+        for prow in prows:
+            if prow.is_parsed:
+                return
+
         alert_table = s3db.cap_alert
         info_table = s3db.cap_info
 
@@ -308,7 +319,7 @@ class S3Parser(object):
                 file = fetch(url)
             except urllib2.HTTPError, e:
                 rss_table = s3db.msg_rss_channel
-                query = (rss_table.channel_id == message.channel_id)
+                query = (rss_table.channel_id == record.channel_id)
                 channel = db(query).select(rss_table.date,
                                            rss_table.etag,
                                            rss_table.url,
@@ -317,7 +328,7 @@ class S3Parser(object):
                                            limitby=(0, 1)).first()
                 username = channel.username
                 password = channel.password
-                if e.code == 401 and username and password:
+                if username and password:
                     import base64
                     request = urllib2.Request(url)
                     base64string = base64.encodestring("%s:%s" % (username, password))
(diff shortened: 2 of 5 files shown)


modules/s3/s3msg.py
```diff
@@ -1952,23 +1952,29 @@ def poll_rss(channel_id):
...
-        if channel.etag:
-            # http://pythonhosted.org/feedparser/http-etag.html
-            # NB This won't help for a server like Drupal 7 set to not allow caching & hence generating a new ETag/Last Modified each request!
-            d = feedparser.parse(channel.url, etag=channel.etag, handlers=[auth_handler] if auth_handler else None)
-        elif channel.date:
-            d = feedparser.parse(channel.url, modified=channel.date.utctimetuple(), handlers=[auth_handler] if auth_handler else None)
+            # Basic Authentication
+            #@ToDo: handle https news feed when required
+            url = channel.url.strip("http://")
+            pass_token = "http://%s:%s@%s" % (username, password, url)
```
flavour: This way of putting the username & password in isn't as good as the previous method as it means that the username/password get recorded in more proxy logs etc

modules/s3/s3msg.py
```diff
@@ -1952,23 +1952,29 @@ def poll_rss(channel_id):
...
-            auth_handler.add_password(None, channel.url, username, password)
-
-        if channel.etag:
-            # http://pythonhosted.org/feedparser/http-etag.html
-            # NB This won't help for a server like Drupal 7 set to not allow caching & hence generating a new ETag/Last Modified each request!
-            d = feedparser.parse(channel.url, etag=channel.etag, handlers=[auth_handler] if auth_handler else None)
-        elif channel.date:
-            d = feedparser.parse(channel.url, modified=channel.date.utctimetuple(), handlers=[auth_handler] if auth_handler else None)
+            # Basic Authentication
+            #@ToDo: handle https news feed when required
```
flavour: The previous method already handled this...so there may be sites which will break due to this now

modules/templates/SAMBRO/parser.py
```diff
@@ -241,19 +241,37 @@ def parse_rss_2_cap(message):
...
+        pquery = (pstable.message_id == message_id)
+        prows = db(pquery).select(pstable.id, pstable.is_parsed)
+        is_parsed = False
+        for prow in prows:
+            if prow.is_parsed:
+                is_parsed = True
+                break
+        if is_parsed:
+            for prow in prows:
+                db(pstable.id == prow.id).update(is_parsed=True)
```
flavour: This shouldn't need to be done inside the parser as it happens here:
https://github.com/sahana/eden/blob/master/modules/s3/s3msg.py#L266

What is different about this parser that that doesn't already do the job?

flavour:
I don't like the change to core RSS handler here for the 2 reasons given inline...which cases does the old system not work with?

biplovbhandari:
@flavour For http://www.dmhwarning.gov.mm/eden/cap/alert.rss, this method gives me '`bozo_exception': SAXParseException('not well-formed (invalid token)')`.

google search gives this error may be because of character-encoding. The myanmar characters are entered by the user themselves. However it works fine for public feed,  http://www.dmhwarning.gov.mm/eden/cap/public.rss, ie there is no bozo exception.

modules/templates/SAMBRO/parser.py
biplovbhandari: @flavour Some of them are not saved because of rollback if there is any error during the xml import

biplovbhandari:
@flavour Myanmar wants to have new line between English and Myanmar content. Is there a way I can join two values using newline. I am combining values in different language here: https://github.com/sahana/eden/blob/master/modules/templates/SAMBRO/config.py#L1328

`"\n".join` won't work until we print the value.

flavour:
> For http://www.dmhwarning.gov.mm/eden/cap/alert.rss, this method gives me 'bozo_exception':  SAXParseException('not well-formed (invalid token)'). google search gives this error may be because of character-encoding.

I don't have a login so can't see the original document...the public version correctly encodes the XML as utf-8. The myanmar char isn't in the username/password is it?

flavour:
> For http://www.dmhwarning.gov.mm/eden/cap/alert.rss, this method gives me 'bozo_exception': SAXParseException('not well-formed (invalid token)')

I fixed this in current trunk. The issue was that feedparser wasn't doing pre-emptive authentication & was crashing trying to parse the 401 response (which in this case was in Burmese).
I have therefore added the header manually.

biplovbhandari:
@flavour corrected...please merge

modules/templates/SAMBRO/parser.py
```diff
@@ -334,96 +345,114 @@ def parse_rss_2_cap(message):
...
-                                        # Import via XSLT
-                                        resource = s3db.resource("cap_alert")
-                                        stylesheet = os.path.join(current.request.folder, "static", "formats", "cap", "import.xsl")
-                                        success = resource.import_xml(File, stylesheet=stylesheet)
-                                break
-                        return
+                elif current.deployment_settings.get_cap_rss_use_links():
+                    # Look into links
+                    # use-case for Philippines
+                    import feedparser
```
flavour: All the code here is highly copied from s3msg.py - would be muich better to make a helper function there & then reuse that here?
DRY => Easier maintenance

biplovbhandari:
@flavour  corrected...please merge

modules/s3/s3msg.py
```diff
@@ -1972,6 +1959,30 @@ def poll_rss(channel_id):
...
+        channel = db(query).select(table.date,
+                                   table.etag,
+                                   table.url,
+                                   table.username,
+                                   table.password,
+                                   limitby=(0, 1)).first()
+        if not channel:
+            return "No Such RSS Channel: %s" % channel_id
+
+        d = current.msg.feed_parser(channel)
```
flavour: This can be S3Msg.feed_parser

modules/s3/s3msg.py
```diff
@@ -1972,6 +1959,30 @@ def poll_rss(channel_id):
...
+
         if d.bozo:
             # Something doesn't seem right
             S3Msg.update_channel_status(channel_id,
```
flavour: Why not have the bozo handline inside the helper function as well? Again it is duplicated verbatim

biplovbhandari:
@flavour corrected...please merge

biplovbhandari:
@flavour Please merge

modules/templates/SAMBRO/parser.py
```diff
@@ -334,96 +345,88 @@ def parse_rss_2_cap(message):
...
-
-                                                # Import via XSLT
-                                                resource = s3db.resource("cap_alert")
-                                                stylesheet = os.path.join(current.request.folder, "static", "formats", "cap", "import.xsl")
-                                                success = resource.import_xml(File, stylesheet=stylesheet)
-                                        else:
+                elif current.deployment_settings.get_cap_rss_use_links():
+                    d = current.msg.feed_parser(channel)
+
+                    # Update ETag/Last-polled
```
flavour: Again, why is this stuff in common not moved inside the helper function?

modules/templates/SAMBRO/parser.py
```diff
@@ -308,7 +319,7 @@ def parse_rss_2_cap(message):
                 file = fetch(url)
             except urllib2.HTTPError, e:
                 rss_table = s3db.msg_rss_channel
```
flavour: Again, this part is in common with the core...something smells wrong here to suplicate it

modules/templates/SAMBRO/parser.py
```diff
@@ -308,7 +319,7 @@ def parse_rss_2_cap(message):
                 file = fetch(url)
             except urllib2.HTTPError, e:
```
flavour: Why would we invoke RSS download when a URL cannot be fetched?

modules/templates/SAMBRO/parser.py
```diff
@@ -334,96 +345,88 @@ def parse_rss_2_cap(message):
...
-                                    except urllib2.URLError, e:
-                                        current.log.error("Getting content from link failed: %s" % e)
+                                        else:
+                                            File = StringIO(file)
+
+                                            # Import via XSLT
+                                            resource = s3db.resource("cap_alert")
+                                            stylesheet = os.path.join(current.request.folder, "static", "formats", "cap", "import.xsl")
+                                            success = resource.import_xml(File, stylesheet=stylesheet)
+                                            # Any error on database hook causes to rollback, so ensure this
```
flavour: Why do these errors come in? This still feels very wrong to workaround an issue like this...seems better to catch & isolate the issue instead

modules/templates/SAMBRO/parser.py
```diff
@@ -334,96 +345,88 @@ def parse_rss_2_cap(message):
...
+                    data = dict(date=now)
+                    etag = d.get("etag", None)
+                    if etag:
+                        data["etag"] = etag
+                    db(query).update(**data)
+
+                    entries = d.entries
+                    for entry in entries:
+                        current_link = entry.get("link", None)
+                        if current_link == record.from_address:
```
flavour: Why do we need this match to follow links?

modules/templates/SAMBRO/parser.py
```diff
@@ -334,96 +345,88 @@ def parse_rss_2_cap(message):
...
+                            links_ = entry.get("links", [])
+                            for link_ in links_:
```
flavour: In general I think that following links & handling them might be best done via a callback function...I haven't thought through exactly how that would work here but then I'm still confused by the whole flow...I don't know why you're polling an RSS feed inside a Parser at all!

modules/templates/SAMBRO/parser.py
```diff
@@ -334,96 +345,88 @@ def parse_rss_2_cap(message):
...
+                                link = link_["href"]
+                                try:
+                                    file = fetch(link)
+                                except urllib2.HTTPError, e:
+                                    if username and password:
```
flavour: Why have this only in the except clause?
Seems better to me to have:
if username/password:
 # Try download using authentication
  try:
    import base64
else:
  # Try download without authentication
   try:
     fetch

modules/templates/SAMBRO/parser.py
```diff
@@ -241,19 +241,30 @@ def parse_rss_2_cap(message):
...
+        pstable = s3db.msg_parsing_status
+        # not adding (pstable.channel_id == record.channel_id) to query
+        # because two channels (http://host.domain/eden/cap/public.rss and
+        # (http://host.domain/eden/cap/alert.rss) may contain common url
+        # eg. http://host.domain/eden/cap/public/xx.cap
+        pquery = (pstable.message_id == message_id)
+        prows = db(pquery).select(pstable.id, pstable.is_parsed)
+        for prow in prows:
+            if prow.is_parsed:
+                return
```
flavour: I still don't understand why we're looking at the parser status inside a parser?
All of that should be handled externally - we should never enter this function for a message if is_parsed is True

Related is why are you looking for multiple records here? Any individual message_id should only have a single record in the parser status table, yet you are assuming multiple & bombing out if any are done.
This seems to be why you're touching this table inside the parser as you are not trusting the external routine....so Why are there multiple records in the parser status table for a single message_id?

modules/templates/SAMBRO/parser.py
biplovbhandari: @flavour This is the case for Philippines Public Alerts http://publicalert.pagasa.dost.gov.ph/feeds/

With feedparser parsing, for a single entry, we get the value of entry['link'] as http://publicalert.pagasa.dost.gov.ph/feeds/a159aa44-96fc-4f8f-8b75-8a771b69e7dc which is not a valid cap feed but if I look at entry['links'] as <type 'list'>: [{'href': u'http://publicalert.pagasa.dost.gov.ph/output/gfa/a159aa44-96fc-4f8f-8b75-8a771b69e7dc.cap', 'type': u'application/cap+xml', 'rel': u'alternate'}], the href inside links containing the proper cap feed.

I told PH group about this, but they don't seem to care about this.

modules/templates/SAMBRO/parser.py
biplovbhandari: Actually these errors are for the feeds from SAMBRO at country where event type id are not there or the draft alerts. These feeds are before I made changes to them.

modules/templates/SAMBRO/parser.py
biplovbhandari: @flavour so as I explained above the entry['link'] will be http://publicalert.pagasa.dost.gov.ph/feeds/a159aa44-96fc-4f8f-8b75-8a771b69e7dc and the record.from_address is the url for that message_id. This is just to avoid duplicate looping

modules/templates/SAMBRO/parser.py
flavour: I have looked at the full file & it still makes no sense to me.
We have downloaded all the entries from the RSS feed in the Scheduled Poll
If we don't yet have a record in the CAP table then we assume that there must be an embedded link. (Aren't such links *always* embedded? Why mention that in the else clause?) and so we download the whole feed again to extract it!?
This seems mightily inefficient.

modules/templates/SAMBRO/parser.py
flavour: Sorry, still not clear.
Are there still errors?
If so, can we isolate them to prevent touching parser status inside the parser?

biplovbhandari: @flavour How do we isolate the broken alerts to prevent touching parser status table?

modules/templates/SAMBRO/parser.py
flavour: ok, so the key problem here is this.
The usual case for RSS is that the link tag inside an entry would point to that item alone...in an unspecified format...could be HTML/RSS/CAP
The parser needs to be generic (does it? Or can we have separate ones in each sub-template?)
We want to look for the CAP.
For some sources this is the link (correct?) & for some it is the links entry.
The links entry is currently not parsed by the main poll, which is why you download the feed again to extract it.
I can't find any note on 'links' as a standard tag in RSS, however the right place to parse it seems to be in the core (to avoid having to parse feeds twice). What I do see is that it is perfectly legal to have multiple 'link' entries, though.
I think we should create a msg_rss_link table to store the multiple links in (component of msg_rss)
I don't think we need a deployment_setting for use_links since it's no faster to check that than simple to check for entry.links...so here the core will extract the links & the parser can simply read them to find the one which is 'type': u'application/cap+xml
How does that sound?
(@ToDo: Later we can reuse the same mechanism to handle multiple <link> entries

modules/templates/SAMBRO/parser.py
flavour: Avoid the 3rd loop ;)
All will go away if we do no RSS downloading in the parser...only the CAP

modules/templates/SAMBRO/parser.py
flavour: I need to understand exactly where it crashes & why

biplovbhandari: @flavour If we try to import this alert http://www.dmhwarning.gov.mm/eden/cap/alert/804.cap, there is no event_type_id (required) field in the xml, so it gives out the error from here: https://github.com/sahana/eden/blob/master/modules/s3/s3import.py#L2109

Not sure if this is helpful to isolate the broken alerts preventing to touch parser status table?

Please review https://github.com/biplovbhandari/eden/commit/9ce0dca65da80975848e92b5141eb3ef6e055b21

nursix: Biplov--

not sure whether this helps, but you can prevent rollbacks on validation errors by adding ignore_errors=True to the import_xml. This will then walk through the entire tree, import all correct records but skip the invalid ones. You can still see all errors in resource.error and resource.error_tree.

Dominic

PR #1324, opened 2016-09-11
SAMBRO: fix CAP RSS Feed import

biplovbhandari:
@flavour Please Merge

### DIFF
--- a/modules/s3/s3msg.py
+++ b/modules/s3/s3msg.py
@@ -2016,6 +2016,8 @@ class S3Msg(object):
         ginsert = gtable.insert
         mtable = db.msg_rss
         minsert = mtable.insert
+        ltable = db.msg_rss_link
+        linsert = ltable.insert
         update_super = s3db.update_super
 
         # Is this channel connected to a parser?
@@ -2104,6 +2106,8 @@ class S3Msg(object):
                     # Don't die on badly-formed Geo
                     pass
 
+            # Get links - these can be multiple with certain type
+            links = entry.get("links", [])
             if exists:
                 db(mtable.id == exists.id).update(channel_id = channel_id,
                                                   title = title,
@@ -2115,6 +2119,25 @@ class S3Msg(object):
                                                   tags = tags,
                                                   # @ToDo: Enclosures
                                                   )
+                if links:
+                    query_ = (ltable.rss_id == exists.id) & (ltable.deleted != True)
+                    for link_ in links:
+                        url_ = link_["url"]
+                        type_ = link_["type"]
+                        query = query_ & (ltable.url == url_) & \
+                                (ltable.type == type_)
+                        dbset = db(query)
+                        row = dbset.select(ltable.id, limitby=(0, 1)).first()
+                        if row:
+                            dbset.update(rss_id = exists.id,
+                                         url = url_,
+                                         type = type_,
+                                         )
+                        else:
+                            linsert(rss_id = exists.id,
+                                    url = url_,
+                                    type = type_,
+                                    )
                 if parser:
                     pinsert(message_id = exists.message_id,
                             channel_id = channel_id)
@@ -2132,6 +2155,11 @@ class S3Msg(object):
                               )
                 record = dict(id=_id)
                 update_super(mtable, record)
+                for link_ in links:
+                    linsert(rss_id = _id,
+                            url = link_["url"],
+                            type = link_["type"],
+                            )
                 if parser:
                     pinsert(message_id = record["message_id"],
                             channel_id = channel_id)
@@ -2258,7 +2286,7 @@ class S3Msg(object):
                                       ).first()
         if old_status:
             # Update
-            if status[0] == "+":
+            if status and status[0] == "+":
                 # Increment status if-numeric
                 old_status = old_status.status
                 try:
--- a/modules/templates/SAMBRO/MV/config.py
+++ b/modules/templates/SAMBRO/MV/config.py
@@ -2,6 +2,8 @@
 
 from gluon import current
 
+from collections import OrderedDict
+
 def config(settings):
     """
         Maldives specific template settings for CAP: Common Alerting Protocol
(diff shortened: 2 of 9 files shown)


biplovbhandari:
@flavour Please merge

modules/s3/s3msg.py
```diff
@@ -2115,6 +2120,24 @@ def poll_rss(channel_id):
...
                                                   # @ToDo: Enclosures
                                                   )
+                query_ = (ltable.rss_id == exists.id)
+                for link_ in links:
+                    url_ = link_["url"]
+                    type_ = link_["type"]
+                    query = query_ & (ltable.url == url_) & \
+                            (ltable.type == type_)
+                    set = db(query)
+                    row = set.select(ltable.id, limitby=(0, 1)).first()
```
nursix: Select inside loop - this should better be moved out

modules/s3/s3msg.py
```diff
@@ -2115,6 +2120,24 @@ def poll_rss(channel_id):
...
+                    if row:
+                        set.update(rss_id = exists.id,
+                                   url = url_,
+                                   type = type_)
+                    else:
+                        linsert(rss_id = exists.id,
+                                url = url_,
+                                type = type_,
+                                )
+                db(ltable.rss_id == exists.id).update()
```
nursix: Empty update?

modules/s3/s3msg.py
nursix: Besides - "set" is a reserved keyword, should not be used as variable name. Use "dbset" instead.

modules/templates/SAMBRO/parser.py
```diff
@@ -317,113 +329,56 @@ def parse_rss_2_cap(message):
...
+                                request = None
+                            try:
+                                file = urllib2.urlopen(request).read() if request else fetch(url)
+                            except urllib2.HTTPError, e:
+                                current.log.error("Getting content from link failed: %s" % e)
+                            else:
+                                File = StringIO(file)        
+                                # Import via XSLT
+                                resource = s3db.resource("cap_alert")
+                                stylesheet = os.path.join(current.request.folder, "static", "formats", "cap", "import.xsl")
```
nursix: Should you really repeat os.path.join inside the loop? Or s3db.resource, for that matter?

modules/templates/SAMBRO/parser.py
```diff
@@ -317,113 +329,56 @@ def parse_rss_2_cap(message):
...
+                    rows_ = db(query_).select(ltable.type,
+                                              ltable.url)
+                    for row_ in rows_:
+                        url = row_.url
+                        if url and row_.type == "application/cap+xml":
+                            if username and password:
+                                import base64
+                                request = urllib2.Request(url)
+                                base64string = base64.encodestring("%s:%s" % (username, password))
+                                request.add_header("Authorization", "Basic %s" % base64string)
```
nursix: Are unsolicited Auth headers really required? Seems wrong to me to unsolicitedly send the same username/password to multiple URLs. Wouldn't a 401 handler (opener) be more appropriate? And wouldn't different URLs require different credentials?

nursix: NB if the linked URLs are external sites, then you're giving away the user's credentials to sites which shouldn't see them. Are you sure this is safe operation in the context of the deployment?

biplovbhandari:
@flavour Please Merge

modules/s3/s3msg.py
```diff
@@ -2016,6 +2016,8 @@ def poll_rss(channel_id):
         ginsert = gtable.insert
         mtable = db.msg_rss
         minsert = mtable.insert
+        ltable = db.msg_rss_link
+        linsert = ltable.insert
```
flavour: These too could be behind the 'if links'

modules/s3/s3msg.py
```diff
@@ -2115,6 +2120,28 @@ def poll_rss(channel_id):
...
+                    query = query_ & (ltable.url == url_) & \
+                            (ltable.type == type_)
+                    dbset = db(query)
+                    row = dbset.select(ltable.id, limitby=(0, 1)).first()
+                    if row:
+                        dbset.update(rss_id = exists.id,
+                                     url = url_,
+                                     type = type_,
+                                     username = username,
+                                     password = password,
```
flavour: How do we know that we should use the same username/password for the links?
Can we check that the links are on the same domain? & only then use the same username/password?

modules/templates/SAMBRO/MV/config.py
```diff
@@ -14,4 +16,26 @@ def config(settings):
...
+        #("my", "မြန်မာစာ"),        # Burmese
+        #("ne", "नेपाली"),          # Nepali
+        #("prs", "دری"),        # Dari
+        #("ps", "پښتو"),        # Pashto
+        #("tet", "Tetum"),
+        #("th", "ภาษาไทย"),        # Thai
+        #("tl", "Tagalog"), # Filipino
+        #("vi", "Tiếng Việt"),   # Vietnamese
+        #("zh-cn", "中文 (简体)"),
+    ])
```
flavour: Same as for MM

modules/templates/SAMBRO/config.py
```diff
@@ -115,6 +115,8 @@ def config(settings):
         #("zh-cn", "中文 (简体)"),
     ])
     settings.cap.languages = languages
+    # For SAMBRO
```
flavour: Since this is the SAMBRO template, a 'For SAMBRO' comment seems superfluous?

modules/s3/s3msg.py
biplovbhandari: links are for individual entry

biplovbhandari:
@flavour Please merge

biplovbhandari:
@flavour Please merge

modules/s3/s3msg.py
flavour: eh?
I am suggesting moving these 2 lines underneath L2126
They are only ever needed 'if links' so better moved there, no?

biplovbhandari: Yes but if links is in loop for entry in entries. So isn't better to define once outside loop?

biplovbhandari:
@flavour Please merge

modules/s3/s3msg.py
flavour: Ah right, ok, couldn't see that in the diff..OK thanks for the other change, merging :)

PR #1325, opened 2016-10-20
SAMBRO: theme for each country + alert hub map view page + some fixes

biplovbhandari:
@flavour Please merge


nursix:
We definitely shouldn't duplicate core stylesheets just for the sake of paths - that is absolutely unmaintainable.

nursix:
Since the problem is only with eden.min.css, I think either the minifier should adjust any relative paths in the concatenated CSS, or place the eden.min.css into a folder at the right level (with a different name, maybe).

But no replicating of core stylesheets please - this is just....irrrgh, no go.

biplovbhandari:
how do i debug build.sahana.py? I am in Windows. I tried new Debug Configuration in Eclipse, and the main module to run as the path to build.sahana.py. My arguments tab has -A css for min css. The process starts and I meet this error: https://github.com/sahana/eden/blob/master/static/scripts/tools/build.sahana.py#L16

flavour:
main module needs to be web2py.py
arguments -S eden -M -R /path/to/build.sahana.py

biplovbhandari:
@flavour Please merge

nursix:
@biplovbhandari Do you really need a full custom theme per country if all you need to customise is the menu logo and favicon? Seems a bit over the top to me.

Check the RMSAmericas theme how things can be made dynamic:
https://github.com/sahana/eden/blob/master/modules/templates/RMSAmericas/views/layout.html#L78
https://github.com/sahana/eden/blob/master/modules/templates/RMSAmericas/layouts.py#L256

Should be easy enough to implement something similar to change the menu logo (and favicon) per settings.template - and minimal style differences can be caught by CSS classes.

flavour:
@nursix : Why not make this the default then?
Currently this is the docs which you are well aware I was writing as I just did exactly this for SCPHIMS:
http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/Themes#Simplemodifications

nursix:
@flavour Making the menu logo of the default layout configurable is already on my list (we discussed this before). However, SAMBRO is using a custom theme/layout, so this wouldn't immediately benefit from the enhancement - and the question here was about the need for sub-themes.

nursix:
And anyway @flavour, my standpoint is that menu logo and favicon are actually contents (i.e. template), not styles (theme) - hence having the logo in the theme isn't quite right.

PR #1326, opened 2016-10-25
SAMBRO: add thai language as well in default theme + SAMBRO: fix + Allow group to be imported through xsl

biplovbhandari:
@flavour Please merge


static/formats/s3csv/pr/person.xsl
```diff
@@ -269,6 +273,19 @@
             </xsl:if>
 
         </resource>
+        
+        <!-- Groups -->
```
nursix: This should really be done via an index (see Education Levels), not per-row.

biplovbhandari:
@flavour Please merge

PR #1327, opened 2016-11-01
SAMBRO: layout and logo change

biplovbhandari:
@flavour Please merge


modules/s3cfg.py
```diff
@@ -2161,6 +2161,13 @@ def get_ui_menu_logo(self):
                                )
                            )
 
+    def get_ui_homepage_icon_url(self):
+        """
+            The url for the homepage icon in the layout
+        """
+
+        return URL(c="default", f="index")
```
biplovbhandari: A similar process is done here: https://github.com/sahana/eden/blob/master/modules/s3layouts.py#L148

flavour: Indeed...changing the Logo without needing to build a custom theme is a process we expect to be very common...with websetup this would be configurable via web with ability to upload the new logo...this enhances accessibility a lot.
However the case you have here I envisage to be extremely uncommon & you already have a custom theme that you are maintaining

nursix: Err...there is no need for this really.

You can easily override the whole title_area in the custom menus.py if you with to link it to somewhere else, and I don't like the idea of having two ways of doing the same thing.

In the menus.py of the template, just add the HTML for the title area as in:

    main_menu = MM(title_area = DIV(...))(

For an edge-case, this should be convenient enough.

modules/s3layouts.py
```diff
@@ -160,7 +162,7 @@ def layout(item):
                     #else:
                         # use as-is (assume HTML or T())
                     title_area = A(logo,
-                                   _href = URL(c="default", f="index"),
+                                   _href = settings.ui.homepage_icon_url or URL(c="default", f="index"),
```
nursix: I don't link this, really - if you need to go that far that you want to change even the URL of the logo, then you can as well override the entire title_area (see a few rows above) with an A() instance.

"title_area" is an option to the MM() class, so you can easily apply that to the main_menu in your template's menus.py.

nursix: Here:
https://github.com/sahana/eden/blob/master/modules/templates/SAMBRO/menus.py#L46

You can easily override the title area of the main menu with your own thing, e.g.:

```
    logo = IMG(_src=settings.get_ui_menu_logo(), 
            _alt=settings.get_system_name_short(),
            _class="logo",
            )
    url = URL(c="what", f="ever")

    main_menu = MM(title_area=A(logo, _href=url))(
```

You can go as fancy here as you like (as long as the CSS covers it)...and sure enough, this can apply different URLs per template if required.

If it's only the logo, then the setting would do - if it's more than the logo, then override the entire title area.

nursix: At the same time, from the discussion on the mailing list it seems that you want to permanently replace the homepage for this template with a custom page - so I wonder why you don't do exactly that (we have the option to override default/index per template), but instead try fumbling around with the link?

Define a custom index controller (class index) in controllers.py, and let it set whatever view you like default/index to be.

If you never need the inherited default/index in this template, then why wouldn't you just override it?

nursix: ...and remember that you can just import classes from another template (in case you want to re-use a certain function in the parent template's controllers.py):

```
from ..controllers import something

class index(something):
    pass
```

...and it would inherit the 'something' function from the parent template, and set it as index-page for this template.

Not complicated at all.

nursix: Ok, more concrete:

in modules/templates/SAMBRO/MM/controllers.py:

```
from templates.SAMBRO.controllers import alert_hub_cop

class index(alert_hub_cop):
    pass
```

That should be enough to make alert_hub_cop the default/index controller for the MM template.

One very important point about that vs. the messing with the home-link is that many redirections go to default/index (e.g. after login!), so changing this link in the menu really isn't a solution.

But replacing the default/index controller with the function you want (while leaving the default/index link intact), that should do what you need.

nursix: Actually, it is even more silly than that :D I just tested this as modules/templates/SAMBRO/MM/controllers.py:

```
# -*- coding: utf-8 -*-

from templates.SAMBRO.controllers import alert_hub_cop as index, subscriptions, user_info
```

Seriously...that does it already ;)

nursix: ...although it would spoil your "Alert Hub" menu entry, so maybe you need:
```
# -*- coding: utf-8 -*-

from templates.SAMBRO.controllers import alert_hub_cop as index, subscriptions, user_info

class alert_hub_cop(index):
    pass
```
after all.

I think as long as the dependency of the parent template is given anyway, there is very little that would forbid importing and subclassing from it. Nothing is going to be DRYer than that.

nursix: ...or more efficient, really.

nursix: All in all, I think the ability to override the default/index controller just like that makes it even less necessary (or useful) to be able to override the home-link in a setting - especially when there are so many things that would redirect directly to default/index.

It's simply so that default/index is assumed to be a safe constant throughout the system -  that it can be hard-coded without risk, that's where the system can go if all else fails, that's where it goes when it goes home. 

I see very little point in being able to override that - especially with such an inconsistent option. Hence I would ask for that option to be removed.

biplovbhandari:
@flavour Please merge

PR #1330, opened 2016-11-09
first step for sending CAP file as attachment in email

biplovbhandari:


modules/s3db/doc.py
```diff
@@ -358,6 +358,21 @@ def doc_file_represent(file):
 
     # -------------------------------------------------------------------------
     @staticmethod
+    def doc_get_file_path(document_id):
```
nursix: Not fully correct - must use the uploadfolder of the field. Besides, upload fields have a method to extract the file.

nursix: See here:
https://github.com/sahana/eden/blob/master/modules/s3/s3xml.py#L1250

biplovbhandari:
@flavour Please merge

modules/s3/s3msg.py
```diff
@@ -680,15 +696,34 @@ def dispatch_to_pe_id(pe_id,
 
         # Set a default for non-SMS
         organisation_id = None
+        attachment_table = s3db.msg_attachment
+        document_table = s3db.doc_document
+        file_field = document_table.file
+        mail = current.mail
+        import os
```
flavour: Normally do imports at the module level, especially for a module like this which will inevitably be loaded anyway

modules/s3/s3msg.py
```diff
@@ -680,15 +696,34 @@ def dispatch_to_pe_id(pe_id,
...
                 message = row["msg_email.body"] or ""
                 from_address = row["msg_email.from_address"] or ""
+                query = (attachment_table.message_id == message_id) & \
+                        (attachment_table.deleted != True) & \
+                        (attachment_table.document_id == document_table.id) & \
+                        (document_table.deleted != True)
+                arows = db(query).select(file_field)
+                for arow in arows:
+                    file = arow.file
+                    if file_field.custom_retrieve_file_properties:
```
flavour: Do this check outside the loop
if file_field.custom_retrieve_file_properties:
    retrieve_file_properties = file_field.custom_retrieve_file_properties
else:
    retrieve_file_properties =field_field.retrieve_file_properties

loop:
    prop = retrieve_file_properties (file)

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1357,50 @@ def get_formatted_value(value, represent=None, system=True):
...
+                                                           "export.xsl"),
+                                      pretty_print=True)
+        file_path = path_join(request.folder,
+                              "uploads",
+                              "%s_%s.xml" % (tablename, s3_str(alert_id)))
+        file = open(file_path, "w+")
+        file.write(cap_xml)
+        file.close()
+
+        # Insert into the doc_entity
```
flavour: No, you should never need to manually create records in the super entity.
You shopudl create records in the instane & then use s3db.update_super() to create/update the super-entity

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1357,50 @@ def get_formatted_value(value, represent=None, system=True):
...
+        resource.add_filter(FS("id") == alert_id)
+        cap_xml = resource.export_xml(stylesheet=path_join(request.folder,
+                                                           "static",
+                                                           "formats",
+                                                           "cap",
+                                                           "export.xsl"),
+                                      pretty_print=True)
+        file_path = path_join(request.folder,
+                              "uploads",
+                              "%s_%s.xml" % (tablename, s3_str(alert_id)))
```
flavour: Since alert_id is an integer then can just use normal str()

biplovbhandari:
@flavour Please merge

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1357,41 @@ def get_formatted_value(value, represent=None, system=True):
...
+                              "%s_%s.xml" % (tablename, str(alert_id)))
+        file = open(file_path, "w+")
+        file.write(cap_xml)
+        file.close()
+
+        # Create doc_document record
+        dtable = s3db.doc_document
+        file = open(file_path, "a+")
+        document_id = dtable.insert(file=file)
+        s3db.update_super(dtable, {"id": document_id})
```
flavour: doc_document is NOT an instance of the doc_entity super entity
In this case it's cap_resource
doc_document is a component of the super-entity...

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1357,40 @@ def get_formatted_value(value, represent=None, system=True):
...
+                              "uploads",
+                              "%s_%s.xml" % (tablename, str(alert_id)))
+        file = open(file_path, "w+")
+        file.write(cap_xml)
+        file.close()
+
+        # Create doc_document record
+        file = open(file_path, "a+")
+        document_id = s3db.doc_document.insert(file=file)
+        s3db.update_super(s3db.cap_resource, dict(id=document_id))
```
biplovbhandari: @flavour some question.
1. DO I need to update_super? Because I am creating records in the component (doc_document) and not in instance (cap_resource). The documentation talks about the instance table record creation/manipulation/deletion <a href="http://eden.sahanafoundation.org/wiki/S3/S3Model/SuperEntities#UpdatingaSuper-Entity" target="_blank">here</a>
2. when I update_super, does that mean a new record in super entity referencing the instance table is created?

flavour: 1) If all you are doing is creating component records then no need to update_super.
However to make it a component then you will need to read the doc_id of the parent & insert that into the component.
2) update_super will create/update records in the super-entity table & then add the superkey to the instance table

nursix: To answer your questions:
1) doc_document is not an SE instance table, so calling update_super is not required. However, calling it will certainly not harm - *if* you call it correctly. Here though you call it for the cap_resource table, but you pass a doc_document record ID, and that is clearly wrong (you must call it for the table the record belongs to!)

2) update_super does what it says - it updates all super-entity records for the instance records (copying all shared fields). Where no SE entry yet exists, it will create one. It is mandatory for all SE instance tables. Now, doc_document is no SE instance, so you could omit it (however, it doesn't harm to call it anyway)

Regarding your procedure here:

The way you do it, you're not attaching the document record to anything? Entries in doc_document can be unattached, of course - but the normal way is to attach document to some master record, and in this case it seems it should be attached to the alert at least.

If you want to create a document record for the alert, then you need to:
- extract the doc_id of the alert record, and add it to the document record
- create a new doc_document record
- run the usual update_super (on the right table!!), set_record_owner, onaccept etc

    # Get the alert's doc_id
    atable = s3db.cap_alert
    row = db(atable.id==alert_id).select(atable.doc_id, limitby=(0, 1)).first()

    # Create a new doc_document record
    dtable = s3db.doc_document
    record = {"doc_id": row.doc_id, "file": file}
    record_id = dtable.insert(**record)
    record["id"] = record_id

    # Update super
    s3db.update_super(dtable, record)

    # Record owner
    auth = current.auth
    auth.s3_set_record_owner(dtable, record_id)
    auth.s3_make_session_owner(dtable, record_id)

    # Execute onaccept
    s3db.onaccept("doc_document", record, method="create")

Of course, if it's not an SE instance (like in this case), then update_super may not be required. And if there is no onaccept defined, then calling onaccept may not be required either.

But neither call will fail (or do anything) if not required - and if anything changes in future or in a template (e.g. an onaccept being added), then calling them in any case you will be on the safe side.

Record ownership may be relevant in any case, and certainly the doc should be attached to the alert

You /can/ have this shorter:

    data = {"doc_id": row.doc_id, "file": file}
    dresource = s3db.resource("doc_document")
    record_id = dresource.insert(**data)

This will do all of the above as required. However, it will also check permissions to create a new record - and if the current user doesn't have permissions to create doc_documents, then this will raise an S3PermissionError. You could override it, of course.

---

Now, I wonder though why you want to store the XML as a doc_document - it's the alert XML, right? Why keep it? For just sending it, it'd be good enough to create a temp file, attach it to the message(s), and remove the file again after send.

I see little point in creating a new doc_document record with the alert XML every time you send the alert - it'll be the same XML over and over again, and it wouldn't be anything that is not already in the alert record or its components - so it seems to me that you are replicating data here? Why?

biplovbhandari:
@flavour Please merge

models/tasks.py
```diff
@@ -235,19 +235,21 @@ def org_facility_geojson(user_id=None):
 if settings.has_module("msg"):
 
     # -------------------------------------------------------------------------
-    def msg_process_outbox(contact_method, user_id=None):
+    def msg_process_outbox(contact_method, file_path, user_id=None):
```
flavour: Why does the task have a file_path?

modules/s3/s3msg.py
```diff
@@ -400,6 +404,31 @@ def send_by_pe_id(pe_id,
             record = dict(id=_id)
             s3db.update_super(table, record)
             message_id = record["message_id"]
+            ainsert = s3db.msg_attachment.insert
```
flavour: Should be inside the branch which uses it

modules/s3/s3msg.py
```diff
@@ -400,6 +404,31 @@ def send_by_pe_id(pe_id,
...
+            if document_id:
+                if not isinstance(document_id, list):
+                    document_id = [document_id]
```
flavour: generally I prefer arg names which imply their meaning
document_id suggests single...if multiple, then perhaps 'documents'

modules/s3/s3msg.py
```diff
@@ -400,6 +404,31 @@ def send_by_pe_id(pe_id,
...
+                for _document_id in document_id:
+                    ainsert(message_id=message_id,
+                            document_id=_document_id,
+                            )            
+            elif file_path and url:
+                # Both file_path and url, given so we can store the url
+                # and skip creating doc_id
```
flavour: One sdhould never create a document_id within message sending anyway, no need to mention that I don't think.
I am really puzzled by this....why have both a file_path & a URL in the same branch?
file_path should be if we have a file
URL if we want to download a URL & send the results...seems a bit of an edge case but handy when wanted so I guess am happy to support within core messaging.

So I see 3 args required for send_by_pe_id:
documents=None (list of document_ids)
files = None (list of file_paths)
urls = None (list of URL to download & attach results)

For this to work then msg_attachment needs to have 3 corresponding fields:
document_id (for when the attachment is a doc_document)
file (for when the attachment is a file)
url (for when the attachment is downloaded, results attached & then file discarded)

I generally don't like the 3rd option, I have to say, as the URL can give different results on different days...not a great history of what is sent, however I do see that it saves storage space in the msg_attachment table.

However for this usecase, contrary to what Dominic says, I personally see no big deal with storing the CAP XML for an Alert with the Alert if you need to send this with some of ther subscriptions.
I certainly agree with him that you shouldn't generate this multiple times, that seems ludicrous...the solution, to me, would be to store this in doc_document & then the subscriptions create msg_attachment pointing to document_id...this seems to be less of a CPU hog then creating the XML for each attachment separately...and CPU is far more of a bottleneck concern to me than diskspace.

So...whilst I see some value to add support to core messaging for attachments which are files & attachments which are URLs, personally I would prefer to just stick to using doc_document as this:
* Ensures we have a proper history of what is sent
* We should be able to avoid storing multiple copies of the same file
* We should be able to avoid the overheads of generating/downloading multiple copies of the same file

modules/s3/s3msg.py
```diff
@@ -680,15 +713,39 @@ def dispatch_to_pe_id(pe_id,
...
+                        (document_table.deleted != True)
+                arows = db(query).select(file_field)
+                if len(arows):
+                    # for document_id
+                    for arow in arows:
+                        file = arow.file
+                        prop = retrieve_file_properties(file)
+                        _file_path = os.path.join(prop["path"], file)
+                        attachments.append(mail_attachment(_file_path))
+                elif file_path:
```
flavour: A run of process_outbox has to be able to support attachments in multiple formats

biplovbhandari:
@flavour Please merge

modules/s3/s3crud.py
```diff
@@ -1971,11 +1971,7 @@ def review(self, r, **attr):
                     resource = current.s3db.resource(r.tablename, r.id,
                                                      approved=False,
                                                      unapproved=True)
-                    try:
-                        success = resource.approve()
-                    except:
-                        success = False
-                    if success:
+                    if resource.approve():
```
flavour: Why did you remove the try/except here?

modules/s3db/cap.py
```diff
@@ -1595,6 +1598,15 @@ def model(self):
             msg_list_empty = T("No Acknowledgements currently received for this alert"))
 
         # ---------------------------------------------------------------------
+        # CAP Attachment table
+
+        tablename = "cap_attachment"
```
flavour: Why this new table?
If you want documents to be a component of Alerts then make cap_alert a doc_entity instance.
Although currently cap_resource already is...so could that be used instead?

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1358,62 @@ def get_formatted_value(value, represent=None, system=True):
...
+    def _get_or_create_attachment(alert_id):
+        """ 
+            Retrieve the CAP attachment for the alert_id if present
+            else creates CAP file as attachment to be sent with the email
+            returns the doc_id for the CAP file
+        """
+
+        s3db = current.s3db
+        atable = s3db.cap_attachment
+        query = (atable.alert_id == alert_id) & (atable.deleted != True)
```
biplovbhandari: but then we can upload multiple doc for a alert from cap_resource UI. SO I created new table to store the cap xml only

flavour:
Can easily modify the UI for a single record by either setting the form's
inline component to multiple=False or even redefining the component as
signle in this view

PR #1332, opened 2016-11-12
first step for sending CAP file as attachment in email

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/templates/SAMBRO/config.py
+++ b/modules/templates/SAMBRO/config.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import json
+import os
 
 from collections import OrderedDict
 
@@ -401,6 +402,9 @@ def config(settings):
             # Single row as we are filtering for particular alert_id
             arow = data["rows"][0]
 
+            # Create attachment
+            cap_document_id = _get_or_create_attachment(alert_id)
+
             if record["scope"] != "Private" and data["numrows"] > 0:
                 # Google Cloud Messaging
                 stable = s3db.pr_subscription
@@ -469,8 +473,8 @@ def config(settings):
                 get_user_id = current.auth.s3_get_user_id
                 query_ = (gtable.id == mtable.group_id) & \
                          (mtable.person_id == ptable.id) & \
-                         (gtable.deleted != True) &\
-                         (mtable.deleted != True) &\
+                         (gtable.deleted != True) & \
+                         (mtable.deleted != True) & \
                          (ptable.deleted != True)
                 count = len(addresses)
                 if count == 1:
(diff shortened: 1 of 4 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1359,70 @@ def get_formatted_value(value, represent=None, system=True):
...
+        file_path = path_join(request.folder,
+                              "uploads",
+                              "%s_%s.xml" % ("cap_alert", str(alert_id)))
+        file = open(file_path, "w+")
+        file.write(cap_xml)
+        file.close()
+
+        # Create doc_entity record
+        etable = s3db.doc_entity
+        doc_id = etable.insert(**{"instance_type": "cap_resource"})
```
flavour: No! As-before the doc_entity record is created by update_super.
In this case the doc_entity should already be created by the cap_resource

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1359,70 @@ def get_formatted_value(value, represent=None, system=True):
 
                 return nvalue
 
+    # -------------------------------------------------------------------------
+    def _get_or_create_attachment(alert_id):
+        """ 
+            Retrieve the CAP attachment for the alert_id if present
+            else creates CAP file as attachment to be sent with the email
+            returns the doc_id for the CAP file
```
flavour: doc_id or document_id?
I think it's document_id

modules/templates/SAMBRO/config.py
flavour: if cap_resource defintiely exists already, then we can (hopefully!) assume that update_suoper was run as part of that record's creation, so we should just read the doc_id from that

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1359,70 @@ def get_formatted_value(value, represent=None, system=True):
...
+        # Create doc_entity record
+        etable = s3db.doc_entity
+        doc_id = etable.insert(**{"instance_type": "cap_resource"})
+
+        # Create doc_document record
+        dtable = s3db.doc_document
+        file = open(file_path, "a+")
+        document_id = dtable.insert(**{"file": file, "doc_id": doc_id})
+
+        # Update the cap_resource table
```
flavour: Create the cap_resource record

NB This needs to happen before the doc_document is created, so that we can read the doc_id

modules/templates/SAMBRO/config.py
```diff
@@ -1349,4 +1359,70 @@ def get_formatted_value(value, represent=None, system=True):
...
+        record = {"alert_id": alert_id,
+                  "resource_desc": T("CAP XML File"),
+                  "doc_id": doc_id,
+                  "mime_type": "cap" # Hard coded to separate from attachment from user
```
flavour: +1: I like this :)

modules/templates/SAMBRO/config.py
biplovbhandari: s3db.update_super(rtable, record) creates new record in doc_entity but how do I read the doc_id? This function returns boolean

flavour: https://github.com/sahana/eden/blob/master/modules/s3/s3msg.py#L394

biplovbhandari:
@flavour Please merge

PR #1336, opened 2016-11-28
UAV: started working on template for AIT

biplovbhandari:


nursix:
Why can't this use the standard supply/asset modules to manage the UAVs? I don't see anything special here that can not be handled with the existing data models.

biplovbhandari:
I am thinking of putting the images coming from UAV, reading EXIF data in the model along with visualization and other

nursix:
So? I still don't see why this can not use the supply/asset module to manage the devices. And images would be doc, so they can easily be linked to context models.

nursix:
I don't think that we want to introduce a separate module for each type of sensor/observation equipment.

The devices themselves are just assets, and would typically be managed together with all other assets of an organization.

The data they produce are either context-specific (e.g. incident data) and should hence go into the respective context modules - or they are generic (e.g. weather data, areal imagery) and then should go into a generic module like doc, gis or the like so that they can be linked to context data.

And /if/ the data are only linked to the UAVs themselves (and not any humanitarian/emergency management operation), then we need to ask whether that should be in Sahana at all. Otherwise I want a lego module to manage the bricks for the local kindergarten.

biplovbhandari:
Yes the data are context-specific (mostly damage assessment, mapping and sometimes incident data). I will try to play with assests module. For images, I was thinking of using gis and doc modules.

nursix:
Okay, I think this will be a better approach. You may want to close this while you're working on it? Or do you want a complete review now?

flavour:
Note that the core Assets module can  be extended with UAV-specifics if-required, just like for Vehicles (& there are stubs for Telephone/Radios)...in which case a UAV module may still develop, but just not to repeat the core stuff like Manufacturer.

biplovbhandari:
Which template uses Vehicle specific assets?

flavour:
Not a template, but a module:
https://github.com/sahana/eden/blob/master/modules/s3db/vehicle.py
https://github.com/sahana/eden/blob/master/modules/s3db/asset.py#L130

PR #1337, opened 2016-12-01
SAMBRO: fixes

biplovbhandari:
@flavour Please merge

### DIFF
--- a/modules/s3/s3utils.py
+++ b/modules/s3/s3utils.py
@@ -519,9 +519,9 @@ def s3_truncate(text, length=48, nice=True):
     text = s3_unicode(text)
     if len(text) > length:
         if nice:
-            return "%s..." % text[:length].rsplit(" ", 1)[0][:45]
+            return "%s..." % text[:length].rsplit(" ", 1)[0][:length-3]
         else:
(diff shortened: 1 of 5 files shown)


modules/s3/s3utils.py
```diff
@@ -516,12 +516,12 @@ def s3_truncate(text, length=48, nice=True):
     """
 
     # Make sure text is multi-byte-aware before truncating it
-    text = s3_unicode(text)
+    text = s3_str(text)
```
flavour: Read the comment above the line you changed & see that you break that

modules/templates/SAMBRO/config.py
```diff
@@ -303,6 +303,29 @@ def onaccept(form):
     # -------------------------------------------------------------------------
     def customise_org_organisation_resource(r, tablename):
 
+        s3 = current.response.s3
+
+        crud_strings_branch = Storage(
+            label_create = T("Add Branch"),
```
flavour: Seems very odd to only ever be able to create branches?
Surely this should only be within a context of some kind?

biplovbhandari:
@flavour fixed...please merge

PR #1338, opened 2016-12-01
SAMBRO: subscription form allows the choice of attachment, send attachment from s3notification and implement same in SAMBRO

biplovbhandari:
@flavour Please merge......

### DIFF
--- a/modules/s3/s3notify.py
+++ b/modules/s3/s3notify.py
@@ -119,6 +119,7 @@ class S3Notifications(object):
                                                   stable.notify_on,
                                                   stable.method,
                                                   stable.email_format,
+                                                  stable.attachment,
                                                   rtable.id,
                                                   rtable.resource,
                                                   rtable.url,
@@ -195,6 +196,7 @@ class S3Notifications(object):
                            "notify_on": s.notify_on,
                            "method": s.method,
                            "email_format": s.email_format,
+                           "attachment": s.attachment,
                            "resource": r.resource,
                            "last_check_time": last_check_time,
                            "filter_query": query_nice,
@@ -356,6 +358,14 @@ class S3Notifications(object):
                                                     r="%(resource)s")
         subject = subject % meta_data
 
+        # Attachment
+        attachment = subscription.get("attachment", False)
+        document_ids = None
+        if attachment:
+            attachment_fnc = settings.get_msg_notify_attachment()
+            if attachment_fnc:
+                document_ids = attachment_fnc(resource, data, meta_data)
+
         # Helper function to find templates from a priority list
         join = lambda *f: os.path.join(current.request.folder, *f)
         def get_template(path, filenames):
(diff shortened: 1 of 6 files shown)


modules/templates/SAMBRO/config.py
```diff
@@ -966,6 +966,8 @@ def pr_subscription_filter_row_options():
             etable = s3db.event_event_type
             ptable = s3db.cap_warning_priority
             filter_options = {}
+            from s3.s3validators import IS_ISO639_2_LANGUAGE_CODE
```
nursix: Just

   from s3 import IS_ISO639_2_LANGUAGE_CODE

flavour:
Seems fine, but I see no pressure to merge this WiP?

biplovbhandari:
@flavour completed...could you review?

modules/s3/s3notify.py
```diff
@@ -356,6 +358,14 @@ def send(cls, r, resource):
                                                     r="%(resource)s")
         subject = subject % meta_data
 
+        # Attachment
+        attachment = subscription.get("attachment", False)
+        if attachment:
+            attachment_fnc = settings.get_msg_notify_attachment()
+            documents_id = attachment_fnc(resource, data, meta_data)
```
nursix: I think that should be "document_ids" instead of "documents_id"? (so it does indicate multiple ids rather than a single id)

modules/s3cfg.py
```diff
@@ -2285,6 +2285,13 @@ def get_msg_notify_renderer(self):
         """
         return self.msg.get("notify_renderer")
 
+    def get_msg_notify_attachment(self):
+        """
+            The document_id that is sent as attachment with the email
```
nursix: Not quite true - it's supposed to return a function that produces an array of document IDs (multiple)

biplovbhandari:
@flavour Please merge

modules/s3/s3notify.py
```diff
@@ -356,6 +358,14 @@ def send(cls, r, resource):
                                                     r="%(resource)s")
         subject = subject % meta_data
 
+        # Attachment
+        attachment = subscription.get("attachment", False)
+        if attachment:
+            attachment_fnc = settings.get_msg_notify_attachment()
+            document_ids = attachment_fnc(resource, data, meta_data)
```
flavour: Would attachment_fnc always be defined?
I think there should be an if attachment_fnc here

modules/s3cfg.py
```diff
@@ -2285,6 +2285,14 @@ def get_msg_notify_renderer(self):
         """
         return self.msg.get("notify_renderer")
 
+    def get_msg_notify_attachment(self):
+        """
+            Custom function that returns the list of document_ids to be sent
+            as attachment in email
+        """
```
flavour: Good to add the args that the fnc needs to support here

biplovbhandari:
@flavour Please merge

PR #1342, opened 2017-01-16
fix error to locale filenames containing '-'

whanderley:
This is a simple modification to prevent errors that occour when a locale filename containing '-', like 'pt-BR', is in use.


flavour:
Thankyou, silly mistake indeed!
I will fix but not as a merge since you also have a bunch of changes within the locale file here that I cannot easily verify. Thanks again :)

PR #1343, opened 2017-01-24
WACOP incident profile page design and layout

dhornbein:
set WACOP fonts

cleared existing custom styles

parent event callout and system wide alert styling

info box inital style

incident profile refactoring

tagit ui style overwrite

refactor info bar

arrange incident-profile for mock v2

incident profile resource section

meta-text mixin

incident profile css update

tab styling

filter form html and styles

post update addition

foundation production

tab design and layout tweak

adding desiered varialbes to controller

adding desiered varialbes to controller

bug fix

implemented new filter_form


modules/templates/WACOP/controllers.py
```diff
@@ -787,6 +787,39 @@ def _html(self, r, **attr):
...
+                               )
+
+        # output["edit_btn"] = ""
+
+        #<a href='#' title='delete'><i class='fa fa-trash'><span class='show-for-sr'>delete</span></i></a>
+        output["trash_btn"] = A(ICON("delete"),
+                                SPAN("delete this",
+                                     _class="show-for-sr",
+                                     ),
+                                _title="delte this",
```
flavour: typo ;)

modules/templates/WACOP/controllers.py
```diff
@@ -787,6 +787,39 @@ def _html(self, r, **attr):
...
+                                _href="#"
+                               )
+
+        output["modified_by"] = "admin@example.com"
+        output["incident_tags_list"] = ""  # UL list of tags with link to list of incidents with same tag
```
flavour: Is this different to the below 'incident_tags' ?

modules/templates/WACOP/controllers.py
```diff
@@ -787,6 +787,39 @@ def _html(self, r, **attr):
...
+        #<a href='#' title='delete'><i class='fa fa-trash'><span class='show-for-sr'>delete</span></i></a>
+        output["trash_btn"] = A(ICON("delete"),
+                                SPAN("delete this",
+                                     _class="show-for-sr",
+                                     ),
+                                _title="delte this",
+                                _href="#"
+                               )
+
+        output["modified_by"] = "admin@example.com"
```
flavour: from s3 import s3auth_user_represent
output["modified_by"] = s3auth_user_represent(record.modified_by)

modules/templates/WACOP/views/incident_profile.html
```diff
@@ -34,183 +32,422 @@ <h1 class="title left">{{=event.name}}</h1>
...
+          </div>
+        </div>
+
+        <div class="row" data-equalizer>
+          <div class="medium-6 columns">
+            <section class="incident-data" data-equalizer-watch>
+              <p class="adr">
+                <span class="street-address">123 Fake Street</span><br>
+                <!-- <span class="street-address-2"></span><br> -->
+                <span class="city">Seattle</span>,
```
flavour: Why are we returning to lots of hardcoded dummy data here?
This seems a ste[p backwards when the previous version had live data in?

modules/templates/WACOP/views/incident_profile.html
```diff
@@ -34,183 +32,422 @@ <h1 class="title left">{{=event.name}}</h1>
...
+            <h1 class="title">{{=name}}</h1>
+            <h6 class="meta">updated {{=modified_on}} by {{=modified_by}}
+              {{#incident_tags_list}}
+              <ul class="tags">
+                <li class="tag"><a href="#">Terrorism</a></li>
+                <li class="tag"><a href="#">Bomb</a></li>
+                <li class="tag"><a href="#">Crime</a></li>
+                <li class="tag"><a href="#">Alert</a></li>
+              </ul>
+              {{#end incident_tags_list}}
```
flavour: I'm not keen to replace a working system with a hardcoded set of dummy data...sure we can work on the styling of the working system, but I think that should happen outside of a Pull Request. NB For the Tags system our options may be limited as we're using a plugin to handle the Tags...we can investigate alternate plugins or even try rolling our own, but this then starts to get a lot more time-consuming

modules/templates/WACOP/views/incident_profile.html
```diff
@@ -34,183 +32,422 @@ <h1 class="title left">{{=event.name}}</h1>
...
+                <div id='dt-event_team'>
+                  {{=event_team_datatable}}
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      </section>
+      
+      <!-- Modal for edit, probably should be converted to jQuer UI element -->
```
flavour: Currently this is a dummy form isn't it?
If you're happy to move to a jQueryUI popup, then I think best not to include this in the Pull Request...just the button which should invoke it...I will then add a default update form which we can then subsequently try to style like this - basically this would be a formstyle like for the filter widgets.

modules/templates/WACOP/views/incident_profile.html
```diff
@@ -34,183 +32,422 @@ <h1 class="title left">{{=event.name}}</h1>
...
+          {{if create_post_form:}}
+          <div id='modal-post-update' class='reveal-modal' data-reveal aria-labelledby='modalPostUpdate' aria-hidden='true' role='dialog'>
+          <h4 id='modalPostUpdateTitle'>New Update</h4>
+          {{=create_post_form}}
+          <a class='close-reveal-modal' aria-label='Close'>&#215;</a>
+          </div>
+          {{pass}}
+          <ul class='dl' id="updates_datalist">
+            {{#updates_datalist}}
+            <!-- Data List Flat HTML -->
```
flavour: Again you're replacing working live code with dummy flat stuff...I really am reluctant to do this in a Pull Request

modules/templates/WACOP/views/layout.html
```diff
@@ -18,9 +18,10 @@
 <link href="{{=sheet}}" rel="stylesheet" type="text/css" media="screen" charset="utf-8" />
 {{pass}}
 <link href="/{{=appname}}/static/themes/foundation/normalize.css" rel="stylesheet" />
-<!--
-<link href='https://fonts.googleapis.com/css?family=Merriweather:400,700|Source+Sans+Pro:400,700' rel='stylesheet' type='text/css'>
--->
+
+<!-- Google Fonts Roboto Condensed and Open Sans -->
+<link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Roboto+Condensed:400,700" rel="stylesheet">
```
flavour: Just a note that if we're using Offsite fonts then that means that we cannot operate the system offline. This may not be an issue for WACOP but in general Incident Management systems need to be able to operate offline...this was a very key issue for the Give2LA site for instance.

static/themes/WACOP/scss/style.scss
```diff
@@ -13,16 +15,17 @@ $include-html-global-classes: false;
...
 // CUSTOM MODULES
 
-@import "custom/page";
-@import "custom/forms";
-@import "custom/rheader";
-@import "custom/buttons";
-@import "custom/footer";
+// @import "custom/page";
+// @import "custom/forms";
+// @import "custom/rheader";
```
flavour: If you remove these then all the other default Sahana-style pages will likely break.
As per conversation with Devin, these will be used for all the Admin stuff like taxonomy changes

static/themes/WACOP/styles.css
```diff
@@ -1,502 +0,0 @@
...
-/* Updates */
-
-.update-meta-text {
-  font-size: 13px;
-}
-
-
-/**
- * Timeline Slider
- */
```
flavour: So you've removed all this CSS, but I don't see you've put it anywhere else?
So this will then break the Incident Browse page...perhaps some of the rest of the code in this file is also needed for that page...did you check it at all?

flavour:
Overall this is premature for a Pull Request: This breaks a lot of functionality for other pages & even within this page it replaces a lot of dynamic, working content with static dummy data....so overall is quite a major step backwards.
In general, smaller pull requests are better than larger ones...easier to review & less likely to give problems.
For providing desired syling updates to dynamic conetns, we need to use a different system to a Pull Request. It can certainly be flat HTML with dummy data, liek this, and it can even be code in GitHub, but then probably a different branch to the one being used to submit Pull Requests so they don't get muddled up.

PR #1348, opened 2017-03-01
WACOP: sass theme & incident profile refactoring

dhornbein:
Applied basic HTML to incident page and set up SASS theme files to style elements.


flavour:
ok, I merged, but I see an issue in that the 'Add Update' button is visible to all, which I believe is incorrect?
This should only show to those people who have permission to do this?
Which for this usecase is all AUTHENTICATED users, so we could just hide this in the view with {{if auth.is_logged_in():}}BUTTON{{}} or we could be more generic & do the permissions check that we have already in the controller & so currently that would be hiding it behind the {{if create_post_form:}}
I think we should be trying to be generic where possible to make it easier to build SAFIRE.

For the Modal create form, I am assuming that we will use a std jQueryUI popup instead of this one? (Foundation I assume?)

flavour:
NB I can make the change, but would rather do with the Popup change, so will wait a mo...just want to confirm it all with you first.

dhornbein:
re: add update, yes the generic method of hiding that button sounds like the right choice

re: modal, it seems to me that it makes sense to keep all pop-ups as jQueryUI for consistency. I'm not clear on *how* to achieve this, if you can replace the foundation modal with a jQueryUI implementation that should be enough to get me clear so I can execute in the future.

Thanks @flavour

PR #1349, opened 2017-03-30
WACOP component theme integration & minor view edits

dhornbein:
Added wacop theme, mixins, and settings

Removed existing tab styling

Added create_post_form output and HTML wrappers to dashboard

Wrapped tab element in column

Hide data wrapper style for now


dhornbein:
I noticed that I've left a few double quotes in my HTML, sorry! Auto complete...

modules/templates/WACOP/controllers.py
```diff
@@ -1491,6 +1491,7 @@ def _html(self, r, **attr):
         s3 = current.response.s3
 
         output = {}
+        output["create_post_form"] = False
```
flavour: This shouldn't be necessary.
The controller now uses the create_post_button variable here (since this is now a button not a form):
https://github.com/dhornbein/eden/blob/90a5049898e61c9e8add64e7dc98ecd2925238c7/modules/templates/WACOP/controllers.py#L901

Looks like I updated the incident_profile.html view but not the dashboard view, so the correct solution is to update the dashboard view like the incident_profile already has been:
https://github.com/sahana/eden/blob/master/modules/templates/WACOP/views/incident_profile.html#L112

No need for the if...it is "" if not present

modules/templates/WACOP/views/dashboard.html
```diff
@@ -1,80 +1,84 @@
 {{extend "layout.html"}}
-<div class='row panel'>
- {{=XML(map)}}
+<div class='row'>
+ <div class="small-12 column">
```
flavour: Wow, a lot of diff showing on GitHub to look throughy but seems just this added twice ;)
Ah well, I guess notning we can do to make that easier

flavour: should this not be 'columns' not just 'column'? (& yes ' not " pls)

flavour:
I'll cleanup the issues post-merge

flavour:
I left it as 'column' btw as not 100% this is a bug

PR #1351, opened 2017-04-03
Traducao

whanderley:
add some pt-br translations presents at home page


flavour:
Thankyou, unfortunately this is very hard to review.
is it possible for you to resubmit with a single commit & with the new translation file alphasorted like the original?
i.e. All Capital letters 1st & then lower-case after that.
This way I can review it more easily...many thanks!

whanderley:
I'm sorry! Apparently editing the locales via the web interface is not a good idea. I will make a new pull request.

PR #1352, opened 2017-04-05
add some pt-br translations presents at home page

whanderley:


languages/pt-br.py
```diff
@@ -2651,7 +2654,7 @@
 'Organization deleted': 'Organização excluída',
 'Organization updated': 'Organização atualizada',
 'Organization': 'Organização',
-'Organizations': 'Organizações',
+'Organisations': 'Organizações',
```
flavour: This looks like a mistake in your version of the code.
Default strings in code should always be en_US - so Organizations
Organisations is an en_GB translation

flavour: OK, I see this is a bug in current code, will fix

flavour: Done: https://github.com/sahana/eden/commit/50f2abbb03f845ab291fc17d68249e28acd01413

flavour:
Great, many thanks, was easy to review this time so merged & reverted that incorrect fix :)

PR #1357, opened 2017-05-11
Map CAP area display options

biplovbhandari:
for issue https://github.com/sahana/SAMBRO/issues/74

### DIFF
--- a/modules/s3db/cap.py
+++ b/modules/s3db/cap.py
@@ -2097,7 +2097,9 @@ current.T("This combination of the 'Event Type', 'Urgency', 'Certainty' and 'Sev
     @staticmethod
     def cap_area_location_onaccept(form):
         """
-            Link alert_id for non-template area
+            - Link alert_id for non-template area
+            - for external alerts (from import feed or rss), make sure the
+              locations are only imported if we set polygons to be imported
         """
 
         form_vars = form.vars
@@ -2117,7 +2119,15 @@ current.T("This combination of the 'Event Type', 'Urgency', 'Certainty' and 'Sev
         if alert_id:
             # This is not template area
             # NB Template area are not linked with alert_id
-            db(db.cap_area_location.id == form_vars["id"]).update(alert_id = alert_id)
+            location_default = current.deployment_settings.get_cap_area_default()
+            if "polygon" not in location_default:
+                table = current.s3db.cap_alert
+                row = db(table.id == alert_id).select(table.external,
+                                                      limitby=(0, 1)).first()
+                if row.external:
+                    db(db.cap_area_location.id == form_vars["id"]).delete()
+            else:
+                db(db.cap_area_location.id == form_vars["id"]).update(alert_id = alert_id)
 
     # -------------------------------------------------------------------------
     @staticmethod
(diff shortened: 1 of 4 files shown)


flavour:
I find this rather unclear I'm afraid. I think the deployment_setting could use a summary of what it does in the docstring...can you add that?

biplovbhandari:
@flavour please review

modules/s3db/cap.py
```diff
@@ -2142,36 +2151,47 @@ def cap_area_tag_onaccept(form):
...
+            # code since polygon change over time, even if the code remains
+            # the same. Hence the historic polygon is excluded as (gtable.end_date == None)
+            tquery = (ttable.tag == same_code) & \
+                     (ttable.value == form_vars.get("value")) & \
+                     (ttable.deleted != True) & \
+                     (ttable.location_id == gtable.id) & \
+                     (gtable.end_date == None) & \
+                     (gtable.deleted != True)
+            trows = db(tquery).select(ttable.location_id)
+            for trow in trows:
```
flavour: I know this is part of old code, not the change, but we would never expect to see multiple rows here would we? In which case a limitby=(0,1) will improve performance noticeably.

modules/s3db/cap.py
```diff
@@ -2097,7 +2097,9 @@ def cap_area_onvalidation(form):
     @staticmethod
     def cap_area_location_onaccept(form):
         """
-            Link alert_id for non-template area
+            - Link alert_id for non-template area
+            - for external alerts (from import feed or rss), make sure the
+              locations are imported from the given setting
```
flavour: locations are only imported if we set polygons to be imported

modules/s3db/cap.py
```diff
@@ -2117,7 +2119,14 @@ def cap_area_location_onaccept(form):
         if alert_id:
             # This is not template area
             # NB Template area are not linked with alert_id
-            db(db.cap_area_location.id == form_vars["id"]).update(alert_id = alert_id)
+            table = current.s3db.cap_alert
+            location_default = current.deployment_settings.get_cap_area_default()
+            row = db(table.id == alert_id).select(table.external,
+                                                  limitby=(0, 1)).first()
+            if row.external and "polygon" not in location_default:
```
flavour: The DB call here is only required if polygon not in location_default...so move it inside that branch to improve performance for the cases where polygon is in location_default

modules/s3db/cap.py
```diff
@@ -2142,36 +2151,47 @@ def cap_area_tag_onaccept(form):
         arow = db(atable.id == area_id).select(atable.alert_id,
                                                limitby=(0, 1)).first()
         alert_id = arow.alert_id
+        table = current.s3db.cap_alert
+        location_default = current.deployment_settings.get_cap_area_default()
+        row = db(table.id == alert_id).select(table.external,
+                                              limitby=(0, 1)).first()
+        import_location = False
+        if row.external:
+            if "geocode" in location_default and tag and same_code:
```
flavour: The DB call here is only required if geocode in location_default...so move it inside that branch to improve performance for the cases where geocode is not in location_default

modules/s3db/cap.py
biplovbhandari: 👍

modules/s3db/cap.py
biplovbhandari: 👍

modules/s3db/cap.py
biplovbhandari: 👍

biplovbhandari:
@flavour Please merge

modules/s3db/cap.py
```diff
@@ -2117,7 +2119,15 @@ def cap_area_location_onaccept(form):
...
             # This is not template area
             # NB Template area are not linked with alert_id
-            db(db.cap_area_location.id == form_vars["id"]).update(alert_id = alert_id)
+            location_default = current.deployment_settings.get_cap_area_default()
+            if "polygon" not in location_default:
+                table = current.s3db.cap_alert
+                row = db(table.id == alert_id).select(table.external,
+                                                      limitby=(0, 1)).first()
+                if row.external:
+                    db(db.cap_area_location.id == form_vars["id"]).delete()
```
flavour: Would be good to add a comment on why here

modules/s3db/cap.py
```diff
@@ -2117,7 +2119,15 @@ def cap_area_location_onaccept(form):
...
+            else:
+                db(db.cap_area_location.id == form_vars["id"]).update(alert_id = alert_id)
```
flavour: More importantly, should this not run when polygon not in location_default but row.external is False?

modules/s3db/cap.py
```diff
@@ -2142,36 +2152,47 @@ def cap_area_tag_onaccept(form):
...
+                row = db(table.id == alert_id).select(table.external,
+                                                      limitby=(0, 1)).first()
+                if row.external:
+                    import_location = True
+        else:
+            # For system alert
+            import_location = True
+
+        if import_location:
+            # SAME tag referes to some location_id in CAP
```
flavour: typo: refers

PR #1359, opened 2017-05-16
Map CAP area display options

biplovbhandari:
@flavour Please merge


modules/s3db/cap.py
```diff
@@ -2117,7 +2119,24 @@ def cap_area_location_onaccept(form):
...
+                    # if it is a external alert (coming from RSS and import_feed),
+                    # don't use this for displaying map in area
+                    db(db.cap_area_location.id == form_vars["id"]).delete()
+                else:
+                    # If "polygon" is not in deployment_settings (eg. in PH),
+                    # but if it is generated from SAMBRO, link area to the alert_id
+                    db(db.cap_area_location.id == form_vars["id"]).update(alert_id = alert_id)
+            else:
+                # Normal case
+                db(db.cap_area_location.id == form_vars["id"]).update(alert_id = alert_id)
```
flavour: Not so critical, but I would DRY these last 2 by having a return after the .delete()
& then having the .update() outside of the if (no need for any else)

modules/s3db/cap.py
```diff
@@ -2136,20 +2155,43 @@ def cap_area_tag_onaccept(form):
...
+                                                          limitby=(0, 1)).first()
+                    if row.external:
+                        # If this is external alert, check the deployment_settings
+                        # to see if location should be imported from "geocode"
+                        if "geocode" in current.deployment_settings.get_cap_area_default():
+                            import_location = True
+                    else:
+                        # For internal alert, import the location
+                        # No need to check the deployment_settings
+                        import_location = True
```
flavour: I still think we're doing an unnecessary DB check here in some branches.
I would do:
if "geocode" in current.deployment_settings.get_cap_area_default():
    import_location = True 
else:
    # check for internal

biplovbhandari: @flavour but this "geocode" check is only while importing the alerts from external feeds (i.e. external alerts). Plus if there is no tag or same_code, we don't need to import as well.

flavour: > this "geocode" check is only while importing the alerts from external feeds (i.e. external alerts)

Yes, however if geocode is in the setting (quick lookup) then no need to check for internal/external...we always import_location=true (according to this logic anyway)

> if there is no tag or same_code, we don't need to import as well.

That is already covered by the 1st import_location = False

biplovbhandari:
@flavour Please merge

PR #1366, opened 2017-07-07
WACOP map, alerts, tabs, and clearfix

dhornbein:
Style all tabs on profile pages

Added tab datalists to browse pages

Added staff and org tabs to top level

resource page clearfix

Set map min height to 400

Minimum height on alert and event boxes content


flavour:
Moving all the tabs to the top-level instead of nested is a definite improvement I think :)
The nested was clever but confusing...

PR #1380, opened 2017-09-19
AAA: Make messages localizable

trendspotter:
This PR aims to make AAA messages and mails localizable in the easiest way possible.


flavour:
There is no need for this: the Messages structure ia already localisable by default

PR #1383, opened 2017-09-21
AAA: Fix approver mail localization

trendspotter:
Slightly related to the mailing list [topic about AAA localization](https://groups.google.com/forum/#!topic/sahana-eden/EIX2nqkWcEc).

The mails sent to the registered users are translated correctly, but the mails for approvers are not. This is caused by the lazy evaluation of `T()` function, so when the translation language is forced via `T.force()`, it needs to be evaluated before the language changes again, otherwise it always evaluates to the current language.

The additional `s3_str()` call in this PR ensures the evaluation before the language is changed.

### DIFF
--- a/modules/s3/s3aaa.py
+++ b/modules/s3/s3aaa.py
@@ -2327,25 +2327,25 @@ $.filterOptionsS3({
             T.force(language)
             if message == "approve_user":
                 subjects[language] = \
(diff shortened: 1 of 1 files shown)


flavour:
Thankyou, yes, have had to do a couple of those in the past :)

PR #1384, opened 2017-09-21
WACOP: spacing and resource page updates

dhornbein:


flavour:
Mergeed but I have no idea what the 'RSS feed goes here is all about?

dhornbein:
Devin requested an RSS feed but didn't specify what it is going to be.

PR #1386, opened 2017-09-23
ORG: Add "Resource Types" to Organisation menu

trendspotter:
Tiny little PR to show *Resource Types* in Organisation menu. Allows browsing, editing, deletion etc.

### DIFF
--- a/modules/s3menus.py
+++ b/modules/s3menus.py
@@ -1551,11 +1551,11 @@ class S3OptionsMenu(object):
                       restrict=[ADMIN])(
                         M("Create", m="create"),
                     ),
-                    M("Resource Types", f="resource_type",
-                      restrict=[ADMIN])(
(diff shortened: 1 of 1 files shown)


modules/s3menus.py
```diff
@@ -1551,6 +1551,10 @@ def org():
                       restrict=[ADMIN])(
                         M("Create", m="create"),
                     ),
+                    M("Resource Types", f="resource_type",
```
nursix: Makes sense since Resources are default as well. But could it go last in the menu?

trendspotter: I've tried to follow the same order as with the actual records using the types. Currently there are *Offices* - *Facilities* - *Resources* in that order, so *Office Types* - *Facility Types* - *Resource Types* made sense. But sure, I can move it to the end.

trendspotter:
Moved to the end as requested.

Issue #1388, opened 2017-09-24
settings.ui.interim_save breaks CAP Alerts


trendspotter:
When using CAP module along with `settings.ui.interim_save = True`, *Alerts* die with following traceback
```
Traceback (most recent call last):
  File "/srv/sambro/gluon/restricted.py", line 227, in restricted
    exec ccode in environment
  File "/srv/sambro/applications/eden/controllers/cap.py", line 1515, in <module>
  File "/srv/sambro/gluon/globals.py", line 417, in <lambda>
    self._caller = lambda f: f()
  File "/srv/sambro/applications/eden/controllers/cap.py", line 1032, in alert
    rheader = s3db.cap_rheader,
  File "/srv/sambro/applications/eden/models/00_utils.py", line 243, in s3_rest_controller
    output = r(**attr)
  File "applications/eden/modules/s3/s3rest.py", line 691, in __call__
    output = self.resource.crud(self, **attr)
  File "applications/eden/modules/s3/s3rest.py", line 1774, in __call__
    output = self.apply_method(r, **attr)
  File "applications/eden/modules/s3/s3crud.py", line 110, in apply_method
    output = self.select(r, **attr)
  File "applications/eden/modules/s3/s3crud.py", line 1239, in select
    form = self.create(r, **attr).get("form", None)
  File "applications/eden/modules/s3/s3crud.py", line 372, in create
    self._interim_save_button()
  File "applications/eden/modules/s3/s3crud.py", line 3130, in _interim_save_button
    settings.custom_submit.insert(0, item)
AttributeError: 'tuple' object has no attribute 'insert'
```

This is caused by the fact that *Alerts* define their own `custom_submit` which is immutable (a tuple). I see two approaches.

1. Either make the `custom_submit` mutable (list instead of tuple), which I have in [21beef1](https://github.com/trendspotter/eden/commit/21beef16742ca9d53b890547f44da70d8d2b0cdb)
2. Or, since the *Save and edit information* button has pretty much the same functionality as interim save, there can be a check in S3CRUD which determines if the  `custom_submit` is mutable (resp. has *insert* property) and if it's not, it doesn't add the interim save button. This I have in [6523bba](https://github.com/trendspotter/eden/commit/6523bba4d502a16951d83876a58925a5bac4c78d).

Which one would you like to see as PR?

nursix:
Both, sort-of.

[...]

PR #1389, opened 2017-09-24
S3CRUD: Allow immutable custom_submits to have interim save button

trendspotter:
PR based on issue #1388. Changes *custom_submit* from tuple to list and also makes S3CRUD always convert it to list, which allows `settings.ui.interim_save` to modify the *custom_submit*.

The S3CRUD conversion in commit [101ff4d](https://github.com/trendspotter/eden/commit/101ff4d4c8557c6a8ef9b23ba291d692ce1a6d1b) might not be needed anymore as *custom_submit* is only in CAP module and CRMT template, so feel free to cherry-pick whatever seems the best.

### DIFF
--- a/modules/s3/s3crud.py
+++ b/modules/s3/s3crud.py
@@ -3126,10 +3126,10 @@ class S3CRUD(S3Method):
         else:
             return
         settings = current.response.s3.crud
+        custom_submit = [item]
         if settings.custom_submit:
-            settings.custom_submit.insert(0, item)
-        else:
-            settings.custom_submit = [item]
+            custom_submit.extend(settings.custom_submit)
+        settings.custom_submit = custom_submit
         return
 
     # -------------------------------------------------------------------------


modules/s3/s3crud.py
```diff
@@ -3127,6 +3127,7 @@ def _interim_save_button():
             return
         settings = current.response.s3.crud
         if settings.custom_submit:
+            settings.custom_submit = list(settings.custom_submit)
```
nursix: Actually, there is a better strategy here - see my comment in your other commit.
https://github.com/trendspotter/eden/commit/6523bba4d502a16951d83876a58925a5bac4c78d#diff-33fc0b179125dd7a80339eede2995c49R3131

Sorry, wasn't quick enough

nursix: My point there is that we don't actually want to extend the setting with the [item], but the [item] with the setting ;)

That way, we always know that it is mutable, and we spare us the explicit conversion to a list and the extra assignment of a dict property.

And then, implementations can do what they want - mutable or immutable doesn't matter. You can basically revert the cap.py changes: it's okay to set a tuple.

nursix: (And under the hood: instantiating tuples is a lot faster ;) - so cap.py setting tuples is a good thing, and performance in CRUD isn't worse with .extend than it would be with .insert)

nursix:
Yeah - that looks good now.

Now, as an exercise - could you please squash the commits into one?
http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/Git#Squashingcommits

nursix:
Easiest way is probably to soft-reset to the commit before the first in this PR, then commit again, and then force-push (this will update the PR accordingly).

Thanks.

trendspotter:
GitHub allows squashing when accepting PR in the droprown menu. But OK, lets's see how it works if I try to squash them on my side.

trendspotter:
About the comment "*cap.py setting tuples is a good thing*". Should I revert the changes for CAP and CRMT and leave just S3CRUD?

flavour:
> GitHub allows squashing when accepting PR in the dropdown menu

So it does :)
Wonder how long that feature has been there without me noticing ;)

nursix:
Yeah, I know - but I'm not the one to merge it, and Fran would normally request multiple commits on the same issue to be squashed to make it easier to review. Never really an issue, that's why I said "as an exercise" ;)

nursix:
Yes, that's what I said: the changes in cap.py can be reverted - tuples are actually good since faster to instantiate than lists, and the CRUD change is performance-neutral yet makes it robust for typles.

nursix:
Perfect :) Sorry for the extended conversation - I'm a bit slow thinking on Sunday mornings. How do you manage? *lol*

PR #1395, opened 2017-09-27
Edu: Add school capacity and contact to XSLT and CSV

trendspotter:
Adds the possibility to define school capacity and contact in imports, since they exist in the *Create* form.

I understand the idea of imports and validation conceptually, but not yet implementation-wise, so I hope I got this one right. (Tests suggest that I did.)


flavour:
Yes, very easy for simple fields like this :)
Good to add to the top of the XSLT too, which is a place for documentation, but can come with anything else you push :)

PR #1396, opened 2017-09-27
Twitter: Allow creation/deletion via admin

trendspotter:
As briefly discussed on [google groups](https://groups.google.com/forum/#!topic/sahana-eden/Lr9M5z8tdbo), this PR
- Allows Twitter accounts to be created/edited/deleted via admin interface
- ~~Allows tweets to have 280 chars (new feature test-driven by Twitter nowadays)~~
- Smuggles forgotten documentation for #1395
- Updates TwitterSearch requirement to version 1.0+

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN apt-get update && apt-get install -y build-essential unzip curl python-pip p
 	python-numpy python-dateutil python-gdal python-yaml python-serial python-xlwt python-shapely python-pil python-gdal \
 	python-reportlab python-reportlab-accel python-tweepy python-xlrd python-pyth python-boto ansible 
 
-RUN pip install selenium\>=2.23.0 sunburnt\>=0.6 TwitterSearch\>=0.78.4 requests\>=2.3.0
+RUN pip install selenium\>=2.23.0 sunburnt\>=0.6 TwitterSearch\>=1.0 requests\>=2.3.0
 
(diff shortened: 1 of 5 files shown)


flavour:
Seems a little early to add the 280 char support if still being test-driven?
Perhaps have this commented ready to go?

Also this will trigger a database migration, which may not be automatic for some databases.
Ideally test that for all 3 supported dbs & come with a migration script for any platform which it doesn't work automatically for (the whole process of migration scripts isn't supported well currently...we do have some back-end support in s3migration.py & plan to build front-end support into WebSetup but this will focus on moving between stable releases: 1.0->1.1 etc)
This case /might/ be straightforward, but good practise to be aware of it

trendspotter:
OK, not really *test driven* but slowly *rolled out to public*. There already are accounts tweeting with the new length, see also related [post on Twitter blog](https://blog.twitter.com/official/en_us/topics/product/2017/Giving-you-more-characters-to-express-yourself.html).

Anyway, for the time being, I've rolled back those changes and I'll push them again when I have the opportunity to test. Could you please give me some hints about the migration (either here or in the Google Groups thread)? Thanks.

trendspotter:
After some more tests it seems that the current 140 char limit in code can cope with longer messages fairly well, so let's revisit that once it's really needed.

Instead, I have reworked the code around TwitterSearch to support version 1.0 and newer. Works the same, but uses mostly PEP-8 compliant function and property names. The other solution would be to limit the requirements to
```
TwitterSearch>=0.78.4,<1.0
```
but since the changes in code are minor, I figured it would be better not to introduce unnecessary legacy dependencies but to update instead.

flavour:
Thankyou :)

PR #1399, opened 2017-09-28
Vehicles: Remove obsolete Vehicle Types

trendspotter:
PR for removal obsolete, duplicit and slightly confusing code around *Vehicle Types* as briefly discussed on [Google Groups](https://groups.google.com/forum/#!topic/sahana-eden/yQzpvlTIPNU). 

- Removes `vehicle_vehicle_type` database table as the functionality is now in `supply_item`
- Removes also CSV and XSLT for the `vehicle_type`
- Updates relevant menu items

This PR effectively makes clear that *Vehicles* is basically only a "view" to *Assets* and *Supply* modules with some extra vehicle-related info. It's rather big PR for my current skills with SE, so I'd like to ask for additional code review. My tests via GUI and import forms shown that it works. Once we're done with the review, I'll squash the commits, if needed.

**Note:** There is a template *ARC* which uses the *Vehicles* module. [This line](https://github.com/sahana/eden/blob/master/modules/templates/ARC/menus.py#L351) is fairly easy for me to fix, but I don't know how to handle [this one](https://github.com/sahana/eden/blob/master/modules/templates/ARC/config.py#L831). Can you help me out here, please?


trendspotter:
Updated and squashed.  
The PR now also resolves conflicts introduced for vehicle imports in *ARC* template (Actually, most of them were magically resolved by themselves simply by removing the old stuff).  
The *represents* line referenced in the question above is no longer needed, as the field doesn't exist and the `supply_item`-provided type is represented nicely by default.

flavour:
Thanks for doing this, however whilst it's true that Vehicles are a specialised type of Asset, and hence a Supply Item, the Supply Item Categories don't contain all the specialised details that Vehicle Types do, nor would I expect them to do so.
I have pushed a fix sufficient for the old demo ARC template: https://github.com/sahana/eden/commit/dddcd0c3005362a76dc83ffc4d7247c0445bf4e4

trendspotter:
The point of this PR was to fix the vehicle modules relations. The need for modification of ARC template was just a side effect triggered by that change. The original problem mentioned in the [Google groups thread](https://groups.google.com/forum/#!topic/sahana-eden/yQzpvlTIPNU) still remains unresolved.

Just to reiterate:
Currently, the *Vehicle Types* menu link leads to `/eden/vehicle/vehicle_type`, however when I'm creating a new vehicle record manually, the options I'm given in the *Vehicle Type* dropdown and the *Create Vehicle Type* link are using `/eden/vehicle/item` instead. This was the *duplicit and slightly confusing* part I mentioned above.

> Supply Item Categories don't contain all the specialised details that Vehicle Types do

I don't think so. That's why you have `supply_item_category.is_vehicle` flag in supply/assets. When an item is in category which has this flag set, the specialized details you're talking about are show on a separate *Vehicle Details* tab in asset detail. 

It seems that you've decided to migrate the vehicles to assets completely and this PR was just to clean up the remaining skeleton in the closet.

PR #1401, opened 2017-10-05
DVI: Fix recreq represent

trendspotter:
A bug exists where creating a *Recovery request* and subsequently linking the request to a *Dead body* throws following error 500 while visiting `/eden/dvi/body/1/read`

```
  File "/srv/sahana/gluon/restricted.py", line 227, in restricted
    exec ccode in environment
  File "/srv/sahana/applications/eden/controllers/dvi.py", line 289, in <module>
  File "/srv/sahana/gluon/globals.py", line 417, in <lambda>
    self._caller = lambda f: f()
  File "/srv/sahana/applications/eden/controllers/dvi.py", line 153, in body
    return s3_rest_controller(rheader=rheader)
  File "/srv/sahana/applications/eden/models/00_utils.py", line 243, in s3_rest_controller
    output = r(**attr)
  File "applications/eden/modules/s3/s3rest.py", line 691, in __call__
  File "applications/eden/modules/s3/s3rest.py", line 1774, in __call__
  File "applications/eden/modules/s3/s3crud.py", line 103, in apply_method
    output = self.read(r, **attr)
  File "applications/eden/modules/s3/s3crud.py", line 649, in read
    format=representation)
  File "applications/eden/modules/s3/s3forms.py", line 516, in __call__
    buttons = buttons)
  File "/srv/sahana/gluon/sqlhtml.py", line 1399, in __init__
    table = self.createform(xfields)
  File "/srv/sahana/gluon/sqlhtml.py", line 1420, in createform
    newrows = formstyle(id, a, b, c)
  File "applications/eden/modules/s3theme.py", line 144, in formstyle_foundation
    return render_row(row_id, label, widget, comment, hidden)
  File "applications/eden/modules/s3theme.py", line 127, in render_row
    submit = widget.element("input", _type="submit")
TypeError: 'NoneType' object is not callable
```

The problem lies in the representation of `recreq`. Before this PR, the request was [represented by its reference ID](https://github.com/sahana/eden/blob/8386145559fa8e4aad4208487d419b2d36c8ded0/modules/s3db/dvi.py#L154). Since the ID is of type `pydal.helpers.classes.Reference` which implements `__getattr__` eventually returning `Null`, the [condition checking if a widget variable has an 'element' attribute](https://github.com/sahana/eden/blob/8386145559fa8e4aad4208487d419b2d36c8ded0/modules/s3theme.py#L126) wrongly passed and subsequent processing resulted in the traceback above.

This PR adds a custom textual representation, [same as shown during *Dead Body* creation](https://github.com/sahana/eden/blob/8386145559fa8e4aad4208487d419b2d36c8ded0/modules/s3db/dvi.py#L153).

Additionally, it removes a duplicit label settings for `dvi_recreq_id()`.

### DIFF
--- a/modules/s3db/dvi.py
+++ b/modules/s3db/dvi.py
@@ -147,11 +147,15 @@ class S3DVIModel(S3Model):
                                  ])
 
         # Reusable fields
+        represent = S3Represent(lookup="dvi_recreq",
+                                fields = ["marker", "date", "bodies_found"],
+                                labels = T("[%(marker)s] %(date)s: %(bodies_found)s bodies"),
+                                )
         dvi_recreq_id = S3ReusableField("dvi_recreq_id", "reference %s" % tablename,
                                         requires = IS_EMPTY_OR(IS_ONE_OF(db,
                                                         "dvi_recreq.id",
-                                                        "[%(marker)s] %(date)s: %(bodies_found)s bodies")),
-                                        represent = lambda id: id,
+                                                        represent)),
+                                        represent = represent,
                                         label=T("Recovery Request"),
                                         ondelete = "RESTRICT")
 
@@ -190,7 +194,7 @@ class S3DVIModel(S3Model):
         morgue_id = S3ReusableField("morgue_id", "reference %s" % tablename,
                                     requires = IS_EMPTY_OR(IS_ONE_OF(db,
                                                     "dvi_morgue.id", "%(name)s")),
-                                    represent = self.morgue_represent,
+                                    represent = S3Represent(lookup="dvi_morgue"),
                                     ondelete = "RESTRICT")
 
         # CRUD Strings
@@ -227,7 +231,7 @@ class S3DVIModel(S3Model):
                      self.pr_pe_label(requires = [IS_NOT_EMPTY(error_message=T("Enter a unique label!")),
                                                   IS_NOT_ONE_OF(db, "dvi_body.pe_label")]),
                      morgue_id(),
-                     dvi_recreq_id(label = T("Recovery Request")),
+                     dvi_recreq_id(),
                      s3_datetime("date_of_recovery",
                                  label = T("Date of Recovery"),
                                  empty=False,
@@ -477,21 +481,5 @@ class S3DVIModel(S3Model):
                            title = c_title,
                            tooltip = c_comment,
                            )
-
-    # -------------------------------------------------------------------------
-    @staticmethod
-    def morgue_represent(id):
-
-        if not id:
-            return current.messages["NONE"]
-
-        db = current.db
-        table = db.dvi_morgue
-        row = db(table.id == id).select(table.name,
-                                        limitby=(0, 1)).first()
-        try:
-            return row.name
-        except:
-            return current.messages.UNKNOWN_OPT
 
 # END =========================================================================


modules/s3db/dvi.py
```diff
@@ -494,4 +494,20 @@ def morgue_represent(id):
...
+
+        if not id:
+            return current.messages["NONE"]
+
+        db = current.db
+        table = db.dvi_recreq
+        row = db(table.id == id).select(table.marker,
+                                        table.date, table.bodies_found).first()
+        try:
+            return "[%s] %s: %s bodies" % (row.marker, row.date, row.bodies_found)
```
flavour: This format looks nice, however it isn't i18n, of course.
Nor is it a scaleable S3Represent
Are you able to enhance in these 2 ways?

trendspotter: > it isn't i18n

True, however [this one](https://github.com/sahana/eden/blob/8386145559fa8e4aad4208487d419b2d36c8ded0/modules/s3db/dvi.py#L153) wasn't either. I guess wrapping in it `T()` should do the trick in both cases.

> Nor is it a scaleable S3Represent

I've just copied and adjusted `morgue_represent()` method, so I take it that it's not a proper one either. Will try to modify both, if I figure out how :)

nursix:
```
S3Represent(lookup = "dvi_recreq", 
                      fields = ["marker", "date", "bodies_found"], 
                      labels = T("[%(marker)s] %(date)s: %(bodies_found)s bodies"),
                      )
```

Something like that.

nursix: ...solves both, i18n and S3Represent.

Morgue_represent is even simpler:
```
S3Represent(lookup="dvi_morgue")
```
...the rest is default anyway.

nursix: Oh, forgot to say - S3Represent is configured during the table definition, like:
```
Field("some_field", represent = S3Represent(...))
```

...so no functions required here at all.

modules/s3db/dvi.py
```diff
@@ -147,11 +147,15 @@ def model(self):
                                  ])
 
         # Reusable fields
+        RECREQ_LABEL = s3_str(T("[%(marker)s] %(date)s: %(bodies_found)s bodies"))
```
trendspotter: I'm not sure is this is correct, but when I put `T()` into `requires`, it doesn't show the string in dropdown at all and when I put `T()` to `S3Represent`'s `labels`, it gets ignored and builds the label from `fields` instead.

nursix: This is fine, but you should use the /same/ S3Represent in both requires and represent. Like this:

```
represent = S3Represent(...)
...requires = IS_EMPTY_OR(IS_ONE_OF(db, "dvi_recreq.id", represent))
...represent = represent
```

nursix: (Obviously, I can fix S3Represent to accept T for labels - but that will take a moment)

nursix: Done:
https://github.com/nursix/eden/commit/c972e9ca296562ceda25f319d1d1a4056b89ef91

Now you should be able to omit the s3_str() and just pass the T() for a label template. However, still, you should use S3Represent in both IS_ONE_OF and requires.

nursix: Example:
https://github.com/nursix/eden/blob/master/modules/s3db/dvr.py#L555

trendspotter:
Updated to work with nursix/eden@c972e9c. This whole S3 thingy seems neat, but there is too much going on in the background so for people unacquainted with the code, it seems like witchcraft. Little by little, example by example it's getting clearer . Is there any documentation for all these `S3Whatever` classes other than docstrings?

nursix:
Some: http://eden.sahanafoundation.org/wiki/S3
However, examples in core modules (pr, org, gis, hrm, dvr) are a better source.

S3 represents the generic standard functionality of Sahana Eden, and it is meant to reduce efforts for implementing that standard functionality for new models (RAD).

Beyond the S3* programming interfaces lies the more complex stuff that use-case development should be able to treat as black box, ignoring how exactly things happen, so that you can focus on getting the models and UI work flows right.

When you do things the S3 way, you can almost always expect that to fulfill our standard functional requirements without needing to be aware of every detail about those requirements - let alone having to implement them again and again.

S3Represent, in particular, is one of the examples where our standard requirements are complex enough to make it nearly impossible to implement new models quickly, especially when you are new to the code base. 

Therefore, using S3Represent saves you a lot of headaches ;)

nursix:
Btw, the core principle of S3 is introspection - and that is what makes it sometimes hard to figure out why things happen the way they do: S3 function calls seem to lack many parameters, so seemingly you have little control.

Awareness of the introspective principle helps, though: those "missing" or "hidden" parameters are, in the vast majority of cases, either 
- taken from configuration settings (s3db.configure or deployment_settings), or
- derived directly from the data model, or
- looked up from the global request object (i.e. the HTTP request context)

And there's where you can influence the behavior.

To some extent, you can sometimes override parameter sources (or defaults) at function call - but the general idea is indeed that S3 shall find the right parameters itself.

Admittedly, that can lead to confusion if you are unaware of those parameter sources - but once you are acquainted with those, it will quickly loose it's "witchcraft" appearance ;)

PR #1402, opened 2017-10-06
Fix pagination for Organisation listing on homepage

trendspotter:
This PR fixes the problem when *dataTable* with organisation listing located on home page did not paginate properly with more that 40 organisation records.

Cause:
The pagination sends requests with GET variables `start` and `limit` whereas the code tried to read `displayStart` and `pageLength` which didn't exist, thus always fell back to the default values.

### DIFF
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -411,9 +411,9 @@ def organisation():
 
     resource = s3db.resource("org_organisation")
     totalrows = resource.count()
-    display_start = int(get_vars.displayStart) if get_vars.displayStart else 0
-    display_length = int(get_vars.pageLength) if get_vars.pageLength else 10
-    limit = 4 * display_length
+    display_start = int(get_vars.start) if get_vars.start else 0
+    display_length = int(get_vars.limit) if get_vars.limit else 10
+    limit = display_length
 
     list_fields = ["id", "name"]
     default_orderby = orderby = "org_organisation.name asc"
@@ -423,6 +423,8 @@ def organisation():
             orderby = default_orderby
         if query:
             resource.add_filter(query)
+    else:
+        limit = 4 * limit
 
     data = resource.select(list_fields,
                            start=display_start,


nursix:
This also drops the 4x pipeline - which I think should be restored.

controllers/default.py
```diff
@@ -411,9 +411,8 @@ def organisation():
 
     resource = s3db.resource("org_organisation")
     totalrows = resource.count()
-    display_start = int(get_vars.displayStart) if get_vars.displayStart else 0
-    display_length = int(get_vars.pageLength) if get_vars.pageLength else 10
-    limit = 4 * display_length
```
nursix: We should keep this to trigger sspag requests only every 4th page instead of every page

trendspotter: If it's supposed to be every 4th page, then I'd reintroduce
```
limit = 2 * display_length
```

because the pagination itself already doubles the amount of shown records in its request (ie. when I set 25 records per page, i get `limit=50` in GET). Witchcraft again :)

trendspotter: Nope. My bad. I'm trying to push the adjusted values to too many places.

trendspotter: But yes, it does :)

Even when I simplify this whole commit to 
```diff
-    display_start = int(get_vars.displayStart) if get_vars.displayStart else 0
-    display_length = int(get_vars.pageLength) if get_vars.pageLength else 10
+    display_start = int(get_vars.start) if get_vars.start else 0
+    display_length = int(get_vars.limit) if get_vars.limit else 10
```

After first 4 pages, the `limit` variable gets set to double the amount of items per page.

trendspotter:
Simpler fix where just the proper `GET` variables are extracted and the pipeline is retained, however after first 4 pages, the `limit` variable is set to double the amount of items per page, effectively making it 8x pipeline. Not sure why.

controllers/default.py
nursix: That was a misunderstanding - it is correct that the pipeline script retrieves 2x as many records as the user wants to see per page. That wasn't the point here.

This doubling doesn't happen for the 1st request - because it doesn't come from the pipeline script, and for that initial page request it should retrieve 4x as many records as in the initial page length setting, so that no pagination request is triggered unless the user browses past the 4th page (or changes the page length so something bigger than that).

So, the intended fetch length is 4x (initial request) - 2x (after the 4th page) - 2x (after the 6th) - 2x (etc...)

nursix:
Well - why :) because you tell it to :P

...the ```limit = 4*display_length``` should obviously be behind an ```if representation != "aadata"```.

nursix:
By the way - this pagination beyond 40 is really a non-feature ;) 

People who need to go that far to find an organization typically switch to Search after the second or at most third page, so retrieving 40 upfront will effectively prevent org box pagination from ever being used.

That's why you are the first to find it broken - after it's been out there for several years :D real-world user behavior.

trendspotter:
Not sure why would the retrieval of 2x the amount of record in single request be obvious.

Moreover, if the default limit is 10, in `if representation != "aadata"` it'll be always 40 and will work only on the first page load, when the data are loaded as html. As soon as you start filtering, the responses will be already be represented via aadata, which results in "standard" two pages. I don't really see the point of having 4x the amount in that one particular case (other than that "*~~640 K~~ 40 records ought to be enough for anybody*").

nursix:
Well, that's simply to cover the majority of cases without any additional requests.

The point is that in most deployments, the average user can see fewer than 40 organizations - so the initial request can fetch them all. That's by far the majority of cases.

When a user really can manage more than 40 organizations (rare case), then the org box is only useful when they can find the org they want on the first 3-4 pages. If not, then the org box is fairly useless for them anyway - and most deployments (all, actually) that have too many organizations, disable the org box anyway or at most replace it by a search-only feature.

Therefore, fetching all (=up to 40) organizations in the initial page request covers the majority of cases without requiring another pagination request, and that is why that 4x pipeline was there and should remain there.

nursix:
Remember that the org box is intended as a shortcut.

If you need to browse through many pages, or use the Search-feature to find a particular org, then that's actually no longer a shortcut, and then you are better off navigating to org/organisation right away.

And that's exactly how it happens in most actual deployments - few orgs=>org box for faster navigation, many orgs=>org/organisation as first stop.

trendspotter:
Although I understand what you're saying and also understand that the PR resolves pettiness rather than an actual issue, I still think that if the feature is there and is considered default, it should work without any defects, even such which are triggered by edge cases.

My confusion in the implementation stems from inconsistency. *Obvious* would be:  
"Load *n* records in initial request, display *n* records, load *n* records in subsequent requests (one request per page)"

Less obvious but still pretty consistent would be:  
"Load *n×k* records in initial request, display *n* records, load *n×k* records in subsequent requests"

However current behaviour was not obvious as there was again some implicit logic involved (aka *witchcraft*)  
"Load *n×k* records in initial request, display *n* records, load *n×k**×2*** records in subsequent requests"

Since you made clear that those 40 records should be loaded right of the bat, then the solution can be simplified either as this (less `if`s):

```python
    display_start = int(get_vars.start) if get_vars.start else 0
    if get_vars.limit:
        display_length = int(get_vars.limit)
        limit = display_length
    else:
        display_length = 10
        limit = 4 * display_length
```

Or as this (less lines):
```python
    display_start = int(get_vars.start) if get_vars.start else 0
    display_length = int(get_vars.limit) if get_vars.limit else 10
    lengts = int(get_vars.limit) if get_vars.limit else 4 * display_length
```

(Or the [current fix](https://github.com/sahana/eden/pull/1402/commits/9dc74ccc31bed44ca8941bab75e78144b0bf485b) (ab)using `else` branch can remain)

Which one would you prefer?

nursix:
Neither. My point was:

The original version loaded 40 records (4 x default page size) in the initial page request. The idea with that was that this number eliminates the need for any subsequent requests in the majority of use-cases.

The issue you found was that the pagination requests were reading the wrong parameters. You fixed that perfectly, but you also removed that initial loading of 4 pages - apparently as an unintended side effect of your fix.

So I asked you to restore that initial 4-page loading, as it is obviously orthogonal to the problem you intended to fix. There is no need for you to change it, so - don't.

That is in no way pettiness. The person who originally implemented the initial 4-page pipeline had convinced me of the good reasons behind it, so I'm defending it while you have no obvious need to remove it.

controllers/default.py
```diff
@@ -423,6 +423,8 @@ def organisation():
             orderby = default_orderby
         if query:
             resource.add_filter(query)
+    else:
+        limit = 4 * limit
```
nursix: I think this one is good enough - does anything speak against just leaving it at that?

trendspotter:
I got your point. The last question was purely about the best practices used in codebase as all 3 methods achieve the same result (restoring the initial 4 pages), only the code look a little bit different.

Current method may not be obvious at first glance when the `limit = display_length` is 10 lines above the  `else` branch, but it gets the work done, so I'll leave it.

(And by the "pettiness" I meant the pagination issue which I came to originally fix - as you said, it's nothing groundbreaking or hugely beneficial. More or less just a cosmetic issue.)

nursix:
Nah,

I didn't see your fixing of the regression problem in the pagination request (which came about due to the datatables 9=>10 migration) as "pettiness". It is absolutely so that this mistake should be corrected - although the real reason may not be as obvious.

It is indeed so that the pagination beyond 4 pages of orgs is an edge-case - but that only applies to /this/ particular box. This box was though originally invented as a pattern, i.e. its primary purpose (and that of the fac-box too) is to illustrate how such boxes can be implemented - and for the users, how they would look+feel like.

So, think RAD: A deployment could need "My Requests", "My Tasks" or "My Projects" - so what you (or say: the average Sahana developer) would do is to take this default homepage controller, copy+paste it into modules/templates/XYZ/controllers.py, and tweak this function to deliver the resource(s) they need.

And then, it absolutely matters that things are correct with default/index - so your fix isn't pettiness or cosmetic at all, but indeed preventing the proliferation of a regression issue.

Pettiness, though, is my role here ;) - I'm the reviewer.

Issue #1405, opened 2017-10-08
Assigning an asset causes error


trendspotter:
How to reproduce:
1. Create an organsation
2. Create a facility belonging to the organisation above
3. Create an asset (eg. air compressor) belonging to the facility above
4. Open the asset and click on one of *Assign to Person*, *Assign to Facility/Site* or *Assign to Organization*

An error with following traceback will occur:
```
Traceback (most recent call last):
  File "/srv/sahana/gluon/restricted.py", line 227, in restricted
    exec ccode in environment
  File "/srv/sahana/applications/eden/controllers/asset.py", line 217, in <module>
  File "/srv/sahana/gluon/globals.py", line 417, in <lambda>
    self._caller = lambda f: f()
  File "/srv/sahana/applications/eden/controllers/asset.py", line 47, in asset
    return s3db.asset_controller()
  File "applications/eden/modules/s3db/asset.py", line 1291, in asset_controller
    rheader = asset_rheader,
  File "/srv/sahana/applications/eden/models/00_utils.py", line 243, in s3_rest_controller
    output = r(**attr)
  File "applications/eden/modules/s3/s3rest.py", line 691, in __call__
    output = self.resource.crud(self, **attr)
  File "applications/eden/modules/s3/s3rest.py", line 1774, in __call__
    output = self.apply_method(r, **attr)
  File "applications/eden/modules/s3/s3crud.py", line 101, in apply_method
    output = self.create(r, **attr)
  File "applications/eden/modules/s3/s3crud.py", line 392, in create
    format=representation,
  File "applications/eden/modules/s3/s3forms.py", line 516, in __call__
    buttons = buttons)
  File "/srv/sahana/gluon/sqlhtml.py", line 1295, in __init__
    inp = self.widgets.options.widget(field, default)
  File "/srv/sahana/gluon/sqlhtml.py", line 313, in widget
    options = requires[0].options()
  File "/srv/sahana/gluon/validators.py", line 2649, in _options
    options = self.other.options(*args, **kwargs)
  File "applications/eden/modules/s3/s3validators.py", line 1093, in options
    self.build_set()
  File "applications/eden/modules/s3/s3validators.py", line 814, in build_set
    show_link=False)
  File "applications/eden/modules/s3db/org.py", line 5489, in bulk
    labels = self._lookup(values, rows=rows)
  File "applications/eden/modules/s3/s3fields.py", line 762, in _lookup
    rows = self.lookup_rows(key, lookup.keys(), fields=fields)
[...]

PR #1406, opened 2017-10-08
Asset: Fix assignment errors, remove usage of GET variables

trendspotter:
A PR to fix issue reported in #1405

Basically, the variables formerly passed via GET are now implied (hardcoded) based on the method used
- `setbase` = `/asset/asset/1/log/create?status=1`
- `assignperson` = `/asset/asset/1/log/create?status=2&type=person`
- `assignsite` = `/asset/asset/1/log/create?status=2&type=site`
- `assignorg` = `/asset/asset/1/log/create?status=2&type=organisation`

The `type` variable is removed altogether as the intended relation type is now inferred from the method. This also allowed a slight change in the condition logic in `asset_log_onaccept()` where the person assignment is now the default option in `ASSET_LOG_ASSIGN` branch unless explicitly stated otherwise via appropriate method.

I'm not aware of any other module where custom methods would be use in the same fashion, so any advice on how to make it prettier and more obvious are welcome.

### DIFF
--- a/modules/s3db/asset.py
+++ b/modules/s3db/asset.py
@@ -700,8 +700,17 @@ $.filterOptionsS3({
         """
 
         request = current.request
-        get_vars = request.get_vars
-        status = get_vars.get("status", None)
+
+        # Custom methods to allow form customization for specific cases
+        # Original method passed from asset_log_prep()
+        method = current.response.s3.asset_log_method
+        if method == "setbase":
+            status = ASSET_LOG_SET_BASE
+        elif method in ("assignperson", "assignsite", "assignorg"):
+            status = ASSET_LOG_ASSIGN
+        else:
+            status = None
+
         if not status:
             if not current.response.s3.asset_import:
                 # e.g. Record merger or Sync
@@ -711,10 +720,6 @@ $.filterOptionsS3({
             form_vars = form.vars
             asset_id = form_vars.asset_id
             status = int(form_vars.status)
-            if status == ASSET_LOG_ASSIGN:
-                # Only type supported right now
-                # @ToDo: Support more types
-                type == "person"
             new = True
         else:
             # Interactive
@@ -733,7 +738,6 @@ $.filterOptionsS3({
 
             current_log = asset_get_current_log(asset_id)
 
-            type = get_vars.get("type", None)
             log_time = current_log.datetime
             current_time = form_vars.get("datetime", None).replace(tzinfo=None)
             new = log_time <= current_time
@@ -757,7 +761,32 @@ $.filterOptionsS3({
                 db(aitable.asset_id == asset_id).update(location_id = location_id)
 
             elif status == ASSET_LOG_ASSIGN:
-                if type == "person":
+                if method == "assignsite":
+                    asset_tracker.check_in(db.org_site, form_vars.site_id,
+                                           timestmp = request.utcnow)
+                    # Also do component items
+                    locations = asset_tracker.get_location(_fields=[db.gis_location.id])
+                    try:
+                        db(aitable.asset_id == asset_id).update(location_id = locations[0].id)
+                    except:
+                        pass
+
+                elif method == "assignorg":
+                    site_id = form_vars.get("site_id", None)
+                    if site_id:
+                        asset_tracker.check_in(db.org_site, site_id,
+                                               timestmp = request.utcnow)
+                        # Also do component items
+                        locations = asset_tracker.get_location(_fields=[db.gis_location.id])
+                        try:
+                            db(aitable.asset_id == asset_id).update(location_id = locations[0].id)
+                        except:
+                            pass
+                    else:
+                        # We can no longer track location
+                        asset_tracker.check_out()
+
+                else:
                     if form_vars.check_in_to_person:
                         asset_tracker.check_in(db.pr_person, form_vars.person_id,
                                                timestmp = request.utcnow)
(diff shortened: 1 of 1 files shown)


modules/s3db/asset.py
```diff
@@ -700,8 +700,17 @@ def asset_log_onaccept(form):
         """
 
         request = current.request
-        get_vars = request.get_vars
-        status = get_vars.get("status", None)
+
+        # Custom methods to avoid passing GET variables
+        method = current.response.s3.asset_log_method
+        print(method)
```
nursix: Debug code left in

nursix:
I think this works well eliminating the vulnerability caused by GET vars inspection far away from the controller, and prettifying further goes beyond the problem at hand.

Passing parameters from the controller to the onaccept via response.s3 is fragile (still better than inspecting GET vars in the onaccept, though), and requires proper documentation, which you have done.

What's really wrong here is that the entity type of the assignment can't be inferred from the asset log record, which is a glaring weakness of the data model. So, any further improvement should try to solve that, so that the onaccept can inspect the record instead of requiring a hint from the controller.

Aliasing/overriding URL methods, on the other hand, is absolutely a valid pattern - we're using that elsewhere (e.g. deploy), it is allowed and intended in the framework.

nursix:
...and when I say "glaring weakness" then I mean it: 

There is no way to reconstruct the current or any historic status of the asset from the log - because you always need the "type" (explicit or implicit) from the request that created the log entry, which is lost as soon as the call is over.

That's quite a concept flaw if you ask me - what kind of a "log" is that if you can't reconstruct the current/historic statuses from the trail of log entries?

So if you really wanted to fix it, then you should add the entity type to the log record - and then make the onaccept inspect the log trail to determine status and location of the asset.

nursix:
Imagine the situation where the assignment happens in an offline instance, and then you sync two or three log entries to the master - how can you determine the current assignment type?

Or, how can the log answer the question to what or whom the asset was assigned at time X somewhere in the past (and I mean: that's the whole point with a log, isn't it?).

One option could be to assign via super-entities (=key tables), i.e. pe_id and site_id. Then, we apply the simple yet intuitive logic that when no site_id is set, the asset is to be tracked with with the person entity identified by the pe_id (organisation, person, team via pe_id) - otherwise it is located at the site specified by site_id.

trendspotter:
Debug code removed, comments slightly adjusted to be a bit more clear (there's no point of mentioning GET variables if the reader doesn't know the history behind it)

> So if you really wanted to fix it, then you should add the entity type to the log record - and then make the onaccept inspect the log trail to determine status and location of the asset.

That's kinda is what I was thinking by [option 2](https://github.com/sahana/eden/issues/1405#issuecomment-335022906) - ie. not to explicitly *require* any of the values by data model based on which method was used, but instead use generic form and check which fields did the user filled in. If they filled in person, then assign to person, if they filled in site, then assign to site etc. I imagine this may get more complicated when the user sets eg. an organization AND person who has nothing to do with the org, but it *should* still be fairly easy to catch in data model validation.

Anyway I think there's more history behind the assignments then it's visible from the code. E.g. the disabled `ASSET_LOG_RETURN`, so ultimately, some deeper modifications may be in order. But that's stage 2. Stage 1 is to make the things work without errors so we can see what they were supposed to do. And once they do, we can ~~break them again~~ improve them.

nursix:
Sure Sure - you were asking for suggestions how this could be made "prettier", so that is why I made suggestions.

Apart from that, I think this PR resolves the original issue - so stage 1 is good to go ;)

modules/s3db/asset.py
```diff
@@ -1030,17 +1043,16 @@ def asset_log_prep(r):
         table.site_id.requires = IS_ONE_OF(db, "org_site.site_id",
                                            table.site_id.represent)
 
-    elif status == ASSET_LOG_RETURN:
-        crud_strings.msg_record_created = T("Returned")
-        table.person_id.label = T("Returned From")
-        table.person_id.default = current_log.person_id
-        table.site_id.readable = False
-        table.site_id.writable = False
+    # Return functionality removed  - as it doesn't set site_id & organisation_id in the logs
```
flavour: I'm not sure this is a good enough reason to remove it completely?
Surely site_id & organisation_id don't need logging anyway, as these are now back to the original for the asset?

modules/s3db/asset.py
```diff
@@ -700,8 +700,16 @@ def asset_log_onaccept(form):
         """
 
         request = current.request
-        get_vars = request.get_vars
-        status = get_vars.get("status", None)
+
+        # Custom methods to allow form customization for specific cases
+        method = current.response.s3.asset_log_method
```
flavour: I would add a little note:
Read method set in prep

modules/s3db/asset.py
```diff
@@ -982,6 +985,20 @@ def asset_log_prep(r):
 
     table = db.asset_log
 
+    # Custom methods to allow form customization for specific cases
+    # Also passed via response.s3.asset_log_method to asset_log_onaccept()
```
flavour: Lovely :)

flavour:
Thanks for this - some definite enhancements here :)
Please don't remove the ASSET_LOG_RETURN though & a suggestion for slight enhancement of docs.

trendspotter:
That `ASSET_LOG_RETURN` comment has been just copied from [here](https://github.com/sahana/eden/blob/c1da768eb804d55cdc988512cadeeeb650b0ab98/modules/s3db/asset.py#L1143) as the functionality is anyway disabled there. I've disabled it even in the `elif` to not confuse the next person reading that code. The `status` is now set either via custom method or via the form and there are no circumstances under which it could be  `ASSET_LOG_RETURN` . Anyway, I've uncommented it again, but It'll be just another piece of dead code.

Added the comment about `prep` in `onaccept`.

PR #1407, opened 2017-10-13
bugfix for more than 2 info segments in the same alert

biplovbhandari:


modules/s3db/cap.py
```diff
@@ -1920,6 +1896,16 @@ def cap_info_onvalidation(form):
                 used for import from CSV
         """
 
+        T = current.T
+        itable = current.db.cap_info
+        irow = current.db(itable.alert_id == form.request_vars.alert_id).\
```
nursix: Should this not check whether itable.id != form.id? (otherwise updates would always be rejected this way, right?)

controllers/cap.py
```diff
@@ -1192,17 +1184,23 @@ def prep(r):
...
+                                                           atable.event_type_id,
+                                                           limitby=(0, 1)).first()
+                    if row and row.scope == "Public":
+                        fn = "public"
+                    else:
+                        fn = "alert"
+                    itable.web.default = settings.get_base_public_url()+\
+                                         URL(c="cap", f=fn, args=alert_id)
+                    itable.event_type_id.default = row.event_type_id
+                elif len(irows) == 2:
```
nursix: Shouldn't this better be
```
elif len(irows) > 1
```

...otherwise 3 info segments would allow adding more again.

controllers/cap.py
```diff
@@ -758,17 +758,23 @@ def resource_widget(r, **attr):
...
+                                                              table.event_type_id,
+                                                              limitby=(0, 1)).first()
+                        if row and row.scope == "Public":
+                            fn = "public"
+                        else:
+                            fn = "alert"
+                        itable.web.default = settings.get_base_public_url()+\
+                                             URL(c="cap", f=fn, args=alert_id)
+                        itable.event_type_id.default = row.event_type_id
+                    elif len(irows) == 2:
```
nursix: Same here - better check >1 than ==2?

controllers/cap.py
```diff
@@ -758,17 +758,23 @@ def resource_widget(r, **attr):
...
-                        fn = "public"
-                    else:
-                        fn = "alert"
-                    itable.web.default = settings.get_base_public_url()+\
-                                         URL(c="cap", f=fn, args=alert_id)
-                    itable.event_type_id.default = row.event_type_id
-
+                    irows = db(itable.alert_id == alert_id).select(itable.language)
+                    # An alert can contain two info segments, in English and local language
+                    if not len(irows) or len(irows) == 1:
```
nursix: Wondering about this...isn't ```if len(irows) < 2``` exactly the same condition? Would be one len() less.

controllers/cap.py
```diff
@@ -1192,17 +1184,23 @@ def prep(r):
...
+                    row = db(atable.id == alert_id).select(atable.scope,
+                                                           atable.event_type_id,
+                                                           limitby=(0, 1)).first()
+                    if row and row.scope == "Public":
+                        fn = "public"
+                    else:
+                        fn = "alert"
+                    itable.web.default = settings.get_base_public_url()+\
+                                         URL(c="cap", f=fn, args=alert_id)
+                    itable.event_type_id.default = row.event_type_id
```
nursix: Above you're checking for ```if row``` - but here you just access row.event_type_id without checking. Is that correct?

controllers/cap.py
```diff
@@ -758,17 +758,23 @@ def resource_widget(r, **attr):
...
+                                                              table.event_type_id,
+                                                              limitby=(0, 1)).first()
+                        if row and row.scope == "Public":
+                            fn = "public"
+                        else:
+                            fn = "alert"
+                        itable.web.default = settings.get_base_public_url()+\
+                                             URL(c="cap", f=fn, args=alert_id)
+                        itable.event_type_id.default = row.event_type_id
+                    elif len(irows) > 1:
```
nursix: ...and if you check ```if len(irows) < 2``` then this can be ```else``` instead of ```elif```. So saving yet another len().

nursix: That is, there is no case where len(irows) would be <0 ;)

biplovbhandari: 👍

controllers/cap.py
biplovbhandari: Yes. `event_type_id` is a required field. The `if alert` makes sure of `row.event_type_id`. But now I am thinking `row` will never be `None`, so I can avoid that `if row` check?

PR #1409, opened 2017-10-17
bugfix + set country sepecific info segment language

biplovbhandari:


flavour:
If modifying css.cfg should we not also rebuild all the minified CSS?
Otherwise looks fine :)

biplovbhandari:
Actually, the minified CSS already has this CSS. Wondering was the path changed recently?

flavour:
I don't think path changed, but you can check via GitHub history, anyway will merge if you're happy :)

biplovbhandari:
this is working fine for me on my local machine. Also @flavour if I want to use a cascading template, how do I give the template name in the configure script? Now I do SAMBRO first and later manually include the cascade?

flavour:
Not the right place for this query...do you mean Installation script?
Don't know, would need to look at the script to see how it works & potentially modify it

PR #1411, opened 2017-11-08
SAMBRO: shorten alert url

biplovbhandari:


controllers/cap.py
```diff
@@ -1407,6 +1407,15 @@ def notify_approver():
                 subject = "%s: Alert Approval Required" % settings.get_system_name_short()
                 url = "%s%s" % (settings.get_base_public_url(),
                                 URL(c="cap", f="alert", args=[alert_id, "review"]))
+                try:
+                    from pyshorteners import Shortener
```
flavour: Should be added as an optional dependency to requirements.txt

controllers/cap.py
```diff
@@ -1407,6 +1407,15 @@ def notify_approver():
...
+                    try:
+                        url = s3_str(Shortener('Tinyurl', timeout=3).short(url))
+                    except:
+                        pass
+                except ImportError:
```
flavour: Better to do as:
try:
    import
except Import Error
    pass
else:
    try:
        url = ...

flavour: & ideally 2nd except would be more specific...why would it fail? Network issues?

modules/templates/SAMBRO/config.py
```diff
@@ -450,21 +450,28 @@ def onapprove(record):
...
                         current.log.debug("tweepy module needed for sending tweets")
                     else:
+                        url = "%s/%s" % (arow["cap_info.web"], "profile")
+                        try:
+                            from pyshorteners import Shortener
+                            try:
+                                url = Shortener('Tinyurl', timeout=3).short(url)
+                            except:
+                                pass
+                        except ImportError:
```
flavour: Sam try/except/else

controllers/cap.py
biplovbhandari: the exception is ```requests.exceptions.ReadTimeout```. Just do not want to import requests, which would be unused. I would make changes with the structure you mentioned

PR #1412, opened 2017-11-12
link alert_id with message

biplovbhandari:


nursix:
See travis report - not working

modules/s3cfg.py
```diff
@@ -2369,6 +2369,19 @@ def get_msg_notify_attachment(self):
 
         return self.msg.get("notify_attachment")
 
+    def get_msg_alert_id(self):
```
nursix: Misleading name - should be something like ```get_msg_alert_id_getter``` instead so that it's clear that it must be configured with a function.

modules/s3/s3msg.py
```diff
@@ -394,12 +395,21 @@ def send_by_pe_id(pe_id,
                 from_address = current.deployment_settings.get_mail_sender()
 
             table = s3db.msg_email
-            _id = table.insert(body=message,
-                               subject=subject,
-                               from_address=from_address,
-                               #to_address=pe_id,
-                               inbound=False,
-                               )
+            if alert_id:
```
nursix: DRY! This branching shouldn't be necessary (it's quite hard to maintain - shouldn't repeat the entire call here) - maybe use a **data dict where you add alert_id if not default.

modules/s3/s3msg.py
```diff
@@ -414,19 +424,33 @@ def send_by_pe_id(pe_id,
...
             record = dict(id=_id)
             s3db.update_super(table, record)
             message_id = record["message_id"]
         elif contact_method == "TWITTER":
             table = s3db.msg_twitter
-            _id = table.insert(body=message,
-                               from_address=from_address,
-                               inbound=False,
-                               )
+            if alert_id:
```
nursix: DRY!

modules/s3/s3msg.py
```diff
@@ -1468,9 +1492,15 @@ def send_tweet(self, text="", recipient=None):
 
         def log_tweet(tweet, recipient, from_address):
             # Log in msg_twitter
-            _id = table.insert(body=tweet,
-                               from_address=from_address,
-                               )
+            if alert_id:
```
nursix: DRY!

modules/templates/SAMBRO/config.py
```diff
@@ -845,6 +861,23 @@ def custom_msg_notify_attachment(resource, data, meta_data):
...
+            Custom Method to get the alert_id for which the message is sent
+            @param resource: the S3Resource
+            @param data: the data returned from S3Resource.select
+            @param meta_data: the meta data for the notification
+        """
+
+        rows = data["rows"]
+        alert_id = None
+        if len(rows) == 1:
+            alert_id = int(rows[0]["cap_alert.id"])
```
nursix: This can raise various exceptions (e.g. row not containing cap_alert, or it not being int-convertible like None) - and none of them is caught anywhere. That being a scheduler process, it may set the scheduler task to FAILED and never run again.

You should consider to make this more robust.

modules/s3/s3notify.py
```diff
@@ -361,6 +361,12 @@ def send(cls, r, resource):
             if attachment_fnc:
                 document_ids = attachment_fnc(resource, data, meta_data)
 
+        # Alert id
+        alert_id = None
+        alert_id_fnc = settings.get_msg_alert_id()
+        if alert_id_fnc:
+            alert_id = alert_id_fnc(resource, data, meta_data)
```
nursix: Doesn't catch any exceptions in the custom alert_id getter - which may kill the scheduler task, which is high-risk.

And what if it isn't an alert notification at all (e.g. because there are other notifications in the deployment)?

This must be made more robust!

biplovbhandari:
@nursix please review

nursix:
There's still a problem here, I'm afraid.

You added self.cap_alert_id in a number of message models, but cap_alert_id is nowhere in names in s3db/cap.py - nor is there any safe default for cap_alert_id when the module is disabled. This I think must be fixed before it can be merged.

modules/s3db/cap.py
```diff
@@ -308,6 +308,7 @@ class S3CAPModel(S3Model):
     """
 
     names = ("cap_alert",
+             "cap_alert_id",
```
nursix: Ah - there it is now, okay.

modules/s3db/cap.py
```diff
@@ -1607,6 +1608,19 @@ def model(self):
 
     # -------------------------------------------------------------------------
     @staticmethod
+    def defaults():
+        """
+            Return safe defaults in case the model has been deactivated.
+        """
+
+        alert_id = S3ReusableField("alert_id", "integer",
```
nursix: 👍

modules/templates/SAMBRO/config.py
```diff
@@ -845,6 +861,28 @@ def custom_msg_notify_attachment(resource, data, meta_data):
...
+    # -----------------------------------------------------------------------------
+    def custom_msg_alert_id_getter(resource, data, meta_data):
+        """
+            Custom Method to get the alert_id for which the message is sent
+            @param resource: the S3Resource
+            @param data: the data returned from S3Resource.select
+            @param meta_data: the meta data for the notification
+        """
+
+        rows = data["rows"]
```
nursix: Can generally be data.rows (it's actually an attribute not a key), which avoids the otherwise necessary mapping of the key to the attribute. But that won't hold up the merge here - just something you could fix the next time you make changes here.

nursix:
Looks okay from my side now - can be merged unless @flavour finds any issues.

nursix:
Oops, closed accidentally. Sorry.

modules/s3db/msg.py
```diff
@@ -642,6 +642,7 @@ def model(self):
                      # Instance
                      super_link("message_id", "msg_message"),
                      self.msg_channel_id(),
+                     self.cap_alert_id(),
```
flavour: I'm really not keen on adding fields to the core of Messaging for this very-specific usecase.
I would much rather see this be a link table stored in the CAP module

modules/s3db/msg.py
```diff
@@ -1506,6 +1507,7 @@ def model(self):
                           self.super_link("message_id", "msg_message"),
                           self.msg_channel_id(),
                           self.org_organisation_id(default = default),
+                          self.cap_alert_id(),
```
flavour: This can just be a single link table if this is cap_alert_message with alert_id & message_id
The latter can point to any of the message types

modules/s3/s3msg.py
```diff
@@ -386,6 +387,8 @@ def send_by_pe_id(pe_id,
 
         s3db = current.s3db
 
+        alert_id = data.get("alert_id", None)
```
flavour: Again, I'm not keen on contamining the core Messaging API with this very specific usecase
**data as an extension into API args is fine

modules/s3/s3msg.py
```diff
@@ -399,6 +402,7 @@ def send_by_pe_id(pe_id,
                                from_address=from_address,
                                #to_address=pe_id,
                                inbound=False,
+                               alert_id=alert_id,
```
flavour: I think we introduce a new hook, something like:
postp =  settings.get_msg_send_postprocess()
if postp:
    postp(message_id, **data)

modules/s3/s3notify.py
```diff
@@ -361,6 +361,13 @@ def send(cls, r, resource):
             if attachment_fnc:
                 document_ids = attachment_fnc(resource, data, meta_data)
 
+        # Alert id for cap
+        alert_id = None
+        if settings.has_module("cap"):
+            resource_id_getter_fnc = settings.get_msg_resource_id_getter()
+            if callable(resource_id_getter_fnc):
+                alert_id = resource_id_getter_fnc(resource, data, meta_data)
```
flavour: Again, wary of use-case specifics inside a generic API.
So would prefer something like:
send_data = settings.get_msg_notify_data()
send(..., **send_data)

biplovbhandari:
@flavour please merge

modules/s3cfg.py
```diff
@@ -2269,6 +2269,14 @@ def get_msg_basestation_code_unique(self):
         """
         return self.msg.get("basestation_code_unique", False)
 
+    def get_msg_send_postprocess(self):
+        """
+            Custom function that performs some action after post process, eg.
+            link alert_id in cap module to message_id in message module
```
flavour: You should add here the expected API for the function (I know we can search for the usecase via grep, but still)

modules/s3cfg.py
```diff
@@ -2269,6 +2269,14 @@ def get_msg_basestation_code_unique(self):
         """
         return self.msg.get("basestation_code_unique", False)
 
+    def get_msg_send_postprocess(self):
+        """
+            Custom function that performs some action after post process, eg.
```
flavour: not 'after post process' - this is the 'post-process'
Custom function that processes messages after they have been sent" perhaps
- although currently the hook is before the send which risks messages not being sent if the postp crashes...so perhaps better to actually move afterwards too...

modules/s3cfg.py
```diff
@@ -2369,6 +2379,19 @@ def get_msg_notify_attachment(self):
 
         return self.msg.get("notify_attachment")
 
+    def get_msg_notify_send_data(self):
+        """
+            Custom function that returns the data in the s3msg.send_by_pe_id
```
flavour: 'Custom function that returns additional arguments to pass to s3msg.send_by_pe_id'

modules/s3cfg.py
```diff
@@ -2369,6 +2379,19 @@ def get_msg_notify_attachment(self):
...
+
+            The function may be of the form:
```
flavour: The function *should* be of the form

flavour:
This looks much nicer doesn't it? Thankyou...almost ready for merge...a couple of minor tweaks to docs & consider moving the postp hook to after send not just after message creation

biplovbhandari:
yeah looks much cleaner now. I will do the changes in few hours. Have to rush to dinner in some time :)

biplovbhandari:
@flavour Please merge

PR #1413, opened 2017-11-21
bugfix for send_tweet + location update for SAMBRO/NP + remove the copy button: not working

biplovbhandari:


nursix:
create?from_record only works with default form - not implemented for custom forms (cap_info uses a custom form, that's why it doesn't work)

biplovbhandari:
@flavour Please merge

flavour:
If the copy button is wanted, but isn't working yet, I suggest commenting it rather than removing it completely?
Can add the reason why it's not working yet

biplovbhandari:
@flavour Done!

biplovbhandari:
@flavour this also contains the bugfix for the tweet. Can you review and if okay merge?

flavour:
Yup, spotted that - although it didn't seem problematic so wouldn't have mattered if I hadn't as it happens ;)

Thanks :)

biplovbhandari:
I am testing for the Nepal training for the 2nd week of December. Might post some (silly!) questions in the group and some PR. Thanks for the merge :)

PR #1416, opened 2017-12-01
minor tweaks

biplovbhandari:


flavour:
I am wary about changing the default Event Types, as this is a list which came from somewhere official.
If your local users don't like it then create a copy & use that one instead

nursix:
Original source were Canadian EMS incident categories ;)

Funny enough, they use consistently singular terms - as do the rest of the default values here.

I tend to agree with Biplov to change the defaults here - makes more sense IMO.

nursix:
Looking around: FEMA, UN - nobody uses plurals for event/incident type taxonomy. Looks rather strange that we do, and just for those two.

flavour:
ok ;)

PR #1417, opened 2017-12-01
Show shelter workflow message only once when interim_save is enabled

trendspotter:
Cosmetic issue. When `settings.ui.interim_save = True`, the message *"The next screen will allow you to..."* in shelter creation form is displayed twice (once per submit button). This PR forces the responsible jQuery snippets to take into account only the first submit button.


flavour:
Thanks!

PR #1419, opened 2017-12-06
comment the old way of xslt

biplovbhandari:
also wondering how to debug xslt in the eclipse. is that even possible? too much of effort to debug xslt.

### DIFF
--- a/static/formats/cap/export.xsl
+++ b/static/formats/cap/export.xsl
@@ -532,7 +532,7 @@
         exclusive, as we do not create or import a location that is both a
         polygon and a circle. The circle and polygon fallbacks are mutually
         exclusive as only one is constructed in xml_post_render.
-    -->
+
     <xsl:template match="resource[@name='cap_area_location']//resource[@name='gis_location_tag']">
         <xsl:choose>
             <xsl:when test="./data[@field='tag']/text()='cap_circle'">
@@ -557,6 +557,7 @@
             </xsl:when>
         </xsl:choose>
     </xsl:template>
+	-->
 
     <!-- ****************************************************************** -->
     <!-- cap_resource -->


flavour:
This is separate to Dominic's fix, right?

XSLT debugging: http://eden.sahanafoundation.org/wiki/XsltTemplates#Debugging

biplovbhandari:
yes this is separate. this one is for the imported alerts from other source.

static/formats/cap/export.xsl
```diff
@@ -319,12 +319,12 @@
...
             <!-- Resources -->
             <!-- Include all Resources within this Info & all that are global to the Alert -->
-            <xsl:apply-templates select="../resource[@name='cap_resource'][reference[@field='info_id' and @uuid=$uuid]]|
+            <xsl:apply-templates select="../resource[@name='cap_resource'][reference[@field='info_id']]|
                                          ../resource[@name='cap_resource'][not(reference[@field='info_id'])]"/>
                                          
             <!-- Areas -->
             <!-- Include all Areas within this Info & all that are global to the Alert -->
-            <xsl:apply-templates select="../resource[@name='cap_area'][reference[@field='info_id' and @uuid=$uuid]]|
+            <xsl:apply-templates select="../resource[@name='cap_area'][reference[@field='info_id']]|
```
nursix: Err...this looks wrong now.

You want to include all cap_areas which have an info_id or do not have an info_id (second part of the OR is the exact opposite of the first) - so why check for info_id at all? I think it was correct as it was.

nursix: Yeah - has to be: the comment says "include all areas which belong either to /this/ info or are global to the alert", but as you have it, it would always include all areas within the alert for every info segment.

biplovbhandari: Actually, you are right. Sorry for the trouble. Will revert back.

PR #1421, opened 2017-12-08
WACOP: style updates, map fix, and empty address handling

dhornbein:
moved read more on event box

Fixed map on home page

removed event/alert boxes

Conditional address data on incident profile

Missing address handling for alert and event html

Tagit visual improvement

Removed list style from update comments

### DIFF
--- a/modules/templates/WACOP/controllers.py
+++ b/modules/templates/WACOP/controllers.py
@@ -486,12 +486,10 @@ class custom_WACOP(S3CRUD):
                     L3 = row["gis_location.L3"]
                     lat = row["gis_location.lat"]
                     lon = row["gis_location.lon"]
+                    location_full = "%s, %s; %s, %s" % (L3, L1_abrv, lat, lon)
                 else:
                     # No location or national
-                    L1_abrv = ""
-                    L3 = ""
-                    lat = ""
-                    lon = ""
+                    location_full = "No Address Given"
                 alerts.append(DIV(TAG["aside"](TAG["header"](UL(LI(_class="item icon",
                                                                    ),
                                                                 LI(status,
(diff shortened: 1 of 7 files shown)


modules/templates/WACOP/controllers.py
```diff
@@ -683,7 +681,7 @@ def _events_html():
                     # No location
                     L1_abrv = ""
                     L3 = ""
-                    addr_street = ""
+                    addr_street = "No Address Given"
```
flavour: I'll wrap this in T()

modules/templates/WACOP/views/incident_browse.html
```diff
@@ -1,8 +1,6 @@
 {{extend "layout.html"}}
 {{=system_wide}}
 <main role='main' class='main'>
- {{=XML(alerts)}}
- {{=XML(events)}}
```
flavour: if we are removing these from the view, then we should also not bother generating them => saves on server load!

dhornbein: agreed, what/where does that happen? We still need them for other pages.

modules/templates/WACOP/controllers.py
dhornbein: thanks! I take it T() does the translation? I'll keep note of that for the future.

modules/templates/WACOP/views/incident_browse.html
flavour: > what/where does that happen?

See my followup commit: https://github.com/sahana/eden/commit/244d7a185019942cb4d16c6f0ac6d4a11b847bc6

> I take it T() does the translation?

Yes.

PR #1422, opened 2017-12-09
activate hrm for tomorrow and next day demo

biplovbhandari:


flavour:
I don't see why we need this in trunk?
- can just stay in your branch...
Would rather avoid trunk churn.

PR #1427, opened 2018-01-13
 readthedocs.io initial configuration

eduarrom:
Initial configuration generated with sphinx for readthedocs.io and what is sahana? as index description


nursix:
IIRC, we had discussed this before and decided not to move the docs to RTD, 
- for the build process it involves (not meaning the automated build process from GitHub, but the one you would need to  perform in order to preview the results of your doc source changes before you submit them to the repo)
- to not bloat our Git work flow with reviews of minor documentation changes - and,
- to not require doc writers having to install+learn Sphinx and Git in the first place. 

So, before making any more efforts here, this should probably brought up on the mailing list first for renewed discussion + decision making.

PR #1429, opened 2018-01-16
Update es.py

JBarrioL:
Starting to update some of the translations since a few of them arent properly translated.


flavour:
Thanks!

PR #1430, opened 2018-01-17
Update es.py

JBarrioL:


languages/es.py
```diff
@@ -39,63 +39,67 @@
 'Partner Type Details': 'Detalles de Tipos de Socio',
 'Edit Partner Type Details': 'Editar Detalles de Tipos de Socio',
 'Import Partner Types': 'Importar Tipos de Socio',
-'List Partner Types': 'Listar Tipos de Socio',
+'List Partner Types': 'Lista de Tipos de Socio',
```
flavour: No, this is an Action

flavour: & the original translation came from the Red Cross in Latin America

languages/es.py
```diff
@@ -7331,55 +7335,55 @@
...
+'flush latrine with septic tank': 'Desagüe de inodoro con tanque séptico',
 'follow-up assessment': 'Evaluación de seguimiento',
 'forehead': 'frente',
 'form data': 'datos del formulario',
 'found': 'encontrado',
 'from Twitter': 'desde Twitter',
-'from_id': 'rom_id',
-'full': 'completo',
-'getting': 'conseguir',
+'from_id': 'from_id',
```
flavour: If it's exactly the same, then better to remove completely

flavour: (This one was probably a temporary bug introducing it)

languages/es.py
```diff
@@ -7331,55 +7335,55 @@
...
+'full': 'lleno',
+'getting': 'obteniendo, consiguiendo.',
```
flavour: Wary of putting multiple translations in...this word could be compounded into a longer string where it wouldn't make sense.
The usecase that I find in the code isn't like this though...it means 'fetching' (an AJAX request)

flavour:
Thankyou, I can see a lot of good enhancements here :)
I wonder if we need to start a separate file for Latin American Spanish as opposed to Peninsular?
- although this seems rather impossible, instead we really need es-ar.py, etc which then gets hard to manage (es-ar only falls back to es if the es-ar file is empty, so can't just add a few alternates here). I guess managing this would require managing strings in a separate build system & then generating all the different variants...which I guess may be more than you wish to take on ;)
Out of interest, where are you from? & what is your interest in Sahana? Just helping out or actually making use?

xuebozhu:
We are a group of students from Spain. Somehow we were interested in Sahana so we want to contribute in anyway, if it is possible. Greetings from Spain.

JBarrioL:
I wouldn't start a separate file for Latin American Spanish and Peninsular , they can both be used with one file, since the differences shouldn't be that noticeable, and it might be to much work to do separate files though if people start contibuting it might work.

languages/es.py
```diff
@@ -7363,9 +7363,8 @@
 'form data': 'datos del formulario',
 'found': 'encontrado',
 'from Twitter': 'desde Twitter',
-'from_id': 'from_id',
 'full': 'lleno',
-'getting': 'obteniendo, consiguiendo.',
+'getting': 'conseguir.',
```
flavour: Thankyou for the corrections. Why the trailing dot?
Can you squash these commits together in a rebase?
http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/Git#Squashingcommits

JBarrioL: The trailing dot was a mistake.

PR #1431, opened 2018-01-18
SAMBRO AlertHub Sub-Template, 1st cut

waidyanatha:
This pull request is almost the same as : https://github.com/sahana/eden/pull/1428 
Should I be closing it or would you delete/reject it?

- The main change is the creation of a cascade template SAMBRO/AlertHub. It includes all the relevant prepop data.
- A small change was made to the cascade template SAMBRO/Demo by moving the cap_alerting_authorities.csv file inside the Demo folder rather than sitting in the default SAMBRO folder, because this data is strictly for demo purpose.
- the new template and propop data was tested and the working instance is now live here: http://cap.sahana.io:8000/eden/default/index


modules/templates/SAMBRO/Demo/tasks.cfg
```diff
@@ -21,10 +21,19 @@ org,organisation,organisation.csv,organisation.xsl
...
+#
+cap,warning_priority,cap_warning_priority.csv,warning_priority.xsl
+#
+# @note (nuwan) - the three files with demo templates should only load in the demo and not other templates
 cap,alert,cap_alert.csv,alert.xsl
 cap,info,cap_info.csv,info.xsl
 cap,area,cap_area.csv,area.xsl
-cap,warning_priority,cap_warning_priority.csv,warning_priority.xsl
+#
+# @note (nuwan) - all the import_xml files should not load in other templates change to CSV to avoid this issue
```
flavour: Not sure why it matters whether these are CSV or XML?
The issue is which tasks.cfg they appear in & which config.py reference those...

modules/templates/SAMBRO/Demo/tasks.cfg
```diff
@@ -21,10 +21,19 @@ org,organisation,organisation.csv,organisation.xsl
...
 gis,config,gis_config.csv,config.xsl
 # CAP
+# -----------------------------------------------------------------------------
+# Alerting Authority
+# 
+cap,alerting_authority,cap_alerting_authority.csv,alerting_authority.xsl
+#
+cap,warning_priority,cap_warning_priority.csv,warning_priority.xsl
+#
+# @note (nuwan) - the three files with demo templates should only load in the demo and not other templates
```
flavour: Demo isn't really a template...it's a collection of sample data that can be loaded into any SAMBRO template

flavour:
Better, but still 2 commits...each with a commit message which is meaningful to Nuwan but not to other trunk users...we want a single commit with a message meaningful to other trunk users, such as 'SAMBRO AlertHub Sub-Template, 1st cut'

modules/templates/SAMBRO/Demo/tasks.cfg
waidyanatha: the data in the Demo is for training only. It is not relevant to other SAMBRO templates (e.g. PH, MM, MV, NP). For example, cap_warning_priorities are implementation specific; i.e. PAGASA might call it a "signal 1", while Myanmar calls it an "Orange" alert. 

The cap_alerting_authorities.csv is for training also, it contains all the global CAP OIDs. So when we do training for people from multiple countries they can use it like 2016 Aug Thailand training. 

Having said that, my assumption was all data that's in the SAMBRO template is relevant to all; e.g. event_type list.

modules/templates/SAMBRO/Demo/tasks.cfg
waidyanatha: ok, I think my comment is confusing. There are some XML file in the Demo folder that are quire old and needs reworking to make them useful for training.

waidyanatha:
I don't think its working out. When I click on "new pull request" it shows these same changes and a large green "view pull request" button that brings me to this same 1431 pull request; i.e. doesn't allow me to create a new one, even after following [option 2](http://eden.sahanafoundation.org/wiki/DeveloperGuidelines/Git#Squashingcommits) with HEAD@{2}.  I'll send the files to Biplov and have him do the pull request, he's knowledgeable, young, and has the patience to do this :-)

please reject all these. Thanks.

flavour:
As per docs, I use click, can't assist with GUI I'm afraid

modules/templates/SAMBRO/Demo/tasks.cfg
flavour: Demo data isn't relevant for Production deplyments, yes, this is normal.
In this case some of the Demo data isn't suitable for specific sub-templates since it is designed for different countries...includign terminilogy, as you point out...but the idea that Demo data isn't meant for Production use is normal.
Yes, data imported within the main SAMBRO template is designed for Production use...and across all Sub-Templates

biplovbhandari:
@waidyanatha  
nuwan you can do `git pull --rebase upstream master` to pull the new changes from the trunk and then to rebase multiple commits, do `git rebase -i HEAD~N` , where N is the number of the head behind to rebase to one commit and finally push changes back to the trunk using `git push origin master --force`

waidyanatha:
Thanks, those were the instructions Fran had shared; but the simple rebase
was not working for me. Had to do a rebase --hard to get it to work. It's
all good now.

PR #1433, opened 2018-01-18
Alphabetical order, additions and fixes to es.py

xuebozhu:
Alphabetical arrangement, added some new translations as 'Surveys': 'Encuestas', and some fixes.


flavour:
Doing the alphasort in the same commit with the other changes makes it really hard to review, can you separate these 2 parts? (maybe best if I alphasort)
Although I'm not a native Spanish speaker & hence my reviews can be limited in nature, I at least know which strings were translated  by the Red Cross & have specific meanings to them & so need to be more careful changing these

PR #1435, opened 2018-01-19
Updates to es.py

JBarrioL:
Reviewed a bit further, removed the bug (from_id), reverted the translations that wasnt righ.
Havent found yet the best translation for "getting" as fetch since fetch cant be translated to one single word, not at leas in the way its used on code .


JBarrioL:
Tried to rebase,and squas did horribly there,closing here then

PR #1437, opened 2018-01-21
SAMBRO AlertHub Sub-Template, 1st cut

waidyanatha:
adding new subtemplate AlertHub to SAMBRO; it is designed to server as a national alert aggregator.


modules/templates/SAMBRO/Demo/tasks.cfg
```diff
@@ -21,10 +21,19 @@ org,organisation,organisation.csv,organisation.xsl
...
+#
+cap,warning_priority,cap_warning_priority.csv,warning_priority.xsl
+#
+# @note (nuwan) - the three files with demo templates should only load in the demo and not other templates
 cap,alert,cap_alert.csv,alert.xsl
 cap,info,cap_info.csv,info.xsl
 cap,area,cap_area.csv,area.xsl
-cap,warning_priority,cap_warning_priority.csv,warning_priority.xsl
+#
+# @note (nuwan) - all the import_xml files should not load in other templates change to CSV to avoid this issue
```
flavour: I still think this note is wrong, but won't hold up the merge for it

modules/templates/SAMBRO/Demo/tasks.cfg
```diff
@@ -21,10 +21,19 @@ org,organisation,organisation.csv,organisation.xsl
...
 gis,config,gis_config.csv,config.xsl
 # CAP
+# -----------------------------------------------------------------------------
+# Alerting Authority
+# 
+cap,alerting_authority,cap_alerting_authority.csv,alerting_authority.xsl
+#
+cap,warning_priority,cap_warning_priority.csv,warning_priority.xsl
+#
+# @note (nuwan) - the three files with demo templates should only load in the demo and not other templates
```
flavour: I still think this note is meaningless...this *is* the Demo folder so should only load when Demos are desired...this isn't about these lines, but the whole file.
Again, I won't hold up the merge though

PR #1439, opened 2018-01-21
Update es.py

jvelasquezc:
Correccción de errores de la preposición en la frase.

### DIFF
--- a/languages/es.py
+++ b/languages/es.py
@@ -111,7 +111,7 @@
 "Please come back after sometime if that doesn't help.": 'Por favor vuelva más tarde si esto no le ayuda.',
 "Press the 'Delete Old' button to have all records which reference this one be repointed at the new one & then the old record will be deleted.": "Oprima el botón de 'Eliminar viejor' para que todos los registros con referencia a éste sean apuntados al nuevo y el registro viejo sea eliminado.",
 "Quantity in %s's Inventory": 'Cantidad en Inventario de %s',
-"Search here for a person's record in order to:": 'Buscar aquí por un registro de una persona a fin de:',
+"Search here for a person's record in order to:": 'Buscar aquí el registro de una de las personas a fin de:',
 "Select a Room from the list or click 'Create Room'": "Elegir una Sala de la lista or hacer click en 'Añadir Sala'",
 "Select a person in charge for status 'assigned'": "Seleccione una persona encargada para el estado 'asignado'",
 "Select this if all specific locations need a parent at the deepest level of the location hierarchy. For example, if 'district' is the smallest division in the hierarchy, then all specific locations would be required to have a district as a parent.": "Seleccione esta opción si todos los lugares específicos que se necesitan un padre en lo más profundo de la jerarquía de ubicación. Por ejemplo, si 'distrito' es la división más pequeño en la jerarquía, entonces todas las ubicaciones específicas deben tener un distrito como padre.",


languages/es.py
```diff
@@ -23,7 +23,7 @@
 'Membership Fee Due Date': 'Fecha de pago de la cuota de membresía',
 'Membership Types': 'Tipos de Membresía',
 'Paid': 'Pagada',
-'Partner ID': 'ID de Socio',
```
flavour: This was the term provided by the Red Cross, so they understand this

languages/es.py
```diff
@@ -165,24 +165,24 @@
...
-'%(PO)s Number': 'Número de Compra',
-'%(REQ)s Number': 'Número de Requisición',
-'%(app)s not installed. Ask the Server Administrator to install on Server.': '%(app)s no instalado. Pregunte al administrador del servidor para instalar en el servidor.',
-'%(count)s Roles of the user removed': '%(count)s Roles de usuarios Removidos',
-'%(count)s Users removed from Role': '%(count)s Usuarios removidos del papel',
-'%(count_of)d translations have been imported to the %(language)s language file': '%(count_of)d traducciones han sido importadas al %(language)s lenguaje de archivo',
-'%(label)s contains %(values)s': '%(label)s contienen %(values)s',
-'%(label)s contains any of %(values)s': '%(label)s contiene alguno de %(values)s',
-'%(label)s does not contain %(values)s': '%(label)s no contiene %(values)s',
+'%(GRN)s Number': 'Porcentaje de número de Nota de Bienes Recibidos',
```
flavour: No...any term with %(...)s is a variable string which must be left exactly as-is

languages/es.py
```diff
@@ -2253,7 +2253,7 @@
 'Driving License': 'Licencia Conducir',
 'Drought': 'Sequía',
 'Drugs': 'Drogas',
-'Due %(date)s': 'Debido %(date)s',
+'Due %(date)s': 'Porcentaje de Fechas de Vencimiento',
```
flavour: No: variable

languages/es.py
```diff
@@ -2581,7 +2580,7 @@
 'Emergency Water Supply': 'Emergencia de Abastecimiento de Agua',
 'Employment Type': 'Tipo de empleo',
 'Enable Crop': 'Habilitar cultivos',
-'Enable in Default Config?': 'Habilitar en Default Config?',
+'Enable in Default Config?': 'Habilitar en la configuracion predeterminada?',
```
flavour: I'm not sure that predeterminada is the right transaltion for Default here...it's not so much the starting configuration as the fallback one: it applies in all cases where it isn't over-ridden in more specific circumstances

languages/es.py
```diff
@@ -2978,7 +2977,7 @@
 'Goods Received Note': 'Nota de Bienes Recibidos',
 'Google Cloud Messaging': 'Mensajería en la nube de Google',
 'Google Layer': 'Capa Google',
-'Government UID': 'UID de Gobierno',
+'Government UID': 'Identificador de Usuario de Gobierno',
```
flavour: I think we should keep UID here as it's a technical term

languages/es.py
```diff
@@ -3128,7 +3127,7 @@
 'Hospital status assessment.': 'Hospital de evaluación de la situación.',
 'Hospital': 'Clínica',
 'Hospitals': 'Hospitales',
-'Host National Society': 'Sociedade Nacional Anfitriona',
```
flavour: This is the Red Cross term, so I guess should stay intact

languages/es.py
```diff
@@ -3138,12 +3137,12 @@
 'Hours added': 'Horas añadido',
 'Hours by Program Report': 'Otras horas, con Informe del Programa',
 'Hours by Role Report': 'Horas de Rol Informe',
-'Hours deleted': 'Horas borran',
-'Hours updated': 'Horas actualizan',
+'Hours deleted': 'Horas borradas',
+'Hours updated': 'Horas actualizadas',
 'Hours': 'Horas',
 'House Design': 'Casa Diseño',
-'Household kits received': 'Kits de hogares recibidos',
```
flavour: I think the term Kit, whilst an anglicisation, is one which is in common usage in the field & has meaning to the users

languages/es.py
```diff
@@ -4255,7 +4254,7 @@
 'News': 'Novedades',
 'Next View': 'Vista siguiente',
 'Next': 'Siguiente',
-'Niger': 'Níger',
+'Niger': 'Nígeria',
```
flavour: Niger != Nigeria
2 different countries

languages/es.py
```diff
@@ -6910,10 +6909,8 @@
 'Vendor': 'Vendedor',
 'Venue': 'Lugar de celebración',
 'Verification Status': 'Estado de verificación',
-'Verified': 'Verificado',
 'Verified?': '¿Verificado?',
 'Verify Password': 'Verifique su contraseña',
-'Verify password': 'Verifique su contraseña',
```
flavour: Why remove these two? I see them in the code, so they are still useful

PR #1440, opened 2018-01-21
some translation added and fixed

enriqueroman:


languages/es.py
```diff
@@ -4376,11 +4388,11 @@
...
 'No Records matching the query': 'Ningún registro coincide con la consulta',
 'No Reports currently registered': 'No hay informes registrados actualmente',
-'No Request Items currently registered': 'No hay articulos de Requisiciónes registrados actualmente',
+'No Request Items currently registered': 'No hay artículos de Requisiciónes registrados actualmente',
 'No Request Shipments': 'No hay Requisiciónes de embarque',
 'No Requests have been made yet': 'Aún no se ha hecho ningun requerimiento',
-'No Requests match this criteria': 'Ninguna Requisición coinciden con este criterio',
-'No Requests': 'Sin Requisiciónes',
+'No Requests match this criteria': 'Ninguna respuesta coinciden con este criterio',
+'No Requests': 'Sin respuestas',
```
flavour: Perhaps the term Requisicióne is more appropriate here as it's used for Logistics?
I think this is the term that the Honduran Red cross wanted

flavour:
Thankyou for adding missing accents & adding a couple of missing translations.
I'm concerned about mopdifying the Requisicione as I think this is the appropriate term for the current Spanish-speaking users of the system.
Please revise & send a single commit, thankyou

languages/es.py
enriqueroman: Yes, it seems to be more appropriate.

Sending a simple commit ;)

PR #1444, opened 2018-01-26
add column KV:cap_oid to preload CAP_OID as an Organization tag

waidyanatha:


waidyanatha:
bad pull need some modifications

PR #1455, opened 2018-02-11
SHARE template first cut

waidyanatha:
the template is built on initial understanding of the Sri Lanka UN-RCO shared requirements. This template serves to start the discussion on the developing the comprehensive blueprint: http://eden.sahanafoundation.org/wiki/BluePrint/ShaRe  

Some minor inconsistencies persist and will be corrected in the next pull request:

1. static/themes/SHARE/js/incident_profile.js needs to be renamed to a more appropriate file name like homepage.js. 
1. modules/templates/SHARE/Demo/tasks.cfg the "req" prepop is not working and the "doc" prepop is also throwing a warning; these data related bugs will be fixed

overall the template is working: http://www.sharehub.lk/ (you may use the default admin login for verifications)

### DIFF
--- a/modules/s3/s3query.py
+++ b/modules/s3/s3query.py
@@ -45,7 +45,7 @@ from gluon.storage import Storage
 
 from s3dal import Field, Row
 from s3fields import S3RepresentLazy
-from s3utils import s3_get_foreign_key, s3_unicode, S3TypeConverter
+from s3utils import s3_get_foreign_key, s3_str, s3_unicode, S3TypeConverter
 
 ogetattr = object.__getattribute__
 
@@ -147,7 +147,7 @@ class S3FieldSelector(object):
 
         try:
             rfield = S3ResourceField(resource, self.name)
-        except:
+        except (SyntaxError, AttributeError):
             colname = None
         else:
             colname = rfield.colname
@@ -207,7 +207,7 @@ class S3FieldSelector(object):
                     value = ogetattr(ogetattr(row, tname), fname)
                 else:
                     value = ogetattr(row, fname)
-            except:
+            except AttributeError:
                 try:
                     value = row[colname]
                 except (KeyError, AttributeError):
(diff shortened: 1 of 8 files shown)


waidyanatha:
@nursix hope this PR is to the required standard?

nursix:
The si.py language files has many empty string translations - they shouldn't be there. If a translation is not available, then it shouldn't appear in the language file at all.

Obviously, it would be better to get all these strings translated than to remove the empty ones, but removing empty strings should be the minimum before this can be merged.

languages/si.py
```diff
@@ -1,4 +1,1163 @@
...
+'About us': 'අප ගැන',
+'Academic': '',
+'Access Level': '',
+'Account Registered - Please Check Your Email': '',
+'Acronym': 'Acronym',
+"Acronym of the organization's name, eg. IFRC.": "Acronym of the organization's name, eg. IFRC.",
+'ACTION REQUIRED': '',
+'activate to sort column ascending': '',
+'activate to sort column descending': '',
+'Active': '',
```
nursix: Remove empty translations like this.

modules/templates/SHARE/menus.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # -------------------------------------------------------------------------
+    @classmethod
+    def menu_modules(cls):
+        """ Custom Modules Menu """
+
+        menu= [MM("Requests", c="req", f="req")(
+                MM("Commitments", c="req", f="commit"),
+                ),
+               MM("Aid & Supplies", c="supply", f="item/summary")(
```
nursix: This isn't fully correct.
```
  MM("Aid & Supplies", c="supply", f="item", m="summary")
```
is what it should be.

modules/templates/SHARE/menus.py
```diff
@@ -0,0 +1,39 @@
...
+                MM("Item Catalog", c="supply", f="catalog_item"),
+                MM("Assets", c="asset", f="asset"),
+                ),
+               MM("Sitreps", c="cms", f="blog/1"),
```
nursix: This isn't correct either. 

Not sure about whether there should really be a hardcoded record ID in the menu (it will fail if another record ID is assigned, e.g. when the record is replaced by another version), it may be better to specify a filter.

However, even if it were okay to have a record ID here, then it should be written like:
```
  MM("Sitreps", c="cms", f="blog", args=["1"])
```

waidyanatha:
OK, I'll get to it after the Chinese new year festivities.

Issue #1461, opened 2018-02-17
Calendar localization while using custom language


trendspotter:
The jQueryUI timepicker and datepicker modules are lacking localization for languages not officially supported by SE, even though the proper i18n JavaScript files exist.

The cause seems to be in *s3widgets.py* on lines [2030](https://github.com/sahana/eden/blob/7e6b87b85b92272409dac12f91c217f242cc5fde/modules/s3/s3widgets.py#L2030) and [2398](https://github.com/sahana/eden/blob/7e6b87b85b92272409dac12f91c217f242cc5fde/modules/s3/s3widgets.py#L2398) as they are trying to look up a value from [date_formats](https://github.com/sahana/eden/blob/002f6708dac733568876c1857e01ba8ac2422764/modules/s3cfg.py#L62) dictionary and if the current language is not present there, the internationalization is simply ignored even though `settings.L10n.date_format` may be set properly.

I see two ways out
1. Add all remaining languages for which there is an i18n JavaScript file to `date_formats` dictionary.
2. Make `date_formats` dictionary configurable in similar fashion like the [`languages`](https://github.com/sahana/eden/blob/002f6708dac733568876c1857e01ba8ac2422764/modules/s3cfg.py#L1000) are.

I'd be happy to provide a PR for the first one, however the second one might be more systematic (and tricker), so that's why I'm opening it as an issue first.

nursix:
Ahm...

current.deployment_settings.date_formats *is* customizable - you can override the class property with an instance property in your template's config.py:
```
settings.date_formats = {"cz": "%d.%m.%Y"}
```
...or, if you just want to extend it:
```
date_formats = dict(settings.date_formats.items())
date_formats["cz"] = "%d.%m.%Y"
settings.date_formats = date_formats
```
But of course you can also extend the default set with more languages if you like - although as I see it, you only need one particular language, right?

nursix:
Btw - the language file for Czech is also still missing. Why not put that in a PR too?

trendspotter:
> override the class property with an instance property

Well yes, but it generally didn't seem as a good idea. Or maybe I just don't understand why this particular dictionary is separated from the others (eg. languages) and put up high in the config files where it feels like it's something too important to be touched.

> you only need one particular language

True, but when I'm already in it, I can also be nice to other geographical minorities.

> the language file for Czech is also still missing. Why not put that in a PR too?

[...]

PR #1462, opened 2018-02-17
Add date formats for more languages

trendspotter:
Adds more date formats (aligned to JavaScript specifications) as discussed in #1461.


nursix:
Whilst it is nice to have them all translated, I see three things:
- some of them will never be used because they don't match a selectable language (e.g. "fo")
- they can be grouped in order to drastically reduce the number of dict entries
- this can be made a lazy property so that requests which don't need them can bypass it

Good for now, but definitely not optimal.

nursix:
Or actually - a lazy property performs worse than a class property, so strike that point. Good as-is, but worth moving into S3Calendar (closer to context, and not as suggestive re: configurability).

trendspotter:
Yeah, but *S3Calendar* loads the format via `current.deployment_settings.get_L10n_date_format()` which points again back to *S3Config*.

About the grouping - How? Something like this?
```
d_slash_m_slash_Y = "%d/%m/%Y"
d_dot_m_dot_Y = "%d.%m.%Y"
date_formats = {"af": d_slash_m_slash_Y,
                    "ar": d_slash_m_slash_Y,
                    "ar-dz": d_slash_m_slash_Y,
                    "az": d_dot_m_dot_Y,
                    "be": d_dot_m_dot_Y,
...
```

PR #1466, opened 2018-02-19
Edu, Inv: Fix display of obsolete flag when NONE is not a single character

trendspotter:
When `NONE` is an empty string, these two places throw error 500 when the page is rendered. When `NONE` is longer than a single character, these two would display just `NONE[0]`, ie. the first character of the string.

Checked also the other occurrences of `(opt and ...` and in the rest the `NONE` is already wrapped to list.

### DIFF
--- a/modules/s3db/edu.py
+++ b/modules/s3db/edu.py
@@ -198,7 +198,7 @@ class S3SchoolModel(S3Model):
                            default = False,
                            label = T("Obsolete"),
                            represent = lambda opt: \
-                                       (opt and [T("Obsolete")] or NONE)[0],
+                                       (opt and [T("Obsolete")] or [NONE])[0],
                            readable = False,
                            writable = False,
                            ),
--- a/modules/s3db/inv.py
+++ b/modules/s3db/inv.py
@@ -266,7 +266,7 @@ class S3WarehouseModel(S3Model):
                            default = False,
                            label = T("Obsolete"),
                            represent = lambda opt: \
-                                       (opt and [T("Obsolete")] or NONE)[0],
+                                       (opt and [T("Obsolete")] or [NONE])[0],
                            readable = False,
                            writable = False,
                            ),


modules/s3db/edu.py
```diff
@@ -198,7 +198,7 @@ def model(self):
                            default = False,
                            label = T("Obsolete"),
                            represent = lambda opt: \
-                                       (opt and [T("Obsolete")] or NONE)[0],
+                                       (opt and [T("Obsolete")] or [NONE])[0],
```
nursix: This is an unnecessarily complex lambda (and slow due to list creation), please change to:
```
represent = lambda opt: T("Obsolete") if opt else NONE
```

I don't know why it has been done the way it was done, but it is not useful however.

modules/s3db/inv.py
```diff
@@ -266,7 +266,7 @@ def model(self):
                            default = False,
                            label = T("Obsolete"),
                            represent = lambda opt: \
-                                       (opt and [T("Obsolete")] or NONE)[0],
+                                       (opt and [T("Obsolete")] or [NONE])[0],
```
nursix: Same here:
```
represent = lambda opt: T("Obsolete") if opt else NONE
```

nursix: NB to make it performing even better, you can store T("Obsolete") in a local closure variable rather than creating it in the lambda, i.e. further up in the model class:
```
OBSOLETE = T("Obsolete")
```
and then in the lambda:
```
represent = lambda opt: OBSOLETE if opt else NONE
```

nursix:
Okay, I did the whole lot here:
https://github.com/nursix/eden/commit/b5676c71a0aed955d3dc61810b8105f9b424ce58

...you want to try it?

trendspotter:
Yup, nursix/eden@b5676c7 works nicely.

Just out of curiosity, why `NONE = messages["NONE"]` but `OBSOLETE = messages.OBSOLETE`? Why not `NONE = messages.NONE`. The reserved word should be `None` with exact casing, so why does this one need to be accessed as key and not as property?

nursix:
In a Messages object, property access automatically translates as T(), whereas key access does not.

Since NONE is only a symbol ("-"), we don't want it to be translated (it would be "-" in every language), so accessing it as ```current.messages["NONE"]``` saves us a T-instantiation and dictionary look-up upon rendering.

PR #1470, opened 2018-02-21
Bulletin Form for OCD PH

biplovbhandari:


modules/s3cfg.py
```diff
@@ -2683,6 +2683,27 @@ def get_cap_area_default(self):
...
+        """
+
+        return self.cap.get("bulletin_method")
+
+    def get_cap_bulletin_officer(self):
+        """
+            Name of the officer who is responsible to issue the bulletin
+        """
+
+        return self.cap.get("bulletin_officer", "Name of Officer")
```
flavour: This stuff seems too dynamic to want to put into settings...although that is about to become more accessible via WebSetup.
Until then it would be better as a DB table? (Even with WebSetup, a DB table provides better UX)

modules/s3cfg.py
```diff
@@ -2683,6 +2683,27 @@ def get_cap_area_default(self):
 
         return self.cap.get("area_default", ["geocode", "polygon"])
 
+    def get_cap_bulletin_method(self):
+        """
+            Name of the method that is used to generate the bulletin for this deployment
+        """
+
+        return self.cap.get("bulletin_method")
```
flavour: Seems odd to have this be configurable via settings like this?
What I would propose would be an S3Method for cap/alert/[id]/bulletin
This would still allow the template to define a custom function in config.py

modules/s3db/cap.py
```diff
@@ -3424,26 +3426,50 @@ def cap_rheader(r):
...
                             # Show these buttons only if there is atleast one area segment
                             area_table = s3db.cap_area
                             area_row = db(area_table.alert_id == alert_id).\
                                                 select(area_table.id,
                                                        limitby=(0, 1)).first()
-                            if area_row and has_permission("update", "cap_alert",
+                            if area_row:
+                                # Show these buttons only if there is at least one area segment
+                                cap_bulletin_method = settings.get_cap_bulletin_method()
+                                if cap_bulletin_method:
```
flavour: If using the S3Method & DB configuration, the cap_bulletin table could be queried & if there is an undeleted row then proceed to add the buttons

flavour:
Sorry for the delay in reviewing this - I just found a lot of GitHub mails in my spam folder!
Other than the suggestions above, the logo should really be in the static/themes/XX/img folder rather than main static/img

biplovbhandari:
Thanks for review @flavour . I will get back soon

PR #1474, opened 2018-02-24
Limit allowed range for some integer values

trendspotter:
It has been brought to my attention that message 
> Enter an integer between -2.14748e+09 and 2.14748e+09

is scary and confusing for squeamish users.

This PR sets allowed ranges for some integer values which are not yet limited. The limits were mostly taken from other places in the code where similar values are set.


nursix:
This is overkill.
Whilst in many cases it makes sense to set a lower limit of 0, it doesn't make sense to set an upper limit of 99999999 or so - this should be removed. The IS_INT_IN_RANGE validator adjusts the error message to only report a lower limit of you use it like IS_INT_IN_RANGE(0, None).

modules/s3db/budget.py
```diff
@@ -332,12 +332,17 @@ def model(self):
                            ),
                      Field("salary", "integer", notnull=True,
                            label = T("Monthly Salary"),
-                           requires = IS_NOT_EMPTY(),
+                           requires = [IS_NOT_EMPTY(),
+                                        IS_INT_IN_RANGE(0, 999999999999),
```
nursix: This is crap (sorry, but it is). Should be ```IS_INT_IN_RANGE(0, None)``` instead.

modules/s3db/budget.py
```diff
@@ -332,12 +332,17 @@ def model(self):
...
+                                        ],
                            ),
                      s3_currency(),
                      Field("travel", "integer",
                            default = 0,
                            label = T("Travel Cost"),
+                           requires = IS_EMPTY_OR(
+                                        IS_INT_IN_RANGE(0, 999999999999),
```
nursix: Same - this upper limit is just nonsensical.

modules/s3db/edu.py
```diff
@@ -177,6 +177,9 @@ def model(self):
                      Field("capacity", "integer",
                            label = T("Capacity"),
                            represent = lambda v: v or NONE,
+                           requires = IS_EMPTY_OR(
+                                          IS_INT_IN_RANGE(0, 99999999)
```
nursix: Neither is this a useful upper limit.

modules/s3db/hrm.py
```diff
@@ -2429,6 +2429,9 @@ def model(self):
                            ),
                      Field("hours", "integer",
                            label = T("Hours"),
+                           requires = IS_EMPTY_OR(
+                                        IS_INT_IN_RANGE(0, 1000)
```
nursix: This is arbitrary - why 1000? There should be no upper limit really.

modules/s3db/transport.py
```diff
@@ -182,9 +182,15 @@ def model(self):
                      # http://en.wikipedia.org/wiki/Runway#Declared_distances
                      Field("runway_length", "integer",
                            label = T("Runway Length (m)"),
+                           requires = IS_EMPTY_OR(
+                                          IS_INT_IN_RANGE(0, 99999)
```
nursix: Nope - no upper limit here (lower limit is ok)

modules/s3db/transport.py
```diff
@@ -182,9 +182,15 @@ def model(self):
...
+                                          ),
                            ),
                      Field("runway_width", "integer",
                            label = T("Runway Width (m)"),
+                           requires = IS_EMPTY_OR(
+                                          IS_INT_IN_RANGE(0, 999)
```
nursix: Same here - no upper limit pls.

modules/s3db/vehicle.py
```diff
@@ -178,11 +178,17 @@ def model(self):
                      Field("mileage", "integer",
                            label = T("Current Mileage"),
                            represent = lambda v: int_represent(v),
+                           requires = IS_EMPTY_OR(
+                                          IS_INT_IN_RANGE(0, 99999999)
```
nursix: Completely nonsensical upper limit - should be None.

trendspotter:
The upper limit also helps to not go out of bounds for the integers. Ie when you set staff salary (`/eden/budget/staff`) to 758530000000000 (which is 1000 USD in Zimbabwean dollars in Jul 2008 :) ), you'll end up with

```
  File "/srv/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "/srv/web2py/applications/eden/controllers/budget.py", line 554, in <module>
  File "/srv/web2py/gluon/globals.py", line 419, in <lambda>
    self._caller = lambda f: f()
  File "/srv/web2py/applications/eden/controllers/budget.py", line 107, in staff
    return s3_rest_controller()
  File "/srv/web2py/applications/eden/models/00_utils.py", line 243, in s3_rest_controller
    output = r(**attr)
  File "applications/eden/modules/s3/s3rest.py", line 687, in __call__
  File "applications/eden/modules/s3/s3rest.py", line 1776, in __call__
  File "applications/eden/modules/s3/s3crud.py", line 111, in apply_method
    output = self.select(r, **attr)
  File "applications/eden/modules/s3/s3crud.py", line 1241, in select
    form = self.create(r, **attr).get("form", None)
  File "applications/eden/modules/s3/s3crud.py", line 393, in create
    format=representation,
  File "applications/eden/modules/s3/s3forms.py", line 473, in __call__
    format = format,
  File "applications/eden/modules/s3/s3forms.py", line 672, in process
    hideerror=False):
  File "/srv/web2py/gluon/sqlhtml.py", line 1965, in accepts
    self.vars.id = self.table.insert(**fields)
  File "/srv/web2py/gluon/packages/dal/pydal/objects.py", line 753, in insert
    ret = self._db._adapter.insert(self, row.op_values())
  File "/srv/web2py/gluon/packages/dal/pydal/adapters/base.py", line 486, in insert
    raise e
DataError: integer out of range
```

modules/s3db/hrm.py
trendspotter: 1000 is taken from https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3db/hrm.py#L2648-L2653

nursix:
I still disagree - if there is a risk that we end up with values beyond the INT range then the field type must be changed to BIGINT. It makes no sense to set an arbitrary upper limit if there /could/ be values beyond it (or how would the users be supposed to enter the values then?? because enter them they must).

trendspotter:
So if you don't like 9999-ish upper limits, is it worth to remove them everywhere?

```
/srv/web2py/applications/eden # grep -r 'IS_INT_IN_RANGE.*9999' . | wc -l
69
/srv/web2py/applications/eden # grep -r 'IS_INT_IN_RANGE.*99999' . | wc -l
55
/srv/web2py/applications/eden # grep -r 'IS_INT_IN_RANGE.*999999' . | wc -l
53
/srv/web2py/applications/eden # grep -r 'IS_INT_IN_RANGE.*9999999' . | wc -l
20
/srv/web2py/applications/eden # grep -r 'IS_INT_IN_RANGE.*99999999' . | wc -l
20
/srv/web2py/applications/eden # grep -r 'IS_INT_IN_RANGE.*999999999' . | wc -l
1
```

modules/s3db/hrm.py
```diff
@@ -3160,6 +3166,9 @@ def model(self):
                                      ),
                      Field("expiry", "integer",
                            label = T("Expiry (months)"),
+                           requires = IS_EMPTY_OR(
+                                        IS_INT_IN_RANGE(0, 1000)
```
nursix: Same here - why 1000 out of all possible values? Surely 1000 months is way beyond a meaningful expiry interval, and just as good as no expiry interval at all. But what else is meaningful?

I don't think we have reason to set an upper limit (but we should have a lower limit - and it shouldn't be zero)

modules/s3db/hrm.py
```diff
@@ -2939,6 +2942,9 @@ def model(self):
                                  ),
                      Field("hours", "integer",
                            label = T("Hours"),
+                           requires = IS_EMPTY_OR(
+                                        IS_INT_IN_RANGE(0, 1000)
```
nursix: Yeah - a training that lasts 1000 hours for an individual is certainly a very...ahm...extensive training, but it's not so unusual. Some trainings can last even longer than that (I had a 1800 hours training in intensive care, for example).

Not sure we have reason to set an upper limit here - but we should have a lower limit for sure.

trendspotter:
> expiry should be more than zero if it is set at all, as it makes no sense otherwise

That goes also for some other values, doesn't it? 0 m<sup>3</sup> warehouse capacity and 0 m runway length don't make sense either, if there's a possibility not to fill them at all.

nursix:
Yeah - I don't like arbitrary validation rules that have no practical meaning.

In most of the cases you are looking at here, there is indeed a need to limit these fields to values greater than 0 - as otherwise the value would definitely be wrong. So, that is what I call a use-case defined limit, and it should be enforced.

But a 99999 upper limit usually has no foundation in the use-case.

The point is that it's not about validating the input against reality - of course it is unlikely that there will be a vehicle in active service with a mileage beyond a billion kilometers. But if the user would enter such a value, then we don't know /why/ he does that, and no real reason in the use-case to deny it.

We have seen that users sometimes deliberately fill in unrealistic values to indicate that while the exact value is not known, it is beyond a certain range - e.g. the exact mileage of a vehicle is not known, but it's definitely beyond service range, so the users have agreed to set 999999999 as mileage for these vehicles.

Similar, we have seen that people fill in XXX as names for people they do not yet know the name of (e.g. an unconscious person) - an definitely XXX is not a valid name, but that's exactly the point /why/ they fill that in.

So, validation rules should be very measured, and exclude only values which actually disrupt the use-case - and not necessarily all unusual values. They can still be adapted in templates if the particular deployment wants it, but by default they should be a strict as necessary and as lenient as possible.

trendspotter:
The explanation is clear and thanks for that. But than again, you *already have* such upper limits in your codebase, so I do take a little offense in being called out doing the same thing. Therefore I ask again - is it worth to remove them everywhere (or at least most of the places)? I have no problem doing it as part of this PR, but I'd like to know that the invested time wouldn't be wasted.

Eg.

https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3db/dvi.py#L106

https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3db/gis.py#L2671

https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3db/hms.py#L542

https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3db/org.py#L2085

https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3db/project.py#L2192

etc.

nursix:
Oh, this isn't /my/ code base ;) and in some places it has been another me eight years ago or so. 

There is definitely a lot of code that could/should be improved upon or even corrected - and there is (as per license) no claim that everything in this code base is okay, acceptable or even functional.

Now, like you, I would have to look at the individual use-cases to judge whether or not the field validation rules are trying to enforce limits beyond what the use-case actually tolerates. And some may do so.

I can not possibly give you a general "green light" to change these rules - but if you change them, I can evaluate your changes. And I will do so following the rule I described above: generic use-cases should only ever exclude values that disrupt the functionality or the semantics of the intended case, but not necessarily unusual or seemingly impossible values.

That's all I can say at this point - but do make changes and I will take a closer look ;) I promise.

trendspotter:
Rationale for changes in the second commit
 - 9999like values converted to `None`.
 - Year ranges aligned to (1850, 2100), including `pr` which had limit on `current.request.now.year` (My CMOS battery may be dead and suddenly I time travel to 1970).
 - Height and weight in `pr` which previously has human-like limits set to `None` due to the remark about deliberately entering invalid values.

Review away.

modules/s3db/pr.py
```diff
@@ -5539,7 +5539,7 @@ def nationality_opts():
                           Field("year_of_birth", "integer",
                                 label = T("Year of Birth"),
                                 requires = IS_EMPTY_OR(
-                                    IS_INT_IN_RANGE(1900, current.request.now.year)
```
nursix: There was a point to this, though - because it may be used to calculate age, in which case a future year of birth could disrupt statistics. CMOS battery is not really a relevant edge-case here ;)

nursix: Lowering the lower limit to 1850 seems to want to allow dead people, for which there isn't really a use-case in Eden. There are no more people on the planet born before 1900, so we can as well leave 1900 as lower limit.

modules/s3db/project.py
```diff
@@ -2017,7 +2017,7 @@ def model(self):
                           Field("year", "integer", notnull=True,
                                 default = None, # make it current year
                                 label = T("Year"),
-                                requires = IS_INT_IN_RANGE(1950, 3000),
+                                requires = IS_INT_IN_RANGE(1850, 2100),
```
nursix: There is real need to change this, or to move it further back in time anyway.

modules/s3db/supply.py
```diff
@@ -409,6 +409,9 @@ def model(self):
                      Field("year", "integer",
                            label = T("Year of Manufacture"),
                            represent = lambda v: v or NONE,
+                           requires = IS_EMPTY_OR(
+                                        IS_INT_IN_RANGE(1850, 2100)
```
nursix: I think this can well be moved 50 years more into the future, and the upper limit should be the current year (I mean, if we now introduce an upper limit) so that age calculations don't produce negative values.

modules/s3db/hrm.py
```diff
@@ -2429,6 +2429,9 @@ def model(self):
                            ),
                      Field("hours", "integer",
                            label = T("Hours"),
+                           requires = IS_EMPTY_OR(
+                                        IS_INT_IN_RANGE(1, None)
```
nursix: Here we should allow 0.

modules/s3db/hrm.py
```diff
@@ -2939,6 +2942,9 @@ def model(self):
                                  ),
                      Field("hours", "integer",
                            label = T("Hours"),
+                           requires = IS_EMPTY_OR(
+                                        IS_INT_IN_RANGE(1, None)
```
nursix: This should allow 0 too.

modules/s3db/hrm.py
```diff
@@ -2648,7 +2651,7 @@ def model(self):
                      Field("hours", "integer",
                            label = T("Hours"),
                            requires = IS_EMPTY_OR(
-                                        IS_INT_IN_RANGE(1, 1000),
+                                        IS_INT_IN_RANGE(1, None),
```
nursix: I'm not sure about the 1 here either - but it was there before, so we can leave it for now (I think 0 should be allowed as deliberate "implausible" value).

modules/s3db/pr.py
```diff
@@ -4844,7 +4844,7 @@ def model(self):
                            ),
                      Field("height_cm", "integer",
                            label = T("Height (cm)"),
-                           requires = IS_EMPTY_OR(IS_INT_IN_RANGE(0, 300)),
+                           requires = IS_EMPTY_OR(IS_INT_IN_RANGE(0, None)),
```
nursix: We should leave the upper limit here because values outside of this range are truly invalid for the use-case (which is about identification)

modules/s3db/pr.py
```diff
@@ -5145,7 +5145,7 @@ def model(self):
                            label = T("Year"),
                            represent = lambda v: v or NONE,
                            requires = IS_EMPTY_OR(
-                                        IS_INT_IN_RANGE(1900, 2100)
+                                        IS_INT_IN_RANGE(1850, 2100)
```
nursix: There is no real point in moving these further back in time - there are no more people on the planet born before 1900, so why should anyone have received awards before 1900?

I'm not sure about the upper limit - one might allow to enter awards that are planned, but certainly not decades into the future. But if we can't easily find a meaningful limitation, then we shouldn't really it ;) so agreed to leave it.

nursix:
I'm sorry to keep you busy with this, I know you want to move on.

But if we are to touch all these *at all*, then we should consider them on a case-by-case basis rather than applying any one-fits-all rules - and this takes just as much time on my side as on yours. I'm just being careful here since we've stumbled over overzealous validation rules before - and then it's often hard to answer the "why?".

However, I think we're almost there ;)

trendspotter:
Fixed and squashed

flavour:
Sorry for the delay in merging this...I just found a load of GitHub mails in my Spam folder!

Issue #1475, opened 2018-02-24
Number formatting does not cope with big numbers which are forced to use scientific notation


trendspotter:
Steps to reproduce:
1. Create a project, fill in the mandatory *Lead Implementer* and *Project Name*, and input 12 or more digits to *Budget*, ie. `123456798012`. (The value is not that outlandish, given the exchange rate of some currencies and size of some projects.)
2. Click *Save*.
3. Observe the value in *Budget*.
4. Edit (*Open*) the project.
5. Observe the value in *Budget* again.

Expected result:
The whole number will always be visible.

Actual result:
The number is converted to scientific notation (`1.23456789012e+11`) and then formatted as decimal with two digits precision, resulting in `1.23`.

nursix:
That looks like an easy one to fix. You want to try this yourself?

https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3/s3validators.py#L385

nursix:
You could e.g. change this:
https://github.com/sahana/eden/blob/9ce58b3aa1083b9b3f87c8a1386f63f96a5277b3/modules/s3/s3validators.py#L477

into:
```
str_number = format(number, '.%df' % precision)
```

...because:
```
>>> value = 123456789012
>>> number = float(value)
>>> str_number = unicode(number)
>>> str_number
u'1.23456789012e+11'
```
but:
```
>>> value = 123456789012
>>> number = float(value)
>>> str_number = format(number, '.02f')
>>> str_number
'123456789012.00'
```

nursix:
...and for integers analogously (although I haven't seen that use scientific notation)

nursix:
P.S. I would like to see a unit test for these, though ;)

nursix:
Okay, my apologies - I see that this was asking too much.

But for your awareness, here is what I meant:
https://github.com/nursix/eden/commit/fe938bccf084a25ef836244238ef9c40f6cf3ea1#diff-5e96f5c2956354ad943a6f2e3093d31aR1385

And here is why:

We're by no means striving after a mythical "100% test coverage" (which usefulness is debatable), but for certain critical core requirements it is useful to have them automatically verified for every trunk commit.

But besides critical core requirements, there is also those less obvious edge cases (like this issue), which may not be thought of when refactoring code - and that is where such test cases are useful documentation.

I will give you another example...in an hour.

trendspotter:
[...]

PR #1476, opened 2018-02-27
Fix #1465, supply model validation

### LINKED ISSUE #1465: Supply Item creation ignores decimal separator settings
Steps to reproduce:
Create an item in *Supply* module, fill in the mandatory fields and enter *Length* as `0,1`.

Expected behavior:
Value is accepted in accordance with `settings.L10n.decimal_separator`.

Actual behavior:
Validation of the value fails unless it is entered using dot as decimal separator.

I understand that this may be tricky to fix if such thing is not directly supported by Web2py.

daveycrockett:
Given #1475, I believe this is all that remains for fixing #1465. Let me know if there's anything else I should add/change/remove from this PR!

### DIFF
--- a/modules/s3db/supply.py
+++ b/modules/s3db/supply.py
@@ -414,26 +414,31 @@ $.filterOptionsS3({
                            label = T("Weight (kg)"),
                            represent = lambda v: \
                                        float_represent(v, precision=2),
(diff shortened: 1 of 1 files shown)


flavour:
Thanks, apologies it took so long to merge...I just found a load of GitHub mails in my spam folder!

PR #1478, opened 2018-03-05
Some contribution in languages

Myeongjung:
We changed some mistakes from arabic and korean.
those are not completely finished yet.
we are a native speaker of that languages.


languages/ar.py
```diff
@@ -107,19 +107,19 @@
 'Add Skill': 'أضافة مهارة جديدة',
 'Add Skill Equivalence': 'أضف مهارة معادلة',
 'Add Solution': 'إضافة الحل',
-'Add Staff': 'أضف موظفين',
+'Add Staff': 'أضف فريق عمل',
```
nursix: An Arabic-speaking friend says you're changing the meaning here - which is not correct.
"Add Staff" has the same meaning as "Add Staff Members", so it should be أضف موظفين

nursix:
Where the English phrase is ambiguous (and sometimes it can be), you should make sure that you check in-context to find the correct meaning - rather than aiming for exact textual translation. The Arabic translation team has done exactly that.

In general, context-sensitive+meaningful is preferable over textual translation.

In the example of "Add Staff" you changed the meaning of the translation to something that means "Add a team", but that's incorrect in the context because the phrase is used on action items to add individual staff members. There is also an action phrase "Add a team" in a different context, which uses exactly the translation you applied for "Add Staff", but does something completely differently (namely, adding a "team").

ExtraMHD:
Thanks for the info :) fixed

PR #1480, opened 2018-03-25
some translation fixed

Myeongjung:
We got the feedback from you guys about Arabic part, and we changed some mistakes there.
Additionally, I just fixed more in korean language file, it's not completely finished, there are still have a lot of things need to change.

I changed them because there were some grammatical mistakes, people can guess what it means however that was hard to read for korean. Not only that some parts of translation were totally wrong, for example 'Camp Service Details' was translated like 'Children Service Details' which is completely wrong, 'Camp Service added' was translated 'Add Camp Service'. Therefore, that makes even more confusing to the user, thus it's critical problem to use Eden system and bad for internationalization also.

I also have a question is there has any reason for make it duplicated between 'To search for a hospital, enter any of the names or IDs of the hospital, or the organization name or acronym,...' with  'To search for a hospital, enter any of the names or IDs of the hospital, separated by spaces...' those seem almost same the only one different thing is organization name or acronym. Therefore, I assume that probably can make one sentence?

Thanks


nursix:
> I also have a question is there has any reason for make it duplicated between 'To search for a hospital, enter any of the names or IDs of the hospital, or the organization name or acronym,...' with 'To search for a hospital, enter any of the names or IDs of the hospital, separated by spaces...' those seem almost same the only one different thing is organization name or acronym. Therefore, I assume that probably can make one sentence?

That is a rather surprising assumption as the difference between these two instructions is quite significant. Two points:

a) You MUST NOT change the original messages (i.e. the English strings on the left). If you change them, the respective translation can not be found by the system. Obviously, that means you can not remove any of the original messages either, nor combine two different strings into one, for whatever reason.

b) Different user instructions for filters typically refer to different filter configurations. Here, in one case you can search hospitals by IDs and names - and in the other case by IDs and names *and* organization name and acronym, which is a significant difference.

So, the answer is "No".

Myeongjung:
Yes, I know that's really important to keep the original messages on the left side which is in English.
It might be connected with other strings which can be translated.
what I assumed that the first filter configuration is also included in second one so, I asked you the possibility of make them simple. I understand what you mean, it's completely different, so it's important also.
Thanks anyway :), I didn't change the original messages only changed the translation part
that might be fine.

flavour:
Thankyou, I will need to trust you that these improvements make better sense in Arabic/Korean as I don't speak these languages, but I can see no syntax errors, so will merge :)

PR #1482, opened 2018-03-29
Allow custom decimal separator in input and representation

trendspotter:
Related to #1465.

### DIFF
--- a/modules/s3db/dvr.py
+++ b/modules/s3db/dvr.py
@@ -103,6 +103,8 @@ class DVRCaseModel(S3Model):
         define_table = self.define_table
         person_id = self.pr_person_id
 
+        float_represent = IS_FLOAT_AMOUNT.represent
+
         beneficiary = settings.get_dvr_label() # If we add more options in future then == "Beneficiary"
         manage_transferability = settings.get_dvr_manage_transferability()
 
@@ -2896,7 +2898,11 @@ class DVRCaseEffortModel(S3Model):
                          comment = None,
                          ),
                      Field("hours", "double",
-                           requires = IS_FLOAT_IN_RANGE(0.0, None),
+                           represent = lambda v: \
+                                       float_represent(v, precision=2),
+                           requires = IS_FLOAT_AMOUNT(minimum=0.0),
+                           widget = S3HoursWidget(precision = 2,
+                                                  ),
                            ),
                      s3_comments(),
                      *s3_meta_fields())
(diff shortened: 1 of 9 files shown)


modules/s3db/dvr.py
```diff
@@ -2896,7 +2898,9 @@ def model(self):
                          comment = None,
                          ),
                      Field("hours", "double",
-                           requires = IS_FLOAT_IN_RANGE(0.0, None),
+                           represent = lambda v: \
+                                       float_represent(v, precision=2),
+                           requires = IS_FLOAT_AMOUNT(0, None),
```
nursix: This would better use the hours widget.

modules/s3db/dvr.py
```diff
@@ -4058,10 +4068,10 @@ def model(self):
                                  3: T("refused"),
                                  4: T("missed"),
                                  }
-        amount_represent = lambda v: IS_FLOAT_AMOUNT.represent(v,
-                                                               precision = 2,
-                                                               fixed = True,
-                                                               )
+        amount_represent = lambda v: float_represent(v,
+                                                    precision = 2,
```
nursix: Indent one more.

modules/s3db/hrm.py
```diff
@@ -1612,8 +1614,10 @@ def model(self):
                              set_max = "#hrm_salary_start_date",
                              ),
                      Field("monthly_amount", "double",
+                           represent = lambda v: \
+                                        float_represent(v, precision=2),
                            requires = IS_EMPTY_OR(
-                                        IS_FLOAT_IN_RANGE(minimum=0.0)
+                                        IS_FLOAT_AMOUNT(0, None)
```
nursix: Using explicit ```minimum=0.0``` is actually better than ```0, None```

modules/s3db/inv.py
```diff
@@ -4588,10 +4590,11 @@ def qnty_adj_repr(value):
             Make unadjusted quantities show up in bold
         """
 
+        repr = IS_FLOAT_AMOUNT.represent(value, precision=2)
```
nursix: Variable name overwrites a built-in function, should not use ```repr``` as variable name.

modules/s3db/project.py
```diff
@@ -10649,7 +10649,11 @@ def model(self):
                      Field("time_estimated", "double",
                            label = "%s (%s)" % (T("Time Estimate"),
                                                 T("hours")),
-                           represent = lambda v: v or "",
+                           represent = lambda v: \
+                                IS_FLOAT_AMOUNT.represent(v, precision=2),
```
nursix: Not sure about the precision argument here - typically, estimates are full hours or should be rounded to full hours.

modules/s3db/project.py
```diff
@@ -11213,8 +11217,8 @@ def model(self):
                      Field("hours", "double",
                            label = T("Effort (Hours)"),
                            requires = IS_EMPTY_OR(
-                                       IS_FLOAT_IN_RANGE(0.0, None)),
-                           represent = lambda hours: "%.2f" % hours if hours else NONE,
+                                       IS_FLOAT_AMOUNT(0.0, None)),
```
nursix: This makes little sense in combination with the S3HoursWidget

modules/s3db/req.py
```diff
@@ -1581,7 +1581,7 @@ def model(self):
                            label = T("Quantity"),
                            represent=lambda v: \
                             IS_FLOAT_AMOUNT.represent(v, precision=2),
-                           requires = IS_FLOAT_IN_RANGE(minimum=1),
+                           requires = IS_FLOAT_AMOUNT(1, None),
```
nursix: Again - explicit ```minimum=1``` is better than ```0, None```

modules/s3db/project.py
nursix: Okay - I found the examples where this precision makes sense, namely the very small estimates (<1h).

So, can be left as 2.

modules/s3db/project.py
nursix: Okay, I fixed IS_FLOAT_AMOUNT to not attempt to convert the output of S3HoursWidget, but just enforce the limits, so this combination should be okay now.

PR #1483, opened 2018-04-02
fixed bug#1243 Record Tab Merger

huynhv:
I added a tabindex attribute to the widgets being created.
Tab should now go through every attribute of one record, and then the same loop for the other record.

### DIFF
--- a/modules/s3/s3merge.py
+++ b/modules/s3/s3merge.py
@@ -442,8 +442,8 @@ class S3Merge(S3Method):
 
         trs = []
         init_requires = self.init_requires
-        count = 1
-        fieldsNum = len(formfields)
+        index = 1
+        num_fields = len(formfields)
         
         for f in formfields:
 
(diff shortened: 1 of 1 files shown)


modules/s3/s3merge.py
```diff
@@ -442,6 +442,9 @@ def merge(self, r, **attr):
 
         trs = []
         init_requires = self.init_requires
+        count = 1
+        fieldsNum = len(formfields)
```
nursix: num_fields please, not fieldsNum (camelCase is for class names only, variables use all-lowercase with underscore as word separator).

modules/s3/s3merge.py
```diff
@@ -486,7 +489,8 @@ def merge(self, r, **attr):
             trs.append(TR(TD(owidget, _class="mwidget"),
                           TD(swap),
                           TD(dwidget, _class="mwidget")))
-
+            
+            count = count+1
```
nursix: Blanks around the + please

modules/s3/s3merge.py
```diff
@@ -442,6 +442,9 @@ def merge(self, r, **attr):
 
         trs = []
         init_requires = self.init_requires
+        count = 1
```
nursix: This should probably be renamed into ```index```, because the next one is actually the "count", this one is an iteration index, and if you have both in the same loop, it's better to name them after what they are.

nursix:
In fact, I think the swap-buttons *should* be part of the tab order because they actually are the most important control in the form (manual edits are really rare).

huynhv:
Changed the code to adhere to code conventions, including variable names and spaces.
The Tab Order now is as follows: it walks through all attributes of one record, then loops through the swap buttons in the same order. It does not loop back to the fields of the other record since I don't think that's necessary.

nursix:
Looks good to me now, many thanks.

PR #1495, opened 2018-11-01
SAMBRO AlertHub minor fixes to pre-pop rss import and language codes

waidyanatha:
mostly update the feeds file.


flavour:
I suspect that changes to languages/* would not be desired...but take a look & see

Issue #1469, opened 2018-02-20
GIS: Selecting "Add PoI" and clicking into the map shows blank popup


trendspotter:
Steps to reproduce:
1. Go to GIS map
2. Click on the *Add PoI* buton in toolbar
3. Click into the map

Expected result:
Popup window with a form opens

Actual result:
![vystrizek](https://user-images.githubusercontent.com/15286128/36419140-5c624962-1631-11e8-8363-8cf451aef587.PNG)

The same goes for all other custom click-to-open-form functionality, such as in the *ARC* template or the commented out example with incident reports in the *default* template.

PR #1503, opened 2019-09-16
fix issue #1469

MehdioKhan:
jQuery selector to find popup content had something extra that removed. Now you can add POI from map.


MehdioKhan:
> Are you (or someone you are supporting) using this application or are you just trying to help out generally?

I'm helping out generally

flavour:
Are you up for trying to change the font to work via import_font to mean that we don't include it in the repo? & then do a clean rebase with single commit as a pull request?

nursix:
rtl.reshaper breaks tests for Python-3 - this needs fixing before merge

nursix:
Also, I'm not sure I would agree with adding the rtl library as dependency - it's unmaintained since 2016 and has no public repo. We could consider adding it to our repo, if it must, but if the BiDi variant works with the right fonts installed, then I'd prefer leaving it at that.

nursix:
By contrast - the bidi library is actively maintained, and has a source repository on GitHub. So definitely preferable over rtl.

PR #1511, opened 2019-10-15
Fix S3LocationSelector._labels when there are no locations

### LINKED ISSUE #1502: <type 'exceptions.TypeError'> 'NoneType' object has no attribute '__getitem__'
I am always getting this issue can anyone help me? Thank you in advance!

  File "C:\Users\Dell 5421\web2py\gluon\restricted.py", line 227, in restricted
    exec ccode in environment
  File "C:/Users/Dell 5421/web2py/applications/eden/controllers/org.py", line 505, in <module>
  File "C:\Users\Dell 5421\web2py\gluon\globals.py", line 417, in <lambda>
    self._caller = lambda f: f()
  File "C:/Users/Dell 5421/web2py/applications/eden/controllers/org.py", line 149, in facility
    return s3db.org_facility_controller()
  File "applications\eden\modules\s3db\org.py", line 7234, in org_facility_controller
    rheader = org_rheader,
  File "C:/Users/Dell 5421/web2py/applications/eden/models/00_utils.py", line 244, in s3_rest_controller
    output = r(**attr)
  File "applications\eden\modules\s3\s3rest.py", line 683, in __call__
  File "applications\eden\modules\s3\s3rest.py", line 1767, in __call__
  File "applications\eden\modules\s3\s3crud.py", line 104, in apply_method
    output = self.create(r, **attr)
  File "applications\eden\modules\s3\s3crud.py", line 400, in create
    format=representation,
  File "applications\eden\modules\s3\s3forms.py", line 1095, in __call__
    *formfields)
  File "C:\Users\Dell 5421\web2py\gluon\sqlhtml.py", line 1809, in factory
    **attributes)
  File "C:\Users\Dell 5421\web2py\gluon\sqlhtml.py", line 1286, in __init__
    inp = field.widget(field, default)
  File "applications\eden\modules\s3\s3widgets.py", line 4670, in __call__
    country=values.get("L0"),
  File "applications\eden\modules\s3\s3widgets.py", line 4937, in _labels
    d[int(level[1:])] = s3_str(T(row[level]))
TypeError: 'NoneType' object has no attribute '__getitem__'

hallamoore:
Fixes #1502

### DIFF
--- a/modules/s3/s3widgets.py
+++ b/modules/s3/s3widgets.py
@@ -5070,11 +5070,12 @@ class S3LocationSelector(S3Selector):
                         labels[level] = d[int(level[1:])] = label
         else:
             row = rows.first()
-            d = compact["d"] = {}
-            for level in levels:
-                if level == "L0":
-                    continue
-                d[int(level[1:])] = s3_str(T(row[level]))
+            if row:
+                d = compact["d"] = {}
+                for level in levels:
+                    if level == "L0":
+                        continue
+                    d[int(level[1:])] = s3_str(T(row[level]))
 
         return labels, compact
 
--- a/modules/unit_tests/s3/s3widgets.py
+++ b/modules/unit_tests/s3/s3widgets.py
@@ -10,7 +10,7 @@ from collections import OrderedDict
 from gluon import *
 from gluon.storage import Storage
 
-from s3.s3widgets import S3HoursWidget, S3OptionsMatrixWidget
+from s3.s3widgets import S3HoursWidget, S3LocationSelector, S3OptionsMatrixWidget
 
 from unit_tests import run_suite
 
@@ -542,11 +542,26 @@ class S3HoursWidgetTests(unittest.TestCase):
                 assertEqual(round(hours, 8), s[1], "'%s' recognized as %s, expected %s" % (s[0], hours, s[1]))
 
 # =============================================================================
+class S3LocationSelectorTests(unittest.TestCase):
+    """ Tests for S3LocationSelector """
+
+    # -------------------------------------------------------------------------
+    def test_empty_labels(self):
+        """ Test _labels when there aren't any locations """
+
+        location_selector = S3LocationSelector()
+        levels = ['L0', 'L1', 'L2', 'L3', 'L4', 'L5']
+        results = location_selector._labels(levels, country=None)
+        self.assertEqual(results[0], {'L0': current.T('Country')})
+        self.assertEqual(results[1], {})
+
+# =============================================================================
 if __name__ == "__main__":
 
     run_suite(
         S3OptionsMatrixWidgetTests,
         S3HoursWidgetTests,
+        S3LocationSelectorTests,
     )
 
 # END ========================================================================


modules/s3/s3widgets.py
```diff
@@ -5070,11 +5070,12 @@ def _labels(levels, country=None):
                         labels[level] = d[int(level[1:])] = label
         else:
             row = rows.first()
-            d = compact["d"] = {}
-            for level in levels:
-                if level == "L0":
-                    continue
-                d[int(level[1:])] = s3_str(T(row[level]))
+            if row:
```
nursix: I think we want this to crash because at least one gis_hierarchy is mandatory, and skipping the prepop will inevitably lead to problems elsewhere, even if it succeeds here.

Though, it may be useful to have a more meaningful error message here that points to the real problem.

modules/unit_tests/s3/s3widgets.py
```diff
@@ -541,12 +541,27 @@ def test_validate_implied_unit_limit(self):
                 assertEqual(error, None)
                 assertEqual(round(hours, 8), s[1], "'%s' recognized as %s, expected %s" % (s[0], hours, s[1]))
 
+# =============================================================================
+class S3LocationSelectorTests(unittest.TestCase):
+    """ Tests for S3LocationSelector """
+
+    # -------------------------------------------------------------------------
+    def test_empty_labels(self):
```
nursix: This test will never succeed because, like I said, the hierarchy prepop is mandatory and hence part of the default template these unit tests are run against. Makes no sense to have a test that relies on an invalid setup.

modules/s3/s3widgets.py
nursix: That said, catching the problem only here makes no sense either.

There must be at least one gis_hierarchy (the "SITE_DEFAULT") - and there will be if the template is configured correctly (specifically: settings.base.prepopulate is added to rather than replaced - it's a common mistake to replace the setting and thereby remove "default/base").

So, ideally this would be caught at the end of first_run - either by showing an error, or by enforcing the import of default/base/gis_hierarchy.csv if no gis_hierarchy is present (I'd opt for the latter).

hallamoore: Interesting! The example config has the prepop settings commented out, and leads with "Over-rides to the Template may be done here", which makes it sound like optional settings that aren't required for base functionality https://github.com/sahana/eden/blob/master/modules/templates/000_config.py#L229 

I'd also lean for using a default value if none is present. I'm gonna dig into what prepopulating is exactly a bit more. I assumed it was just populating fake data, but now I suspect it's defining a schema for the data? But then I'll see what I can do with this new direction.

Thanks for the context!

nursix: It is correct that you _may_ override settings.base.prepopulate, but then your alternative prepop must include the mandatory data sets. There are not very many of them, though.

----

More generally speaking: prepopulate means simply to populate the database with data during the first request cycle ("first run"). This is an integral part of the initialization routine for any Sahana instance: "first run" creates the database and all its tables, and then pre-populates it with some data. The exact procedure can be found in models/zzz_1st_run.py.

This does, in itself, not say what the pre-populated data are for - some of them may be required for the respective data models to function correctly (e.g. the SITE_DEFAULT gis_hierarchy), some could be standard data that are not strictly required but commonly used by all Sahana deployments (e.g. the set of countries for gis_location), and some are indeed fake data for testing.

A minimum data set containing the required data and some standard data is provided by "default/base" - and all templates would normally include this. And even if, for some odd reason, you absolutely do not want to include default/base, then you must make sure that your alternative prepop still provides the required base data.

Additionally:

Which data are initially loaded into a Sahana instance is primarily determined by the template's config.py (modules/templates/.../config.py). This is important because it may include other mandatory data sets beyond default/base that are required for the **template** to work properly.

Whilst the models/000_config.py may contain local overrides for that particular server - it is of course not required (nor very common) to override template settings, and thus, the prepopulate-setting (and any other overrides) in the example 000_config.py are commented out - just as they would be in most models/000_config.py.

----

So, **if** you override the template's prepopulate-setting in models/000_config.py - and this won't happen in a standard setup procedure - then you're doing this deliberately, and hence should know what you're doing - and then it simply is as it is: omitting mandatory data sets can stop certain functionality from working.

Where possible and meaningful, such functionality would be automatically disabled when required data sets are missing. The decision whether to gracefully auto-disable or raise an error is always tricky - but as a rule of the thumb: if it is obvious or most likely that auto-disable cannot be the user's expectation, then it is better to raise (and raise early!) to make the system administrator aware that the setup was incomplete/incorrect.

In this case, I'd argue that the absence of the SITE_DEFAULT gis_hierarchy cannot be deliberate, but must be a mistake. And for that reason, it should raise - not compensate by auto-disabling all location selectors. However, it should ideally raise already during first run.

nursix: P.S. And no - "overrides" does not mean "optional settings that are not required for base functionality". 

On the contrary: it really means to deliberately _override_ essential settings of the template, in order to alter functionality (including base functionality) on that particular server, away from what the programmers originally intended. 

And as such, these overrides always require advanced knowledge - both of how the settings work in general, and how changing them will affect the template's functionality.

The normal way, especially for inexperienced admins, is to run with template settings, and leave any changes in 000_config.py to the setup tool.

hallamoore: Woah, I was going nuts trying to figure out where `zzz_1st_run.py` was even invoked, and finally figured out that gluon runs every file in `models/`. Is that why there's weird `000`, `00`, `zz`, `zzz` file name prefixes, to make sure they get run in a specific order?

nursix: Yes, that's right - all files in models/ are run during every request cycle (before entering the controller, and in the same global environment as the controller). They are run in alphabetical order - so the name prefixes make sure of the correct execution order.

PR #1512, opened 2019-10-19
Prepopulate the gis_hiearchy with default values if none are provided

### LINKED ISSUE #1502: <type 'exceptions.TypeError'> 'NoneType' object has no attribute '__getitem__'

hallamoore:
The gis_hierarchy is required for the app to be functional. This
modifies the first run script to check that a gis_hiearchy has been
populated. If not, then the default gis_hierarchy is loaded.

Fixes #1502

### DIFF
--- a/models/zzz_1st_run.py
+++ b/models/zzz_1st_run.py
@@ -314,6 +314,17 @@ if len(pop_list) > 0:
                 s3_str = s3base.s3_str
                 info("\n".join(s3_str(el) for el in error))
 
+    # Check to see if the "SITE_DEFAULT" gis_hierarchy was prepopulated.
+    # Use our default gis_hierarchy if not.
+    table = s3db.gis_hierarchy
+    query = (table.uuid == "SITE_DEFAULT")
+    row = db(query).select(table.id, limitby=(0, 1)).first()
+    if not row:
+        info("\nWarning: No gis_hierarchy provided, using default.")
+        csv = path_join(request_folder, "modules", "templates", "default", "base", "gis_hierarchy.csv")
+        xsl = path_join(request_folder, "static", "formats", "s3csv", "gis", "hierarchy.xsl")
+        bi.execute_import_task([1, "gis", "hierarchy", csv, xsl, None])
+
     info("\nUpdating database...")
 
     # Restore setting for strict email-matching


models/zzz_1st_run.py
```diff
@@ -314,6 +314,16 @@ def duration(msg, start):
                 s3_str = s3base.s3_str
                 info("\n".join(s3_str(el) for el in error))
 
+    # Check to see if the "SITE_DEFAULT" gis_hierarchy was prepopulated.
+    # Use our default gis_hierarchy if not.
+    query = (current.s3db.gis_hierarchy.uuid == "SITE_DEFAULT")
+    rows = current.db(query).select(limitby=(0, 1))
+    if len(rows) == 0:
+        info("\nWarning: No gis_hierarchy provided, using default.")
+        csv = "applications/eden/modules/templates/default/base/gis_hierarchy.csv"
```
nursix: We cannot hardcode the application name "eden" - must use current.request.application

nursix: Or actually, current.request.folder.

nursix: Like this:
```
folder = current.request.folder
csv = os.path.join(folder, "modules", "templates", "default", "base", "gis_hierarchy.csv")
xsl = os.path.join(folder, "static", "formats", "s3csv", "gis", "hierarchy.csv")
```

models/zzz_1st_run.py
```diff
@@ -314,6 +314,16 @@ def duration(msg, start):
                 s3_str = s3base.s3_str
                 info("\n".join(s3_str(el) for el in error))
 
+    # Check to see if the "SITE_DEFAULT" gis_hierarchy was prepopulated.
+    # Use our default gis_hierarchy if not.
+    query = (current.s3db.gis_hierarchy.uuid == "SITE_DEFAULT")
+    rows = current.db(query).select(limitby=(0, 1))
```
nursix: This is inefficient, as it will extract all fields whilst we only want to know whether the record exists or not.

Should be like:
```
table = current.s3db.gis_hierarchy
rows = current.db(table.uuid == "SITE_DEFAULT").select(table.id, limitby=(0, 1)).first()
if not row:
    ...
```

models/zzz_1st_run.py
hallamoore: Got it! Does this file get copied to other application folders for deployments?

models/zzz_1st_run.py
hallamoore: Ah, good call, thanks!

nursix:
This looks correct to me now, good to merge if @flavour agrees.

----

However, let's think one more time here: 

Since the default gis_hierarchy is just one record - do we really need to use the importer to get it in, or could we just create it directly (which is only a fraction of the effort)?

The point with SITE_DEFAULT is that the default location hierarchy can be template-specific - but if the template does not provide any SITE_DEFAULT, then this would make it fall back to the standard.

Now, this standard has been the same ever since it has been introduced, so I'd see it as very unlikely that it will change in the future - but **if** it would change, and we had hardcoded it here, then we would have to change it in two places: in the CSV and here in this fallback.

And we don't like to have the same thing in two places.

But - if we would hard-code the standard SITE_DEFAULT in zzz_1st_run.py and insert it into the table before running the imports, then any possible imported SITE_DEFAULT from the template would override it - and if there is none imported, then it would remain standard as-is.

Or, in other words: templates would only need to import the SITE_DEFAULT if they do not want to use the standard default. That would make default/gis_hierarchy.csv obsolete, and then it's only one place to change the standard default...if that now ever happens.

This strategy would also save us one import during first run in most templates, as the SITE_DEFAULT is indeed rarely overridden.

I'd like to hear @flavour 's opinion on that. Still, this one looks good to merge for a start.

hallamoore:
Somewhat unrelatedly, but I was wondering why the gis_hiearchy is stored in a database table. I still don't completely understand everything the gis_hierarchy is used for, so I might be completely off base, but so far it seems like it's only set on the first run and values aren't inserted or updated afterwards. Could the hierarchy perhaps be 
1. defined in a file beforehand 
2. read from file on app start-up and stored in memory

instead of 

1. defined in a file beforehand 
2. imported into database on first run 
3. read from database on each request

?

nursix:
Technically, there is no defined "app start-up" in Sahana - as it's not running permanently. In normal deployments the interpreter is started on-demand, and shut down after some idle time. The only thing you have as a trigger is the current HTTP request cycle - and everything must happen within its boundaries.

That said, we do not actually read the gis_hierarchy from the database for every request cycle. After it has been read from the database once, it's indeed _cached_ in memory for re-use by subsequent request cycles in the same interpreter instance. Additionally, the database caches the query result (if so configured), which is extremely efficient even across interpreter boundaries.

That we have the gis_hierarchy in the database at all is simply because it's a table - one hierarchy per country plus said SITE_DEFAULT (all of which can be template-specific, or even deployment-specific - and generally editable).

In a database-driven application, the database is the obvious place for a table like that - and there is simply no good argument to have it in a file instead. And given the efficient caching, I think we already have the right solution.

hallamoore:
@nursix That makes sense, thanks!

hallamoore:
@flavour Updated! Are all `current` attributes already in scope in the models context?

flavour:
Thanks, merged :)
I wouldn't assume all of current.* are in scope, but many are, familiarity will tell which and meanwhile you can easily test through the interactive Web2Py shell:
python web2py.py -S eden -M

PR #1515, opened 2019-10-29
Don't URLise CMS content if <a> tags already exist

### LINKED ISSUE #1452: CMS mangles links in HTML posts
I have found the issue when I've tested the SAMBRO RSS and CAP feeds, but later found out that basically any CMS post with HTML link in it exhibits the problem. Whenever a HTML link is present in the post, CMS reparses it again in some weird way, creating a new link out of the `href` part while leaving the surrounding code unchanged. Eg from
```
<a href="http://blabla">blabla</a>
```
it creates
```
<a href="<a href=" http: blabla"">http://blabla"</a>"blabla
```

See the screenshot:

sambro

hallamoore:
Fixes #1452

### DIFF
--- a/modules/s3db/cms.py
+++ b/modules/s3db/cms.py
@@ -47,6 +47,7 @@ __all__ = ("S3ContentModel",
 import datetime
 import json
 import os
+import re
 
 from gluon import *
 from gluon.storage import Storage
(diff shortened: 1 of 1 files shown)


nursix:
P.S. would you want to complement with a unit test? One thing to pay a little extra attention to is multi-byte unicode - although this is only relevant for Python-2.7, which is on the way out. Not a requirement for merge, just a thought.

hallamoore:
> would you want to complement with a unit test?
Yeah, I'll circle back and add one!

Issue #1513, opened 2019-10-27
GIS config missing


hallamoore:
The form at `eden/org/office/create` isn't displayed when there is no GIS config. There weren't any errors displayed on the webpage, but with a pdb breakpoint in the application I eventually checked out `s3db.org_office_controller()['form'].xml()` and saw an HTTP 303 printed. The pdb session wasn't enough to get a traceback, but I was able to get one by opening a `code.interact` session within the pdb session and running the same thing. There were a bunch of recursive xml calls, but the bottom of the traceback looks like:

```
File "applications/eden/modules/s3/s3gis.py", line 7124, in xml
  result = self._setup()
File "applications/eden/modules/s3/s3gis.py", line 6443, in _setup
  redirect(URL(c="default", f="index"))
File "/home/halla/sahana-eden/web2py/gluon/http.py", line 182, in redirect
  **headers)
HTTP: 303 SEE OTHER
```

So it appears that my GIS config is missing.

Question 1: 
How do I include a GIS config? I tried switching my `settings.base.prepopulate` to `["default"]`, but that resulted in a bunch of `gis_layer_config: config_id: value(s) required` lines, and no change in behavior for the form.

Question 2:
Similar to how we included a default hiearchy in #1512, is there some default I could try to implement here? (Maybe not, I don't know what all is included with the GIS config, so maybe a default doesn't make any sense. Which leads to question 3...)

Question 3:
It looks like there's a `current.session.error = current.T("Map cannot display without prepop data!")` line right before the redirect, but I only see this error message when I
navigate to the next page. Is there a way to get this to appear on the current page?

nursix:
Rephrasing the problem:

The map cannot be shown without a GIS config - which is simply because the GIS config determines what is to appear on the map (e.g. which base layer).

That no GIS config is provided during prepop is not wrong - the system administrator may choose to not import one during prepop, but create one interactively instead. Until they do so, of course, the map cannot be displayed - but the system will work without crashing, and that is where it differs from #1512.

Thus, the correct response of the system to a missing gis_config is indeed be to indicate that no map can be displayed unless/until a gis_config is created.

---

As for your questions:

Q1: a standard basic gis_config (along with a bunch of layers) is provided by the "default/base" prepop - which is in the settings.base.prepopulate setting by default. If you extend the setting (```+=```) rather than overriding it (```=```), you shouldn't have any problems with that.

If you override the setting, then you can include "default/base" as first element of the array.

[...]

PR #1516, opened 2019-10-31
Display error in map component if there is no GIS config

hallamoore:
Displaying the error in the map component allows the rest of the webpage
to load normally and removes an unsafe redirect from a lower level
function, as suggested in #1513.

This commit also moves the instantiation of the loader component into
the `_setup()` method after we've fetched the config, to ensure that the
error message doesn't appear with a permanent loading icon above it.

### DIFF
--- a/modules/s3/s3gis.py
+++ b/modules/s3/s3gis.py
@@ -6440,6 +6440,7 @@ class MAP(DIV):
             # No prepop - Bail
             if auth.s3_has_permission("create", "gis_hierarchy"):
                 error_message = DIV(_class="mapError")
+                # Deliberately not T() to save unneccessary load on translators
                 error_message.append("Map cannot display without GIS config!")
                 error_message.append(XML(" (You can can create one "))
                 error_message.append(A("here", _href=URL(c="gis", f="config")))
@@ -6447,7 +6448,7 @@ class MAP(DIV):
                 self.error_message = error_message
             else:
                 self.error_message = DIV(
-                    "Map cannot display without GIS config!",
+                    "Map cannot display without GIS config!",  # Deliberately not T() to save unneccessary load on translators
                     _class="mapError"
                     )
             return None
@@ -7597,9 +7598,22 @@ class MAP2(DIV):
 
         if options is None:
             # No Map Config: Just show error in the DIV
-            self.components = [DIV("Map cannot display without prepop data!", # Deliberately not T() to save unneccessary load on translators
-                                   _class="error"),
-                               ]
+            auth = current.auth
+
+            if auth.s3_has_permission("create", "gis_hierarchy"):
+                error_message = DIV(_class="mapError")
+                # Deliberately not T() to save unneccessary load on translators
+                error_message.append("Map cannot display without GIS config!")
+                error_message.append(XML(" (You can can create one "))
+                error_message.append(A("here", _href=URL(c="gis", f="config")))
+                error_message.append(")")
+            else:
+                error_message = DIV(
+                    "Map cannot display without GIS config!",  # Deliberately not T() to save unneccessary load on translators
+                    _class="mapError"
+                    )
+
+            self.components = [error_message]
             return super(MAP2, self).xml()
 
         map_id = self.opts.get("id", "default_map")
--- a/modules/s3/s3gis.py
+++ b/modules/s3/s3gis.py
@@ -6432,7 +6432,6 @@ class MAP(DIV):
         # Fresh _setup() call, reset error message
         self.error_message = None
 
-        T = current.T
         auth = current.auth
 
         # Read configuration
@@ -6440,19 +6439,15 @@ class MAP(DIV):
         if not config:
             # No prepop - Bail
             if auth.s3_has_permission("create", "gis_hierarchy"):
-                error_message = DIV()
-                error_message.append(SPAN(
-                    T("Map cannot display without GIS config!"),
-                    _class="mapError"
-                    ))
-                error_message.append(XML(T(
-                    " (You can can create one <a href='%(url)s'>here</a>)" %
-                    {"url": URL(c="gis", f="config")}
-                    )))
+                error_message = DIV(_class="mapError")
+                error_message.append("Map cannot display without GIS config!")
+                error_message.append(XML(" (You can can create one "))
+                error_message.append(A("here", _href=URL(c="gis", f="config")))
+                error_message.append(")")
                 self.error_message = error_message
             else:
                 self.error_message = DIV(
-                    T("Map cannot display without GIS config!"),
+                    "Map cannot display without GIS config!",
                     _class="mapError"
                     )
             return None
--- a/modules/s3/s3gis.py
+++ b/modules/s3/s3gis.py
@@ -6387,6 +6387,8 @@ class MAP(DIV):
         # We haven't yet run _setup()
         self.setup = False
         self.callback = None
+        self.error_message = None
+        self.components = []
 
         # Options for server-side processing
         self.opts = opts
@@ -6395,16 +6397,6 @@ class MAP(DIV):
 
         # Options for client-side processing
         self.options = {}
-
-        # Components
-        # Map (Embedded not Window)
-        components = [DIV(DIV(_class="map_loader"),
-                              _id="%s_panel" % map_id)
-                      ]
-
-        self.components = components
-        for c in components:
-            self._setnode(c)
 
         # Adapt CSS to size of Map
         _class = "map_wrapper"
@@ -6437,16 +6429,36 @@ class MAP(DIV):
               into scripts (callback or otherwise)
         """
 
+        # Fresh _setup() call, reset error message
+        self.error_message = None
+
+        T = current.T
+        auth = current.auth
+
         # Read configuration
         config = GIS.get_config()
         if not config:
             # No prepop - Bail
-            current.session.error = current.T("Map cannot display without prepop data!")
-            redirect(URL(c="default", f="index"))
+            if auth.s3_has_permission("create", "gis_hierarchy"):
+                error_message = DIV()
+                error_message.append(SPAN(
+                    T("Map cannot display without GIS config!"),
+                    _class="mapError"
+                    ))
+                error_message.append(XML(T(
+                    " (You can can create one <a href='%(url)s'>here</a>)" %
+                    {"url": URL(c="gis", f="config")}
+                    )))
+                self.error_message = error_message
+            else:
+                self.error_message = DIV(
+                    T("Map cannot display without GIS config!"),
+                    _class="mapError"
+                    )
+            return None
 
         T = current.T
         db = current.db
-        auth = current.auth
         s3db = current.s3db
         request = current.request
         response = current.response
@@ -6480,6 +6492,12 @@ class MAP(DIV):
                 "gis_too_many_features": T("There are too many features, please Zoom In or Filter"),
                 "gis_zoomin": T("Zoom In"),
                 }
+
+        ##########
+        # Loader
+        ##########
+
+        self.append(DIV(DIV(_class="map_loader"), _id="%s_panel" % self.id))
 
         ##########
         # Viewport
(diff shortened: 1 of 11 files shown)


hallamoore:
I'd like to add a test to make sure that `_setup()` returns `None` and that the `error_message` is not `None` when there's no GIS config. What's the best way for me to monkeypatch the `GIS.get_config()`?

hallamoore:
Whoops, I also forgot to remove the translations of the previous error message. I gotta head out right now but I'll update that later.

modules/s3/s3gis.py
```diff
@@ -6434,19 +6426,36 @@ def _setup(self):
...
+        T = current.T
+        auth = current.auth
 
         # Read configuration
         config = GIS.get_config()
         if not config:
             # No prepop - Bail
-            current.session.error = current.T("Map cannot display without prepop data!")
-            redirect(URL(c="default", f="index"))
+            if auth.s3_has_role(current.session.s3.system_roles.MAP_ADMIN):
```
nursix: This should rather check for permission to create (because other roles can have permission too):
```
if auth.s3_has_permission("create", "gis_hierarchy"):
```

modules/s3/s3gis.py
```diff
@@ -6434,19 +6426,36 @@ def _setup(self):
...
+                error_message = DIV()
+                error_message.append(SPAN(
+                    T("Map cannot display without GIS config!"),
+                    _style='color:red;'
+                ))
+                error_message.append(' (You can can create one ')
```
nursix: Should be a single T(), including the link tag, like:

```
XML(T("(You can create one <a href='%(url)s'>here</a>)") % {"url": URL(c="gis", f="config")})
```

Translators are advised to leave HTML tags untouched.

modules/s3/s3gis.py
```diff
@@ -6434,19 +6426,36 @@ def _setup(self):
...
         if not config:
             # No prepop - Bail
-            current.session.error = current.T("Map cannot display without prepop data!")
-            redirect(URL(c="default", f="index"))
+            if auth.s3_has_role(current.session.s3.system_roles.MAP_ADMIN):
+                error_message = DIV()
+                error_message.append(SPAN(
+                    T("Map cannot display without GIS config!"),
+                    _style='color:red;'
+                ))
```
nursix: Closing brackets should have the same indent as the enclosed lines.

modules/s3/s3gis.py
```diff
@@ -6434,19 +6426,36 @@ def _setup(self):
...
         config = GIS.get_config()
         if not config:
             # No prepop - Bail
-            current.session.error = current.T("Map cannot display without prepop data!")
-            redirect(URL(c="default", f="index"))
+            if auth.s3_has_role(current.session.s3.system_roles.MAP_ADMIN):
+                error_message = DIV()
+                error_message.append(SPAN(
+                    T("Map cannot display without GIS config!"),
+                    _style='color:red;'
```
nursix: We use double quotes throughout. Semicolon is not required here. Ideally use a class and allow the style to be set in style sheets rather than hard-coding the style here.

modules/s3/s3gis.py
```diff
@@ -6434,19 +6426,36 @@ def _setup(self):
             - separated from xml() in order to be able to read options to put
               into scripts (callback or otherwise)
         """
+        # Fresh _setup() call, reset error message
```
nursix: Blank line after docstring for better readability and consistency of coding style.

nursix:
The easiest thing for a unit test would be to indeed the delete all gis_config in the test setup, test the fallback, and then roll back the DB transaction (=the deletion) in the tearDown - i.e. test the fallback inside an uncommitted transaction.

Obviously, you'll have to remove the cached config (current.response.s3.gis.config), and restore it in the tearDown (using current.gis.get_config).

modules/s3/s3gis.py
hallamoore: Ooo, that's useful!

modules/s3/s3gis.py
hallamoore: Alright, I wasn't sure if it was bad practice to manually include HTML instead of invoking web2py's HTML classes. Are there any guidelines for when to use them or not?

modules/s3/s3gis.py
```diff
@@ -6435,18 +6427,37 @@ def _setup(self):
...
+                    ))
+                error_message.append(XML(T(
+                    " (You can can create one <a href='%(url)s'>here</a>)" %
+                    {"url": URL(c="gis", f="config")}
+                    )))
+                self.error_message = error_message
+            else:
+                self.error_message = DIV(
+                    T("Map cannot display without GIS config!"),
+                    _class="mapError"
```
hallamoore: So I changed this to a class, but I can't figure out where the best place to add the css definition is. I tried both `eden/static/styles/gis/style.css` and `eden/static/scripts/gis/openlayers/theme/default/style.css`, but the results didn't show up on the webpage. I also had `settings.base.debug = True` in my config, in case it was because my change wasn't reflected in the minimized files, but no dice. Any suggestions?

hallamoore:
Side question, I've been force-pushing because I'm used to a code review work-flow that involves a single commit per diff, but with github pull requests I don't see a way to view the changes between pushes. Should I be using separate commits per update so the changes are easier to view here?

modules/s3/s3gis.py
flavour: > I wasn't sure if it was bad practice to manually include HTML instead of invoking web2py's HTML classes. Are there any guidelines for when to use them or not?

The HTML classes are lovely, but not critical...the critical thing is that i18n strings should be as entire as possible as translations can vary a lot by context & positioning.

modules/s3/s3gis.py
flavour: debug is definitely required to see raw CSS that is normally minified.
Which CSS get loaded is defined by your template/css.cfg
styles/gis/style.css is loaded in most templates including the default, so very odd it didn't show for you there
scripts/gis/openlayers/theme/default/style.css should generally be left as-is

For this case, I would use themes/default/widgets.css & in fact I think we can re-use the normal .error class here (makes bold as well as red)

flavour:
> I've been force-pushing because I'm used to a code review work-flow that involves a single commit per diff, but with github pull requests I don't see a way to view the changes between pushes. Should I be using separate commits per update so the changes are easier to view here?

Great that you're asking the question :)
In general if we're reviewing a change from scratch then best if we're reviewing just a single commit.
However the other reason we used to have of minimising the number of commits ion trunk is gone as we can easily squash when merging...so if we're mid-review and there are some tweaks being done  then I would certainly be happy to now see these as separate commits as we can then review just the delta (as well as the review history being clearer)

modules/s3/s3gis.py
hallamoore: Ah, that makes a lot of sense!

modules/s3/s3gis.py
hallamoore: Hmm, I was using an improper prepopulate setting to trigger the error message, so maybe I also caused it to leave out that css file.

Anyways, when I use just `error`, the css styles get overwritten by https://github.com/sahana/eden/blob/master/static/themes/default/theme.css#L338 (and also https://github.com/sahana/eden/blob/master/static/themes/default/foundation/foundation.css#L4379), so it looks like this instead: 
image

I think that disconnects the "(You can create one here)" part too much, so adding another error class name might be easiest. I can just add the new class name to the definition with `error` so that they have the same exact style. Any preferences on what I name that, or should I go down an entirely different route?

modules/unit_tests/s3/s3gis.py
```diff
@@ -773,11 +773,38 @@ def tearDownClass(cls):
         current.auth.override = False
         current.db.rollback()
 
+# =============================================================================
+class S3NoGisConfigTests(unittest.TestCase):
+    # -------------------------------------------------------------------------
+    @classmethod
+    def setUpClass(cls):
+        current.s3db.event_config.truncate()
```
nursix: Goodness - this looks complicated. Would it be an alternative idea to temporarily overwrite current.gis.get_config with a dummy that always returns None?

modules/s3/s3gis.py
nursix: I think you should not use the "error" class at all - as it's specific for form field errors.

Have a specific CSS class for this particular case instead (like "map-no-config") - and in the default theme, let it have a (red) border around it with the message inside. I don't think the message itself needs to be red.

nursix: P.S. the important part is to leave the styling to the theme. I don't see a strict need to provide a default style in widgets.css.

In fact I really don't like the idea of having a widgets.css base style for this class - for the same reason as I dislike most other styles in widgets.css: maps can appear in different places (standalone, embedded, in forms etc), so the error message may need different formatting depending on context and DOM hierarchy. That can be very fiddly if there is a base style in widgets.css that needs overriding - and since such styles are often overridden in themes, it's also pointless download volume.

I'd say don't provide a base style, leave it entirely to the theme to style it.

nursix: To explain that a little: we generally have a cascade of widgets.css=>theme.css(=>style.css).

The point with widgets.css is to provide common base styles for complex widgets in order to reduce the complexity of themes, and to ensure that functional styles (e.g. visibility and placement of widget components) are present and consistent in all themes.

The actual look, on the other hand - coloring, font, spacing etc. - should be left to the theme, and be controlled by semantic CSS classes that are not tied to specific layouts, because different themes can - and often do - use different layouts (e.g. formstyle, custom view templates, inline component layouts, data list layouts).

If widgets.css specifies looks, and especially when those looks assume certain outer DOM hierarchies, then overriding them in the theme can be quite fiddly - and since such looks are likely to be overridden in theme, they become pointless ballast in the download.

In this case, the empty map DIV has no inner layout, so there is no strict reason to provide visibility and placement styles in widgets.css. However, it makes sense to provide a standard style in the default's theme SCSS - because these SCSS sources are usually included when building custom themes.

flavour: > really don't like the idea of having a widgets.css base style for this class

Whilst I don't disagree with the logic presented here, I do have a counter: This theory is lovely but in practise no themes are going to invest time in styling an edge condition which will likely never affect them & if it did happen to would only affect them during a very early development stage.
I therefore think that this is a good candidate for being in core CSS.

Similarly I actually think we don't need T() for such an error message (save some load on our poor, overworked translators)

nursix: Agreed - edge case ;)

Still, the general guideline should be that widgets.css should provide functional base styles for UI elements and be limited to their inner DOM tree, so that (hypothetically) the UI element would still be recognizable and functional with no theme CSS.

However, it does not need to look good with the base style alone - so be minimalist there, no polishing/pixel-pushing in widgets.css please.

modules/unit_tests/s3/s3gis.py
hallamoore: Ooo, yes!

hallamoore:
That force-push was a rebase to resolve conflicts. 55740b5 and 1be44ab  are the recent changes

flavour:
Looks good, but the Travis tests are failing...seems to be in your new tests, so can you check those?

modules/unit_tests/s3/s3gis.py
```diff
@@ -773,11 +773,37 @@ def tearDownClass(cls):
         current.auth.override = False
         current.db.rollback()
 
+# =============================================================================
+class S3NoGisConfigTests(unittest.TestCase):
+    # -------------------------------------------------------------------------
+    @classmethod
+    def setUpClass(cls):
+        cls.original_get_config = current.gis.get_config
+        current.gis.get_config = lambda: None
```
nursix: I'm afraid MAP._setup calls `GIS.get_config` as class method rather than `current.gis.get_config` as instance method.

Try putting the dummy in the class instead?

hallamoore: Oh! I'll try that. Mocking the instance method passes locally for me though, I wonder why that is. :thinking:

flavour:
Thankyou!

PR #1520, opened 2019-11-19
Use S3PentityAutocompleteWidget for "Person or OU" field in map profile

### LINKED ISSUE #1425: Map profile not showing organization
I added an organization (say org X) using org module. Then I created an user and assign him to the org X. Then in map profile, I tried to create a profile of the organization. However, the `Person or OU` field is only loading person and not the org X. Is this the default behaviour? How do I create map profile for the organization so that every user of the organization get the map profile of the organization?

hallamoore:
And specify that organisations can be part of the autocomplete as well.

Fixes #1425

### DIFF
--- a/static/scripts/S3/s3.autocomplete.js
+++ b/static/scripts/S3/s3.autocomplete.js
@@ -721,14 +721,16 @@
         if (min_length == undefined) {
             min_length = 2;
         }
-        if (types) {
-            url += '?types=' + types;
-        }
 
         dummy_input.autocomplete({
             delay: delay,
             minLength: min_length,
             source: function(request, response) {
+                url = real_input.data('url')
+                if (types) {
+                    url += '?types=' + types;
+                }
+
                 // Patch the source so that we can handle No Matches
                 $.ajax({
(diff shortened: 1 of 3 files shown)


static/scripts/S3/s3.autocomplete.js
```diff
@@ -731,7 +731,7 @@
             source: function(request, response) {
                 // Patch the source so that we can handle No Matches
                 $.ajax({
-                    url: real_input.data('url'),
+                    url: url,
```
nursix: Nope - the idea is that it shall be possible to change the url after setup from the outside by modifying real_input.data, so the source callback must read it from there.

hallamoore: Ah, I see :+1:

nursix: Better.

Generally, extending a URL query fragment by blindly appending a '?xyz' to it isn't the best way to do it - but since it was done that way before, and fixing it would be out of context for this PR, I think it can be left as-is.

PR #1521, opened 2019-11-19
SAMBRO: Display error DIV if map layer is missing

### LINKED ISSUE #1344: SAMBRO Homepage 500 Error
Error trace 

```
  File "/var/www/web2py/gluon/restricted.py", line 227, in restricted
    exec ccode in environment
  File "file stream", line 425, in <module>
  File "/var/www/web2py/gluon/globals.py", line 430, in write
    self.body.write(xmlescape(data))
  File "/var/www/web2py/gluon/html.py", line 131, in xmlescape
    return data.xml()
  File "applications/eden/modules/s3/s3gis.py", line 7098, in xml
    self._setup()
  File "applications/eden/modules/s3/s3gis.py", line 6913, in _setup
    options["feature_resources"] = addFeatureResources(feature_resources)
  File "applications/eden/modules/s3/s3gis.py", line 7515, in addFeatureResources
    url = layer["url"]
KeyError: 'url'
```

hallamoore:
Fixes #1344

### DIFF
--- a/modules/templates/SAMBRO/controllers.py
+++ b/modules/templates/SAMBRO/controllers.py
@@ -51,7 +51,7 @@ class index(S3CustomController):
             current.log.error("Cannot find Layer for Map")
             layer_id = None
             output["_map"] = DIV(
-                "Cannot find Layer for Map",  # Deliberately not T() to save unneccessary load on translators
+                T("Cannot find Layer for Map"),
                 _class="mapError"
                 )
         else:
--- a/modules/templates/SAMBRO/controllers.py
+++ b/modules/templates/SAMBRO/controllers.py
@@ -50,25 +50,29 @@ class index(S3CustomController):
         except:
             current.log.error("Cannot find Layer for Map")
             layer_id = None
+            output["_map"] = DIV(
+                "Cannot find Layer for Map",  # Deliberately not T() to save unneccessary load on translators
+                _class="mapError"
+                )
+        else:
+            feature_resources = [{"name"      : T("Alerts"),
+                                  "id"        : "search_results",
+                                  "layer_id"  : layer_id,
+                                  # @ToDo: Make the filter update timestamp on refresh in the client side
+                                  "filter"    : "~.info.expires__gt=%s&~.external__ne=True" % request.utcnow,
+                                  # We activate in callback after ensuring URL is updated for current filter status
+                                  "active"    : False,
+                                  }]
 
-        feature_resources = [{"name"      : T("Alerts"),
-                              "id"        : "search_results",
-                              "layer_id"  : layer_id,
-                              # @ToDo: Make the filter update timestamp on refresh in the client side
-                              "filter"    : "~.info.expires__gt=%s&~.external__ne=True" % request.utcnow,
-                              # We activate in callback after ensuring URL is updated for current filter status
-                              "active"    : False,
-                              }]
-
-        _map = current.gis.show_map(callback='''S3.search.s3map()''',
-                                    catalogue_layers=True,
-                                    collapsed=True,
-                                    feature_resources=feature_resources,
-                                    save=False,
-                                    search=True,
-                                    toolbar=True,
-                                    )
-        output["_map"] = _map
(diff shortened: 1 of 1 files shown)


nursix:
Looks good - except that in this case, we do want T() for the error message ;) so if you could fix that either before or after merge, would be nice.

hallamoore:
Sure thing! Is that because this error message will be more common than the ones that we didn't use `T()` on? (Running the SAMBRO preprop loaded the layers for me, so I wasn't sure if it would be)

nursix:
Yes, unlike a missing gis_config, a missing CAP layer is not really an edge-case.

CAP is not default, SAMBRO is often set up with custom configs (and hence, custom prepops), and then map layers can be edited/deleted after prepop - so an accidental loss of the layer configis quite a possibility, and has happened before.

Besides, this is not looking for _a_ layer config, but for one that is specific for this template. And template-specific should always be translatable.

nursix:
To put it differently: a missing gis_config is a developer mistake, whereas a missing CAP layer can be an administrator error. And whilst developers will (necessarily) understand English, administrators may expect/need translated messages.

In fact, this error message is a bit calling for 2nd-level support - it doesn't give much of a hint as to what layer is missing, or how the problem can be solved. Could be improved (not giving you this as a task, just saying).

flavour:
> whilst developers will (necessarily) understand English, administrators may expect/need translated messages.

More so than developers certainly, however I confess that even here I would err on the side of reducing the load on overburdened translators. To solve such error messages, typically a Google search would be used to look for an answer & so having these all related answers within 1 language would consolidate them...sure this could be different if we see a major take-up of a template within a  country then a local community around that template would develop & so we'd want the answers to be in the admin's own language, however at this stage I'd say we're best off all in English...especially for the SAMBRO template of which there will only ever be a few max per country...sure there are several countries which speak Spanish & French however I'd still argue that at this stage no T() is better than T() here overall ;)

nursix:
This argument holds if we leave the error message as it is, and, for it's lack of clues how to overcome the issue, necessitate the Google search or a support request / issue report.

However, what you would find when searching for an error message in English would be answers in English - and if the error message wasn't helpful for being English, then how would the Google search be any better? Remember that Myanmar folks (and those are the ones deploying SAMBRO currently) even shun the mailing list for their poor command of the English language.

So, if we improve the error message with a hint how to resolve the situation - and then translators indeed do translate it, then we're doing them a greater service. And in all honesty, that's what translations are for - sometimes a little load on translators is the lesser problem.

Besides, the mere existence of a T() in a template (! not core) does not generally increase the load for translators - as it wouldn't appear in translation downloads unless you're targeting SAMBRO in particular. A T() makes a message translatable - but does in itself not necessitate the translation.

flavour:
> what you would find when searching for an error message in English would be answers in English

Yes, of course, however an English answer is better than no answer at all, which was my argument.

> wouldn't appear in translation downloads unless you're targeting SAMBRO in particular

or 'all templates' as a community contribution

> A T() makes a message translatable - but does in itself not necessitate the translation.

True, however, it takers a ton of effort to work out which strings are important to translate & which aren't...I have spent many hours on this many times over.

nursix:
The prospect of an English answer may be as good as no answer ;)

Imagine the situation in reverse - if user-facing error messages, documentation and support were entirely in Russian or Chinese, you would probably not deploy the software in the first place, as resolving issues would always be an uphill battle. I wouldn't.

I'm not really afraid to add this one error message to the workload of translators, especially "community contributions" which have no pressure or obligations at all - but I am worried to make administrators struggle with problems due to a language barrier.

flavour:
Admin-only stuff like this is not what I would call user-facing typically...this often gets left out of translations if we can separate them.
I often see/feel the pain of translators.

Anyway, I left it as T() although won't go into normal exports, so it can still be added

nursix:
I think we can afford different opinions in this case - and @hallamoore can have her own too ;)

hallamoore:
Very interesting! I think I don't know enough about translator load to have a thoroughly informed opinion, but these are all great points to think about.

PR #1527, opened 2019-12-04
SAMBRO Demo update with some Fiji data

waidyanatha:
Add some CAP templates into cap_alert.csv and cap_info.csv. 
Add two languages fo Fiji.
Commented out some, what I feel, redundant code? involves setting the homepage title


modules/templates/SAMBRO/Demo/users.csv
```diff
@@ -1,13 +1,11 @@
...
-Maldives,Alert Approver,mv_approver@example.com,eden,ALERT_APPROVER,National Disaster Management Center
-Nepal,Alert Approver,np_approver@example.com,eden,ALERT_APPROVER,Department of Hydrology and Meteorology
+Map Admin,Fiji MRD,fj_mapadmin@example.com,testing,MAP_ADMIN,Mineral Resource Department
+Alert Editor,Suva WW,sww_editor@example.com,testing,ALERT_EDITOR,Suva Water Works
+Alert Approver,Suva WW,sww_approver@example.com,testing,ALERT_APPROVER,Suva Water Works
+Alert Editor,FSM,fsm_editor@example.com,testing,ALERT_EDITOR,Fiji Meteorological Service
+Alert Approver,FSM,fsm_approver@example.com,testing,ALERT_APPROVER,Fiji Meteorological service
+Alert Editor,MRD,mrd_editor@example.com,testing,ALERT_EDITOR,Suva Water Works
+Alert Approver,MRD,mrd_approver@example.com,testing,ALERT_APPROVER,Suva Water Works
+Normal,User,normal_user@example.com,testing,ANONYMOUS,
```
nursix: @waidyanatha : the ANONYMOUS and AUTHENTICATED user roles cannot be assigned - they are automatically given to users depending on their login status:

- any user receives the ANONYMOUS role, regardless whether they are logged-in or not
- any logged-in user additionally receives the AUTHENTICATED role, regardless which account they login with

The rationale behind this is this:
1) Permissions given to the ANONYMOUS role are those which do not require login - whereas permissions given to AUTHENTICATED do require login, but just that and nothing more.
2) Any user, irrespective their login status, shall have at least the permissions of an anonymous user (or, turn it around: anonymous users cannot have more permissions than authenticated users).

Therefore, these roles are auto-assigned depending on login-status - and they are not realm-restricted.

nursix: I removed these two from the file in trunk.

PR #1531, opened 2020-01-16
Bug Fix IS_PHONE_NUMBER Validator Error Messages

aeturnum:
Control flow previously guaranteed default international error
message would never be displayed. The fix refactors the validation
function to be easier to follow and have fewer return points.

All unit tests pass.

### DIFF
--- a/modules/s3/s3validators.py
+++ b/modules/s3/s3validators.py
@@ -1964,43 +1964,42 @@ class IS_PHONE_NUMBER(Validator):
                      is valid. With international=True, the value returned
                      is converted into E.123 international notation.
         """
+        # All non-strings are errors
+        if not isinstance(value, basestring):
+            error_message = self.error_message
+            if not error_message:
+                # default fallback message
+                error_message = current.T("Enter a valid phone number")
 
-        if isinstance(value, basestring):
-            value = value.strip()
-            if value and value[0] == unichr(8206):
-                # Strip the LRM character
-                value = value[1:]
-            number = s3_str(value)
-            number, error = s3_single_phone_requires(number)
-        else:
-            error = True
+            return (value, error_message)
 
-        if not error:
-            if self.international and \
-               current.deployment_settings \
-                      .get_msg_require_international_phone_numbers():
+        value = value.strip()
+        if value and value[0] == unichr(8206):
+            # Strip the LRM character
+            value = value[1:]
+        number = s3_str(value)
+        number, error_message = s3_single_phone_requires(number)
 
-                # Configure alternative error message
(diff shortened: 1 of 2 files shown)


nursix:
Sorry, but this is not correct.

This change eliminates self.error_message entirely - but actually, self.error_message must override any of the default error messages.

Admittedly, there is a bug: the default international-phone-number error is overridden again after being established, so the final `error_message = self.error_message` should actually come before that branch (and be common to both branches).

Still - the refactoring throws out the baby with the bathing water.

nursix:
Bug fixed in:
https://github.com/nursix/eden/commit/37537712aa390e1912bde13010ed609e5b0d27dc

modules/s3/s3validators.py
```diff
@@ -1964,43 +1964,42 @@ def __call__(self, value):
...
-        if not error:
-            if self.international and \
-               current.deployment_settings \
-                      .get_msg_require_international_phone_numbers():
+        value = value.strip()
+        if value and value[0] == unichr(8206):
+            # Strip the LRM character
+            value = value[1:]
+        number = s3_str(value)
+        number, error_message = s3_single_phone_requires(number)
```
nursix: This overrides self.error_message for all cases - but the opposite is required: self.error_message must override all defaults.

aeturnum: I made a mistake in reading the diff and thought the error from s3_single_phone_requires was returned in some cases. Thank you for fixing it and apologies for the bad fix. I understood the intent of the error message overriding, but I don't think there's a test case for it.

nursix: Ah - sorry, that was not intended as a lecture...I just wasn't sure whether or not you were aware of the requirement with the customizable error message.

Fixing it was however required, so the PR had value in any case even if not acceptable as-is.

nursix: And no, there's no test case as to what error message exactly is returned.

Whilst we could unit-test it routinely, regression-testing too much detail can add excessive rigor thus make development harder. A wrong error message is an annoyance, but neither does it break the intended functionality, nor does it jeopardize the data integrity - so it is low priority and should probably not be CI-tested.

And as you see - review caught it anyway ;)

aeturnum: Oh, I don't feel lectured at all - just embarrassed. Thank you for explaining your thinking. It's helpful to understand how to contribute. I realized the intents of the function and was trying to refactor it to have fewer returns and I forgot the second error message override. You're right that review caught it. I'm frustrated that I didn't catch it before submission! :)

PR #1532, opened 2020-02-08
Use == and != when comparing literals to fix python3.8 SyntaxWarning

trendspotter:
Comparing string and numerical literals using `is` and `is not` operators causes python 3.8 to emit SyntaxWarning, such as the ones below. This PR aims to fix the comparisons. Relevant issue on python bug tracker is https://bugs.python.org/issue34850

```
/srv/web2py/applications/eden/modules/s3/s3resource.py:3928: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if id_column is 0:
/srv/web2py/applications/eden/modules/s3/s3gis.py:1350: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif config_id is 0:
/srv/web2py/applications/eden/modules/s3/s3import.py:1203: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  elif totals[1] is not 0:
/srv/web2py/applications/eden/modules/s3/s3import.py:1205: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  elif totals[2] is not 0:
```

### DIFF
--- a/modules/nameparser/parser.py
+++ b/modules/nameparser/parser.py
@@ -691,7 +691,7 @@ class HumanName(object):
                 # http://code.google.com/p/python-nameparser/issues/detail?id=11
                 continue
 
-            if i is 0:
+            if i == 0:
                 new_piece = " ".join(pieces[i:i+2])
                 if self.is_title(pieces[i+1]):
                     # when joining to a title, make new_piece a title too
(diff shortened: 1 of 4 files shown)


flavour:
Great, thanks, I know we have a few Python 3.8 issues to resolve so good to get some done :)

Issue #1468, opened 2018-02-20
Translation export fails when "Select all templates" is used


trendspotter:
Steps to reproduce:
1. Go to `/eden/admin/translate/create`
2. Check the *Select all templates (All modules included)* checkbox
3. Enter a new language code (eg. *xx*)
4. Click *Submit*

Expected result:
XLS file with translation strings will be offered for download.

Actual result:
Error 500 + following traceback
```
Traceback (most recent call last):
  File "/srv/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "/srv/web2py/applications/eden/controllers/admin.py", line 1164, in <module>
  File "/srv/web2py/gluon/globals.py", line 419, in <lambda>
    self._caller = lambda f: f()
  File "/srv/web2py/applications/eden/controllers/admin.py", line 1087, in translate
    output = s3_rest_controller("translate", "language")
  File "/srv/web2py/applications/eden/models/00_utils.py", line 243, in s3_rest_controller
    output = r(**attr)
  File "applications/eden/modules/s3/s3rest.py", line 693, in __call__
    output = postprocess(self, output)
  File "/srv/web2py/applications/eden/controllers/admin.py", line 851, in postp
    output = X.export_file(code, modlist, [], filetype, all_template_flag)
  File "applications/eden/modules/s3/s3translate.py", line 1112, in export_file
    return self.write_xls(Strings, langcode)
  File "applications/eden/modules/s3/s3translate.py", line 1290, in write_xls
    wbk.save(output)
  File "/usr/lib/python2.7/site-packages/xlwt/Workbook.py", line 710, in save
    doc.save(filename_or_stream, self.get_biff_data())
  File "/usr/lib/python2.7/site-packages/xlwt/Workbook.py", line 674, in get_biff_data
    shared_str_table   = self.__sst_rec()
  File "/usr/lib/python2.7/site-packages/xlwt/Workbook.py", line 636, in __sst_rec
    return self.__sst.get_biff_record()
  File "/usr/lib/python2.7/site-packages/xlwt/BIFFRecords.py", line 77, in get_biff_record
    self._add_to_sst(s)
  File "/usr/lib/python2.7/site-packages/xlwt/BIFFRecords.py", line 92, in _add_to_sst
    u_str = upack2(s, self.encoding)
  File "/usr/lib/python2.7/site-packages/xlwt/UnicodeUtils.py", line 55, in upack2
    raise Exception('String longer than 32767 characters')
[...]

PR #1533, opened 2020-02-10
Fix S3Translate Pootle export

trendspotter:
This PR finishes the work done in #1468 also for Pootle files.

 - Fixes a variable forgotten during refactoring (line 810)
 - CSV files need to be always handled as text
 - _csv2po_ is imported and handled as python module instead of calling an external binary. This has several advantages
   - _csv2po_ binary is provided by _translate-toolkit_ python package. If we have the package, we don't need to run an extra process for it. Calling binaries is expensive.
   - subprocess.call() with shell=True (as required on Windows) requires the command to be a string, not list. This would require escaping for filename params (can be done via pipes.escape() on py<=2.6 and shlex.escape() on py>=2.7), but would uglify the code even further.
   - Errors in processing are obvious. E.g. when the _translate-toolkit_ was missing now, the error message complained about missing .po file and one needed to read the code to see that _csv2po_ has been called before and silently failed. Now the message would complain about missing module.
 - Temp file names for CSV and PO are generated from the same base name to ease debugging

Tested and works on
 - Python 2.7.17 @ GNU/Linux 4.15.0-66-generic x86_64 (Ubuntu 18.04, glibc)
 - Python 3.8.1 @ Linux 5.4.12-1-virt x86_64 (Alpine 3.11, musl libc)

### DIFF
--- a/modules/s3/s3translate.py
+++ b/modules/s3/s3translate.py
@@ -807,7 +807,7 @@ class TranslateReadFiles(object):
             f = open(user_file, "rb")
             user_data = f.read().decode("utf-8")
             f.close()
-            for line in f:
+            for line in user_data:
                 oappend(line)
 
         # Append user strings if not already present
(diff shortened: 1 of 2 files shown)


flavour:
Great, thanks...I think we can remove those @ToDos...having translate-toolkit as a python module is just fine...it was only the binaries I was trying to avoid :)

PR #1540, opened 2020-04-12
Switch GeoNames API URL to HTTPS endpoint

trendspotter:
GeoNames service can't be used on HTTPS-enabled Sahana Eden instances, because the GeoNames API is hardcoded with http:// protocol, which results in a request blocked due to mixed-content in sufficiently new browsers. This PR changes the GeoNames endpoint to the HTTPS-enabled one as given on http://forum.geonames.org/gforum/posts/list/35422.page


flavour:
Thanks!

PR #1552, opened 2020-04-24
Deploy: Fix missing gluon.html imports

trendspotter:
This PR crudely fixes missing gluon html classes imports in *Deployment* module, which cause a crash on alert creation.

Steps to reproduce:
1. Go to *Deployments* -> *Missions* -> *Create*
2. Create a mission filling in required fields (name)
3. Click on *Create Alert* in the mission
4. Create an alert filling in required fields (message)
5. Observe error after clicking on *Save*
```
  File "/srv/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "/srv/web2py/applications/eden/controllers/deploy.py", line 1001, in <module>
  File "/srv/web2py/gluon/globals.py", line 429, in <lambda>
    self._caller = lambda f: f()
  File "/srv/web2py/applications/eden/controllers/deploy.py", line 641, in alert
    return s3_rest_controller(rheader = s3db.deploy_rheader,
  File "/srv/web2py/applications/eden/models/00_utils.py", line 245, in s3_rest_controller
    output = r(**attr)
  File "/srv/web2py/applications/eden/modules/s3/s3rest.py", line 688, in __call__
    output = handler(self, **attr)
  File "/srv/web2py/applications/eden/modules/s3db/deploy.py", line 2340, in deploy_alert_select_recipients
    rheader = attr["rheader"](r)
  File "/srv/web2py/applications/eden/modules/s3db/deploy.py", line 1442, in deploy_rheader
    send_button = BUTTON(T("Send Alert"), _class="alert-send-btn")
NameError: name 'BUTTON' is not defined
```

The root cause is usage of wildcard imports (which is quite excessive in Sahana Eden), resulting in import of only whatever is exposed by the `__all__` module variable. In case of web2py's gluon it's just a subset of what gluon.html module exposes. Is there any interest to remove as much wildcard imports as conveniently possible?

Or do you think it would be worth to open a PR to have it fixed in web2py directly? The current status when only some html classes are exposed doesn't seem to be intentional.

### DIFF
--- a/modules/s3db/deploy.py
+++ b/modules/s3db/deploy.py
@@ -39,6 +39,7 @@ __all__ = ("S3DeploymentOrganisationModel",
            )
 
 from gluon import *
+from gluon.html import *
 from gluon.tools import callback
 
 from ..s3 import *


flavour:
Weird regression....looks like web2py must have made a change, so does seem better to fix there, but I see no major issue with working around it here too.
Non-wildcard imports can cause way too much maintenance overhead

nursix:
> Non-wildcard imports can cause way too much maintenance overhead
Agreed re:maintenance overheads.

However, I'm a strong advocate of explicit import-lists wherever bearable, especially to prevent name collisions and also to avoid cluttering the module namespace - and have converted most wildcard-imports in e.g. s3 modules. 

This also helps to prevent accidental cross-imports (i.e. wildcard imports from modules which themselves use wildcard imports, thus indirectly importing things you don't intend to import).

That said, I would not want a wholesale "fix" of all wildcard-imports - but indeed a more thoughtful approach: where it makes sense and doesn't unnecessarily increase maintenance efforts, yes. Otherwise perhaps no.

nursix:
And just to say it: modules/s3db/* all benefit maintenance-wise from wildcard imports, whilst modules/s3/* do most certainly not (except perhaps s3widgets.py).

So, no wholesale fixing - but thinking about where it makes sense and where not.

Issue #1554, opened 2020-04-25
Attempt to import invalid organization logo fails with ValueError


trendspotter:
Steps to reproduce:
1. Go to *Organizations* -> *Import*
2. Attempt to import organization with logo using URL without protocol - [organizations.txt](https://github.com/sahana/eden/files/4532604/organizations.txt) (rename to *.csv)
3. Clicking on *Submit* fails with
```
Traceback (most recent call last):
  File "/srv/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "/srv/web2py/applications/eden/controllers/org.py", line 511, in <module>
  File "/srv/web2py/gluon/globals.py", line 429, in <lambda>
    self._caller = lambda f: f()
  File "/srv/web2py/applications/eden/controllers/org.py", line 175, in organisation
    return s3db.org_organisation_controller()
  File "/srv/web2py/applications/eden/modules/s3db/org.py", line 6877, in org_organisation_controller
    output = current.rest_controller("org", "organisation",
  File "/srv/web2py/applications/eden/models/00_utils.py", line 245, in s3_rest_controller
    output = r(**attr)
  File "/srv/web2py/applications/eden/modules/s3/s3rest.py", line 688, in __call__
    output = handler(self, **attr)
  File "/srv/web2py/applications/eden/modules/s3/s3rest.py", line 1775, in __call__
    output = self.apply_method(r, **attr)
  File "/srv/web2py/applications/eden/modules/s3/s3import.py", line 274, in apply_method
    output = self.generate_job(r, **attr)
  File "/srv/web2py/applications/eden/modules/s3/s3import.py", line 405, in generate_job
    self._generate_import_job(upload_id,
  File "/srv/web2py/applications/eden/modules/s3/s3import.py", line 1004, in _generate_import_job
    resource.import_xml(src,
  File "/srv/web2py/applications/eden/modules/s3/s3resource.py", line 2483, in import_xml
    success = self.import_tree(id, tree,
  File "/srv/web2py/applications/eden/modules/s3/s3resource.py", line 2712, in import_tree
    success = add_item(element = element,
  File "/srv/web2py/applications/eden/modules/s3/s3import.py", line 3142, in add_item
    if not item.parse(element,
  File "/srv/web2py/applications/eden/modules/s3/s3import.py", line 1832, in parse
    data = xml.record(table, element,
  File "/srv/web2py/applications/eden/modules/s3/s3xml.py", line 1559, in record
[...]

PR #1555, opened 2020-04-25
Org: Allow year to be imported

trendspotter:
(Starting with something simple to get an idea how the imports work.)

This PR adds the option to import year the organization was founded, as discussed in #1554.


flavour:
Starting simple is good :)

PR #1556, opened 2020-04-25
Unify spelling of "organisation" in imports

trendspotter:
This PR unifies spelling of the word "*organi**s**ation*" in imports. (I can already see the frowns.)

There seems to be a rule that everything in backend is spelled as "*organi**s**ation*" and everything visible to user in frontend as "*organi**z**ation*", so there can be a single en-GB translation file which translates organizations into organisations.

The only 2 occurrences where this is not honored in imports are:
- *inv/inv_item* - where "*Owned By (Organi**z**ation/Branch)*" is used consistently.
- *doc/card_config* - where *Organi**s**ation Statement* is already everywhere **except** the CSV.

If you're not willing to accept the change in *inv_item* (which would be understandable since it's already used in production), then at least the *doc* one should be accepted as that's an obvious typo.

(Background: There's some confusion during preparation of import data when the user views the UI in the default American English.)


nursix:
That's right: the standard UI language is american English, whilst everything internal shall be British English.

With CSV templates that's slightly ambiguous: they are both user-facing and internal.

A possible strategy in this case would be to use column aliases (which match both spellings), but for a start we can agree that we use the internal British English as standard - and ideally consistently. So +1 for this PR.

nursix:
Btw - given that we're an international, multi-language team - such glitches are forgivable. 

That said, we must be careful not to take CSV column name changes too lightly: backwards-compatibility matters hugely in CSV imports (basically more important than English spelling), because there are so many CSVs out there that we cannot control. People will expect that they are still importable.

So, not frowning at the consistency you're after - but neither trying to encourage more changes like this ;) Sometimes, spelling details just don't matter so much.

PR #1565, opened 2021-04-23
Update Docker base image ubuntu to 18.04

PeterDaveHello:
Ubuntu 14.10 was EOL-ed for years, currently the Ubuntu version being
used on Travis CI by eden is 18.04(bionic), should align the versions.

Also set DEBIAN_FRONTEND=noninteractive to prevent interactive apt-get
dialog from tzdata package after upgraded to Ubuntu 18.04.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-FROM ubuntu:14.10
+FROM ubuntu:18.04
+
+ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install -y build-essential unzip curl python-pip python-dev python-matplotlib python-lxml \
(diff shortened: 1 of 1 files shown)


flavour:
Thanks, yes, Docker hasn't been maintained, so could use some love!
What I suggest the strategy should be is to have it use the same script to configure as Web Setup...this should minimise the maintenance work (although these core changes will still be needed, of course)
https://raw.githubusercontent.com/sahana/eden_deploy/master/user-data.sh

PeterDaveHello:
@flavour yes, it'll be better to use the same script to build it, before that, I can help make it work again, and follow some best practices to make it more robust and maintainable 😄

PR #1569, opened 2021-08-11
Enhance the number of features requested with wms getfeatureinfo

armin11:
This is e.g. needed when requesting clustered features. The default value of 10 is not enough. The minified version should be generated again with the closure compiler


flavour:
Thanks!
I applied the 1-line change manually rather than a merge...and also minified

armin11:
thanx ;-)

PR #1572, opened 2021-09-08
Re-routing REST controllers

nursix:
This allows re-routing of any controller/function combination to the custom/rest controller - which is a vanilla s3_rest_controller, i.e. without any prep, postp or attr. The latter can then be added in customise_controller.

To use this, a template will configure e.g.:
```
settings.base.rest_controllers = {("org", "organisation"): ("org", "organisation")}
```
...which re-routes org/organisation to custom/rest, and runs it as a s3_rest_controller with prefix="org" and name="organisation". This re-routing is completely transparent, i.e. the s3_rest_controller (and anything called from there) sees a normal org/organisation request. Consequently, to customise this call, one would use customise_org_organisation_controller (not customise_custom_rest_controller!), as usual.

As this allows re-routing of any controller/function combination, one can also "add" controllers that do not exist, e.g.:
```
settings.base.rest_controllers = {("org", "suppliers"): ("org", "organisation")}
```
This would though still be customised via customise_org_organisation_controller (=>can inspect r.function in prep to distinguish), but page access authorisation would now be through org/suppliers rather than org/organisation (so can have different permissions for those).

When adding a route with a custom module prefix:
```
settings.base.rest_controllers = {("my", "organisations"): ("org", "organisation")}
```
...then a corresponding module entry for that must be added to settings.modules - though this is useful anyway if menus are to be added automatically etc. Again, page authorisation would go through my/organisation here instead of org/organisation, so there could be different permission sets.

Note: this feature is orthogonal to default/index/*, which is still useful for implementing non-REST custom controllers, and should therefore be retained.


nursix:
I don't know what happened here with the other changes originally committed by you - did you remove these commits from trunk, or force-commit at some point? (they were even conflicting in the pull, but I checked out theirs, so this is originally yours).

PR #1573, opened 2021-09-09
Working minimal Demo of Custom Models

flavour:
This is NOT yet ready for merge...it is just a proof of concept for review/discussion.

It does work & doesn't seem to have major overheads.


models/zz_last.py
```diff
@@ -48,9 +48,9 @@
 
 # Re-routed REST controllers
 c, f = request.controller, request.function
-if c == "custom":
-    # Must not be accessed directly
-    raise HTTP(404, 'invalid controller (%s/%s)' % (c, f))
+#if c == "custom":
```
nursix, line 51: Not sure why you commented this? Given how it works, this would be a loophole to bypass page permissions.

nursix (commented):
Other than the problem with allowing direct access to the custom/rest controller, I see no general issues. So, go ahead.

models/zz_last.py
flavour: If the CUSTOM_PREFIX is 'custom' then we normally want to call the controller as custom/name, which would be blocked otherwise.
Obviously don't want a security hole...maybe block f=="rest" instead?

nursix: Yeah, or change the CUSTOM_PREFIX, simply ;)

nursix:
Remember to resolve the load_all_models case before merging it ;)

PR #1574, opened 2021-09-13
Better implementation of custom models with some old modules moved in…

flavour:
…to their templates


flavour:
What do you think? Looks really nice to me :)
We can do the same thing with moving DVR into DRKCM

PR #1576, opened 2021-09-17
BT: Disable Postcodes

flavour:


nursix:
How can I possibly know whether this is correct, or necessary? Makes no sense to request a review if I cannot decide anything.

PR #1590, opened 2023-01-10
A branch will never be taken.

openrefactory:
In file: parser.py, the comparison of Collection length creates a logical short circuit. The path in the code will never be taken. I am not suggesting a fix, just raising attention to the inconsistency here.


flavour:
This code is in a 3rd-party vendored submodule, so I am reluctant to patch it unless absolutely necessary.
We can update our submodule if there is a better upstream version, or if it still needs fixing upstream then that is best done there.

